PR #210, opened 2015-01-16
Yices native support

mebsout:
Support for native Yices 1 input.
Need to revert the change regarding extra properties for subrange constraints (remove mutable type).


chsticksel:
Merged to separate branch, see #211

Issue #212, opened 2015-01-20
Strange condact behavior


agacek:
For the following Lustre program, Kind 2 reports a counterexample of length 4, but I believe the minimal counterexample length is 12.

```
node pass(x : int) returns (y : int);
let
  y = 0 -> pre x;
tel;

node main() returns ();
var
  count : int;
  toggle : bool;
  ok : bool;
  slow_count : int;
let
  count = 3 -> 1 + pre count;
  toggle = false -> not pre toggle;
  slow_count = condact(toggle, pass(count), -1);

  ok = slow_count <= 10;
  --%PROPERTY ok;
tel;
```

PR #218, opened 2015-01-28
Bugfix condact init flag

AdrienChampion:
In condact translation, `ticked` flag has been replaced by a `first_tick` flag which is true from the first state up to the state when the clock ticks for the first time, including that state. The flag is then forever false.

This flag is passed down to the called node as its init flag. This is used by two state invariant lifting to rewrite init guarding: `init or two_state_inv` becomes `fist_tick or two_state_inv` and then `clock => first_tick or two_state_inv`.

The problem before was that the init flag was passed down, instead of the `first_tick` flag. But as the predicates of the subnode constrain their init flag formal parameter, the transition relation could become unsatisfiable.

As a consequence, the equations encoding the condact behavior now are:
```
init = (first_tick = true)
  and (clock => shadow_input = actual_input)
  and (clock => init(first_tick, args))
  and (not clock => out = default)

trans = (first_tick' = first_tick and not clock)
  and (clock' => shadow_input' = actual_input')
  and (not clock' => args' = args)
  and (clock' and first_tick' => init(first_tick', args'))
  and (clock' and not first_tick' => trans(first_tick', args', first_tick, args))
```

For a more in-depth explanation, see the documentation of `lustreTransSys.ml(i)`.


chsticksel:
Fixes #212

PR #229, opened 2015-02-18
Modular + compositional analysis (contracts)

AdrienChampion:
See documentation for details regarding modular and compositional analysis.

#### Contract node / contract calls

Contract nodes are used to factor contracts out of the nodes they specify. Contract nodes look like
```
contract <ident> (<typed_ident_list>) returns (<typed_ident_list>) ;
let
  --@require ... ;
  ...
  --@ensure ... ;
  ...
tel
```
Contract nodes can be called in the specification of a node:
```
node bla (...) returns (...) ;
--@contract <ident> ;
...
```
For the contract call to be legal, the set of inputs (outputs) of the contract node must be a subset of the set of inputs (outputs) of the node, with *consistent* types. "Consistent" means the same type, or a subtype in the subrange of int case. If the node specified has a subrange of int input (output) `in`, then a contract with an input (output) `in` of type int is considered consistent.

#### Restrictions:

* local variables are not allowed in contract nodes for now, we need ghost local vars for that.
* the types appearing in the signature of a contract node cannot be dependent types: no arrays or subranges.

#### Encoding:

Contract nodes and contract calls are handled by a preprocessing step which inlines contract calls in all nodes, and discards now useless contract nodes. This is done right after constructing the Lustre AST.


AdrienChampion:
Better version coming soon.

PR #265, opened 2015-05-06
Allow forward referenced nodes.

mebsout:
Might be inefficient for large files.


mebsout:
We also probably need to do this for types and constants.

PR #273, opened 2015-05-29
Tiny tests

AdrienChampion:
* modified Kind 2 exit code upon termination without an error
* wrote a script to run regression tests

## Exit code

Upon termination without error, Kind 2 returns with code

* 0 if some properties are Unknown or KTrue (timeout)
* 10 if some properties are False (falsifiable)
* 20 if all properties are Invariant (success)

## Regression tests

The script `tests/run.sh` runs tests using a directory naming convention. The files are separated in *normal* and *contract* (will run in modular and compositional mode). The files for both are then divided into "success", "falsifiable" and "error". See help for more details:

```
> ./tests/run.sh -h
Usage: run.sh <DIR> <CMD>
with
  * <DIR> the test directory
  * <CMD> the Kind 2 command to test
(Passing "-h" or "--help" as argument prints this message.)

Tests Kind 2 on the lustre files in a directory. The directory should be
structured as follows. Directory
  * "success"     contains the files with only valid properties
  * "falsifiable" contains the files with at least one falsifiable property
  * "error"       contains the files on which Kind 2 fails with error code 2
  * "contracts"   contains the files to run in modular and compositional mode
                  organized in "success", "falsifiable" and "error" as
                  described above.
```

Some basic tests are already set up in `tests/regression/`.


AdrienChampion:
Wrong branch, my bad...

PR #331, opened 2015-10-13
Mode identifiers refer to a mode's requirements

AdrienChampion:
Long story short, it is possible to use the name *n* of a mode above the current one to refer to the predicate "mode *n* is active", *i.e.* all its requirements are true.

Note that the following cause parse errors:
* self/forward reference: "Undeclared identifier", see regression test [here][self ref test] and [here][forward ref test]
* shadowing of an identifier in scope: "mode identifier *id* is shadowing another identifier", see regression test [here][shadowing test]

For an actual example see the very last section of this PR, after Motivation and Details.

# Motivation

## Init mode

A very simple example is when we have an `init` mode. It's useful to have one when in the first states the
system is doing things which are hard to describe and we don't really care about. If we have a `duration` node that counts for how long some boolean has been true, we can write
```
@mode init (
  @require duration(true) < init_steps
) ;
@mode nominal(
  @require not init
  ...
) ;
...
```

This is a more natural, less error-prone way of requiring `duration(true) >= init_steps` in the modes different from `init`.

Obviously this is not interesting enough to motivate this pull request, but the next example is.

## Mode / numeric separation in the model

*The following is a (brutal) simplification of my experience of specifying the TCM.*

Consider a node which inputs are sensor readings and outputs are actuator commands along with two bools `left` and `right`. Say the node only calls two nodes: `mode_logic` and `numeric_stuff`.
* `mode_logic` is given the sensor readings and returns `left` and `right`;
* `numeric_stuff` outputs actuator commands based on sensor readings, `left` and `right.

We write the contract, and say wether `left` or `right` is true is not simple and we want to make the most of multi-property.
So we have several modes the ensures of which is `left`, and the same for `right`.

```
@mode left_1 (
  @require <something> ;
  @require <complex> ;
  @ensure left ;
) ;
@mode left_2 (
  @require <something> ;
  @require <else> ;
  @ensure left ;
) ;
...
```

Now we write contracts about what the actuator commands should be, and that depends on wether `left` or `right` is active. It does not depend on which mode `left_`i --or `right_`i-- is active, just that `left` is true.

Obviously we don't want to `require left` in the numeric specification as `left` is an output.

We don't want to write the numeric specification of what happens when `left` in each of the `left_`i either as it's redundant, and the numeric specification itself might consist of several modes. We get way too many modes, its super messy/unmaintainable/error-prone.

So with this new feature we just do
```
@mode num_left_1 (
  @require left_1 or left_2 or ... or left_n ;
  @require <something>
  @require <complex> ;
  @ensure we_tell_actuators_to_go_left ;
) ;
@mode num_left_2 (
  @require left_1 or left_2 or ... or left_n ;
  @require <something> ;
  @require <else> ;
  @ensure we_tell_actuators_to_go_left_differently_maybe ;
) ;
...
```

# Details

When the expressions corresponding to the statements inside a node are constructed, the context is augmented with an identifier/expression substitution for each mode `m` previously constructed in the
current contract.
The substitution is `id -> and(req_1, ..., req_n)` where `id` is the identifier (name) of `m`, and `req_1` ... `req_n` are its requirements.

Since it's a substitution at context level, the identifier is replaced with the conjunction of requirements when creating the expression: it is not a let-binding, and no state variable for the mode identifier is created.

# Example

[Regression test][nominal test], mode `always_true` requires that mode `false_so_far` is not active:

```
node abs (in: int) returns (out: int) ;
let
  out = if in < 0 then -in else in ;
  (*%PROPERTY out >= 0 ; *)
  /*!PROPERTY : if in < 0 then (out = -in) else (out = in) ; */
tel

node has_been_true(in: bool) returns (out: bool) ;

/*@contract

  @mode false_so_far (
    @require not (in or (false -> pre out)) ;
    @ensure not out ;
  ) ;

  @mode in_true (
    @require in ;
    @ensure out ;
  ) ;

  @mode always_true (
    @require not false_so_far ;
    @require in -> pre out ;
    @ensure out ;
  ) ;
*/
let
  out = in or (false -> pre out) ;
tel


node my_node (in: int) returns (out: int) ;

(*@contract

  const bound = 10 ;
  var in_bounded = abs(in) < bound ;

  @require in_bounded ;
  @ensure out >= - bound ;

  @mode in_neg (
    @require in < 0 ;
    @ensure out = in ;
  ) ;

  @mode in_pos (
    @require in >= 0 ;
    @ensure out >= 0 ;
    @ensure true -> (
      (pre has_been_true(in > 0)) => out > in
    ) ;
  ) ;
*)
var mem: int ;
let
  mem = abs(in) + (0 -> pre mem) ;
  out = if in < 0 then in else mem ;

  (* Old style still works. *)
  --%PROPERTY out - in >= 0 ;

  (* Doing the same thing using different annotations. *)
  (*!MAIN: true ; *)
  (*%MAIN ; *)
  --%MAIN ;
  --!MAIN: true ;
tel
```

[self ref test]: https://github.com/AdrienChampion/kind2/blob/contracts-when-clause/tests/regression/error/mode_reqs_by_idents_no_self_ref.lus (Self reference regression test)
[forward ref test]: https://github.com/AdrienChampion/kind2/blob/contracts-when-clause/tests/regression/error/mode_reqs_by_idents_no_forward_ref.lus (Forward reference regression test)
[shadowing test]: https://github.com/AdrienChampion/kind2/blob/contracts-when-clause/tests/regression/error/mode_reqs_by_idents_shadowing.lus (Shadowing regression test)
[nominal test]: https://github.com/AdrienChampion/kind2/blob/contracts-when-clause/tests/regression/success/mode_reqs_by_idents.lus (Nominal test for mode identifiers)


tinelli:
Looks good.

PR #373, opened 2016-04-20
To rust improvements docs

AdrienChampion:
Original lustre files are now included in the docs and can be (line-)referenced from the docs.

- also rmed spurious print statement from previous PR
- fixed contract item ordering (most were reversed, ok internally but weird in result of compilation)


AdrienChampion:
Closed to relaunch the tests

PR #379, opened 2016-05-17
Step cex xml fix

AdrienChampion:


AdrienChampion:
Re-opening again I guess.

PR #388, opened 2016-06-03
Fix to set_status in TransSys

AdrienChampion:
Fixes a bug when two property have the same name, *i.e.* either the same expression or the same user-specified name.
In that case, assuming it proved both, Kind 2 would say it manages to prove one but never terminate.


mebsout:
I might have overridden this with my last pull request #389. I'll check.

PR #390, opened 2016-06-06
Fixes regarding unguarded pre-s in contracts

AdrienChampion:
Bans unguarded pre-s from

- contract items: assumes, guarantees, requires, ensures
- contract imports: in/out params
- node called in contracts

So, unguarded pre-s are completely banned from the cone of influence of the contracts.


AdrienChampion:
Re-opening. This is getting old.

PR #394, opened 2016-06-10
Mode information in XML output + subdir decomposition

AdrienChampion:
Does what it says.

XML schema has been updated.


AdrienChampion:
...

PR #396, opened 2016-06-15
Fixes for Rust generation (implem/oracles) and test generation

AdrienChampion:
Fixes

- name clashes in test gen for nested contract imports
- float problem in rust generation
- problems when compiling mode references under a pre
- most ocaml 4.03 warnings


AdrienChampion:
Ocaml 4.03 warnings fixes moved to https://github.com/kind2-mc/kind2/pull/397.

AdrienChampion:
Actually reverted the 4.03 fixes

AdrienChampion:
Store invariants as an hashtbl to certificates

mebsout:
Is it a mistake 194af1a ?

AdrienChampion:
Yes and no.

I ended up committing the fixes for ocaml 4.03 in the branch used for the PR which messed up the whole thing. I tried to revert the commit (which is what you're pointing to) but I failed miserably. In the end I backtracked the 4.03 fixes by hand (https://github.com/kind2-mc/kind2/pull/396/commits/890b22254fc8c61eeb413bbaf62e1f19bfdbd4fd) and that's the version that got merged in.

I just checked and the fixes announced in the original PR post are indeed there, so everything should be fine.

PR #400, opened 2016-06-17
Use relaxed type (for unproven int subranges) in declarations for SMT solvers and stability parallelism

mebsout:


AdrienChampion:
Stability's looking good :D

PR #426, opened 2016-10-20
Restart of nodes

mebsout:
This extension of Lustre (which is in Scade 6) allows to reset nodes to their internal states with a call of the form:
```Lustre
(restart n every c)(x1, ..., xn)
```

This is a call to the node `n(x1, ..., xn)` and every time the Boolean stream `c` is true, the internal state of this instance of `n` is reset to its initial state.


mebsout:
Let me know if this extension and the syntax makes sense (it will be useful to encode automaton with clocks).

src/lustre/lustreTransSys.ml
```diff
@@ -763,10 +764,56 @@ let rec constraints_of_node_calls
...
+           instance does not have an activation condition *)
+        (fun _ t -> t)
+        subsystems
+    in
+
+    let restart_trans = E.cur_term_of_state_var TransSys.trans_base restart in
+    (* Reset state of node to initial state when restart condition is true *)
+    let trans_term = Term.mk_ite restart_trans
+        (Term.bump_state Numeral.(TransSys.trans_base - E.cur_offset) init_term)
+        trans_term in
```
mebsout: The encoding for TransSys is performed here.

PR #431, opened 2016-11-14
Automata

mebsout:
## Enumerated data types in Lustre

```
type t = enum { A, B, C };
node n (x : enum { C1, C2 }, ...) ...
```
Enumerated datatypes are encoded as subranges so that solvers handle arithmetic constraints only. This also allows to use the already present quantifier instantiation techniques in Kind 2.

## N-way merge

Possibility to do a merge on a clock of an enumerated datatype.

```
merge c
 (A -> x when A(c))
 (B -> w + 1 when B(c));
```

Arguments of merge have to be sampled with the correct clock. Clock expressions for merge can be just a clock identifier or its negation or `A(c)` which is a stream that is true whenever `c = A`.

Merging on a Boolean clock can be done with two equivalent syntaxes:

```
merge(c; a when c; b when not c);

merge c
  (true -> a when c)
  (false -> b when not c); 
```

## Node restart / reset

Node can be reseted (or restarted, that's just a terminology naming) with a construct similar to `activate`. This follows the Scade 6 syntax.

```
(restart n every c)(a1, ..., an)
```

is a call to the node `n` that is restarted whenever the clock `c` is true. In practice, when this is the case, the _internal state_ of the node is reset to its initial value.

Restart and activate can also be combined in the following way:

```
(activate (restart n every r) every c)(a1, ..., an)
(activate n every c restart every r)(a1, ..., an)
```

These two calls are the same (the second one is just syntactic sugar). The (instance of the) node `n` is restarted whenever `r` is true and the _resulting call_ is activated when the clock `c` is true. Notice that the restart clock `r` is also sampled by `c` in this call.

## Automata in Lustre

The major contribution of this pull request is the support for Lustre automata. The syntax is very close 
to the one of Scade 6.

```
node n (i1, ..., in : ...) returns (o1, ..., on : ...);
let

   automaton automaton_name

     initial state S1:
       unless if c restart Si elsif c' resume Sj else restart Sk end;
       var v : ...;
       let
          v = ...;
          o1 = i1 -> last o2 + 1;
          o2 = 99;
       tel
       until c restart S2;

     state S2:
       let
          ...; 
       tel
     ...
   returns o1, o2;

   o3 = something () ...;
tel
```

An automaton is declared _inside a node_ (there can be several) and can be anonymous. This is followed by a list of states and a `returns` statement that specifies which variables (locals or output) are modified by the automaton.

Each state has a name and one of them can be declared `initial` (if no initial state is specified, the first one is considered initial). They can have local variables (to the state).
The body of the state contains Lustre equations (or assertions) and can use the operator `last`. In contrast to `pre x` which is the value of the stream `x` the last time the state was in the declared state, `last x` is the previous value of the stream `x` on the base clock. This construct is useful for communicating information between states.

States can have a _strong_ transition (declared with `unless`) placed before the body and a _weak_ transition placed after the body. The unless transition is taken when entering the state, whereas the until transition is evaluated after execution of the body of the state. If none are applicable then the automaton remains in the same state. These transitions express conditions to change states following a branching pattern. Following are examples of legal branching patterns (`c*` are Lustre Boolean expressions):

```
c restart S
```
```
if c1 restart S1
elsif c2 restart S2
elsif c3 restart S3
end;
```
```
if c1
  if c2 restart S2
  else if c3 resume S1
  end
elsif c3 resume S3
else restart S0
end;
```

Targets are of the form `restart State_name` or `resume State_name`.  When transiting to a state with `restart`, the internal state of the state is rested to its initial value. On the contrary when transiting with `resume`, execution in the state resumes to where it was when the state was last executed.

In counter-examples, we show the value of additional internal state information for each automaton:
`automaton_name.state` is a stream that denotes the state in which the automaton is and `automaton_name.restart` indicates if the state in which the automaton is was restarted in the current instant.

## Other bug fixes

- Functional constraints that are added in the solver are now scoped.
- A bug was fixed for abstracting of `pre` which was not done correctly in some cases.


mebsout:
Mmmh, we have a performance issue with the non-guarded pre. Too many intermediate state variables for some benchmarks. This is due to the fact that we want `pre x = pre x` not to be true if non-guarded.

AdrienChampion:
I'm not saying yet again that we should ban unguarded `pre`'s because it makes no sense in a model checker on so many levels.

Nice try.

PR #508, opened 2018-01-11
Provide decimal approximations in models

correnson:
When Kind-2 outputs counter examples, decimal values are represented by rationals which are, in my opinion,  cumbersome to read. By the way, those rationals are messy Lustre values, since _eg._ `1/8` is actually the _integer_ `0` and not the expected _real_ `0.125` value.

Indeed, the code documentation (which is quite excellent, I must say) for printing model values contains this note (file `model.ml`, line 71):

> `(* TODO do we not want legal lustre values? *)`

Hence, I experimented printing an approximation of rationals in models, plus an upper bound of the « rounding » error. For instance, instead of reading decimal `1215678/251162565001`,
you would read now `+4.8402e-06+f22`, which means: approximate value `+4.8402e-06` with a (positive) _relative_ error less than `2^-22` (which fits on floating point number with 22-bits mantissa).

I found those informations much more useful than the rational notations when interpreting models in textual mode. But probably you would like to add some flags to control such a display.

My suggestion (_not_ implemented yet) for such a new flag would be:

```
--model-precision (R|E|P)
        adjust precision of decimals in models
        R: print rational value (eg. 1/8)
        E: print scientific value (eg. 3.4e-6)
        P: print relative error (eg. 3.4e-6+f21) 
            (f<nn> means a relative error less than 2^-nn)
```

### DIFF
--- a/src/flags.ml
+++ b/src/flags.ml
@@ -1901,6 +1901,43 @@ module Global = struct
   let output_dir () = !output_dir
 
 
+  (* Real precision. *)
+  type real_precision = [
+    `Rational | `Float
+  ]
+  let real_precision_of_string = function
+    | "rational" -> `Rational
+    | "float" -> `Float
+    | _ -> raise (Arg.Bad "Bad value for --real_precision")
+  let string_of_real_precision = function
+    | `Rational -> "rational"
+    | `Float -> "float"
+  let real_precision_values = [
+    `Rational ; `Float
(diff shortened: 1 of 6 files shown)


daniel-larraz:
Hello @correnson. Thank you again for your contribution. I added a new flag called `--real_precision` to print real values in different formats.

correnson:
Thanks !

PR #575, opened 2019-11-09
Interpreter

E-Sh4rk:
Add support for JSON inputs for the interpreter.
Arrays, tuples and records are supported.
A page about the interpreter has been added to the documentation.


E-Sh4rk:
It must be failing because of the new Yojson dependency. I added it to Makefile.in but if the package is not installed it fails...

PR #637, opened 2020-06-16
Catchup with ocaml 4.09

apoorvingle:
- `make test`-ed with ocaml 4.07 and 4.09
- `Pervasives` -> `Stdlib` 
- Travis CI may need changes as we now depend on a new library.(Edit: updated now)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 # Log files.
 *.log
+*.err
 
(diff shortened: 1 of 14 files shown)


apoorvingle:
Opam upgrade is also needed for travis (1.2.2), while I am using 2.0.7

PR #638, opened 2020-06-17
Allow arrays in top-level constant declarations

amosr:
Hi Daniel,

There was a bug where a top-level constant declaration like
```
const simple_array_literal : int^3 = [1, 5, 9];
```
was treated as a type error. This was happening because the type checking
for constants expected the trie of the type to be the same shape as
the trie of the values. However, for arrays, the values had a shape like
`[0]: _, [1]: _, [2]: _`
but the array in the type trie hadn't been flattened out, so it had
no nesting. I've fixed this for now by flattening out the array in the
type trie, so in this example, the original type:
```
Array Int (IntRange 0 3)
```
becomes the trie:
```
[0]: Int, [1]: Int, [2]: Int
```
and then we can more easily match each element of the value trie against its corresponding entry in the type trie.

I couldn't find an obviously nicer way to do this, but I got a bit lost in the type checking for node equations. Let me know if you think it can be made cleaner.

Thanks,
Amos

### DIFF
--- a/src/lustre/lustreDeclarations.ml
+++ b/src/lustre/lustreDeclarations.ml
@@ -419,8 +419,9 @@ let eval_const_decl ?(ghost = false) ctx = function
 
         (try 
 
-           (* Evaluate type expression *)
-           let type_expr' = S.eval_ast_type ctx type_expr in 
+           (* Evaluate type expression. Flatten any arrays so that we check
+            * each element against its expected type *)
+           let type_expr' = S.eval_ast_type_flatten true ctx type_expr in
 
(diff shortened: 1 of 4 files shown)


daniel-larraz:
Hi Amos,

Thank you for sharing this fix with us.

We are aware that there are some issues with our current type checker for compound types. Improving that part of the code is in our Todo list. In the meantime, we appreciate any fix you or other community members may provide to us.

Thanks,
Daniel

amosr:
Cheers - thanks for the quick fix for slice_nodes as well.

Amos

PR #665, opened 2020-08-08
Lustre Ast Helpers

apoorvingle:
Refactor `lustreAst` into two modules 
1. `LustreAst` for their structure and pretty printing
2. `LustreAstHelpers` for ast manipulations and misc structure checks


src/lustre/lustreAstHelpers.mli
```diff
@@ -0,0 +1,49 @@
+(* This file is part of the Kind 2 model checker.
+
+   Copyright (c) 2020 by the Board of Trustees of the University of Iowa
```
daniel-larraz, line 3: Since this new file only includes old code, you should write the year when the code was originally written (2015). If you add new code later, you should update the copyright line as follows:

>  Copyright (c) 2015, 2020 by [...]

src/lustre/lustreAstHelpers.ml
```diff
@@ -0,0 +1,599 @@
+(* This file is part of the Kind 2 model checker.
+
+   Copyright (c) 2020 by the Board of Trustees of the University of Iowa
```
daniel-larraz, line 3: Since this new file only includes old code, you should write the year when the code was originally written (2015). If you add new code later, you should update the copyright line as follows:

>  Copyright (c) 2015, 2020 by [...]

src/lustre/lustreAstHelpers.mli
apoorvingle: this is fixed now.

src/lustre/lustreAstHelpers.ml
apoorvingle: This is fixed now.

PR #674, opened 2020-08-31
Building graph and obtaining dependency sorted lustre declarations for typechecking

apoorvingle:
- Detects circular dependencies and throws an error for constant and type declarations.

- Updates readme for unit test changes
    - makes build now runs `dune build src @install` 

- The github CI flow now installs `ounit2` and runs unit tests on `Graph` and `GraphAdapter` modules


apoorvingle:
How do we add `ounit2` as a test only dependency using dune?

daniel-larraz:
> How do we add `ounit2` as a test only dependency using dune?

The default Makefile rule runs `dune build` which builds everything that is installable. An easy solution is to change the `build` rule in the Makefile so that it runs `dune build src @install` instead of `dune build`. In addition, section _Direct Installation Using Dune_ in `doc/usr/source/home.rst` should be updated and the README file generated again.

.github/workflows/kind2-ci.yml
```diff
@@ -17,9 +17,11 @@ jobs:
           - os: macos-10.15
             z3_install: brew install z3
             zmq_install: brew install zmq
+            ounit2_install: opam install ounit2
```
daniel-larraz, line 20: If the command for installing ounit2 is the same for both operating systems, put the command where it is used without introducing an intermediate variable

apoorvingle: fixed

src/utils/graph.ml
```diff
@@ -0,0 +1,306 @@
...
+    include (Set.Make (struct
+                 type t = (Ord.t *  Ord.t)
+                 let compare (v11, v12) (v21, v22)
+                   = let c = Ord.compare v11 v21 in
+                     if c <> 0 then c else Ord.compare v12 v22
+               end))
+    let flatten: t list -> t = fun sets ->
+      List.fold_left union empty sets
+  end 
+  (** A set of vertices  *)
```
daniel-larraz, line 176: A set of **edges**

src/utils/graph.mli
```diff
@@ -0,0 +1,118 @@
...
+  (** returns true if the graph has only one vertex *)
+  
+  val add_vertex: t ->  vertex ->  t
+  (** Add a [vertex] to a graph  *)
+
+  val add_edge: t ->  edge ->  t
+  (** Add an [edge] to a graph  *)
+
+  val remove_vertex: t ->  vertex ->  t
+  (** Remove a [vertex] from a graph *)
```
daniel-larraz, line 84: (** Remove a [vertex] from a graph **and its associated [edges]** *)

tests/ounit/utils/testGraph.ml
```diff
@@ -0,0 +1,58 @@
...
+
+let dos_connected_g = G.add_edge dos_g (G.mk_edge v0 v1)
+let dos_cycle_g = G.add_edge dos_connected_g (G.mk_edge v1 v0)
+
+let g_verts = List.fold_left (G.add_vertex) G.empty ["t0";"t1";"t2";"int"; "c"; "1"]
+let g = List.fold_left (G.add_edge) g_verts [G.mk_edge "t0" "t1"
+                                       ; G.mk_edge "t1" "t2"
+                                       ; G.mk_edge "t2" "int"
+                                       ; G.mk_edge "c" "t0" 
+                                       ; G.mk_edge "c" "1" ]
```
daniel-larraz, line 44: `g_verts` and `g` are not used. It would be good to have a test with them.

src/utils/graph.ml
```diff
@@ -0,0 +1,306 @@
...
+        r_topological_sort_helper new_g (sorted_vs @ VSet.elements no_outgoing_vs)
+    in r_topological_sort_helper g []
+  (** Computes a topological ordering of vertices 
+   *  or throws an [CyclicGraphException] if the graph is cyclic.
+   *  Implimentation is based on Kahn's algorithm 
+   * https://en.wikipedia.org/wiki/Topological_sorting *)
+
+  let topological_sort: t -> vertex list = fun g ->
+    Log.log L_trace "Phase 1.1 Building graph\n---------\n"
+    ; reverse_topological_sort g |> List.rev
```
daniel-larraz, line 304: `reverse_topological_sort` seems to return the elements already in topological order. Why are you reversing the result?

tests/ounit/utils/testGraph.ml
apoorvingle: We check this in `TestGraphAdapter` I have removed this now.

src/utils/graph.ml
apoorvingle: It is necessary given the way we build the dependency graph. 
`a -> b` means that `a` depends on `b`, the topological sort would mean `[a, b]`
but we need a dependency sort, not topological sort. Hence we expect an order `[b, a]`

Maybe I should change the name of the function...

src/utils/graph.ml
apoorvingle: fixed

src/utils/graph.mli
apoorvingle: fixed

src/utils/graph.ml
daniel-larraz: The current names of the functions are misleading. Since you are reversing the list explicitly before returning the result, that is, you are not returning the reversed list as as a side-effect of your implementation, you should consider reversing the list outside this function, in the function that computes dependencies.

apoorvingle: I have changed the function names now.

src/utils/graph.ml
```diff
@@ -0,0 +1,301 @@
...
+                                          
+  let is_point_graph: t -> bool = fun (vs, es) ->
+    ESet.is_empty es
+    
+  let union: t -> t -> t = fun (v1s, e1s) (v2s, e2s) ->
+    (VSet.union v1s v2s, ESet.union e1s e2s) 
+  (** Unions two graphs *)
+    
+  let topological_sort: t -> vertex list = fun ((vs, es) as g) ->
+    let rec r_topological_sort_helper: t -> vertex list -> vertex list
```
daniel-larraz, line 277: The `r_` prefix in `r_topological_sort_helper` doesn’t make sense anymore. Maybe you can just call the function `topological_sort'` following extended naming conventions for auxiliary helper functions.

apoorvingle: i would rather prefer `topological_sort_helper` rather than `topological_sort'` 
the `'` is a very small difference and can easily be missed while reading.

daniel-larraz: Sounds good to me too.

src/lustre/lustreTypeChecker.ml
```diff
@@ -1114,7 +1112,7 @@ and check_type_struct_def: tc_context -> LA.eq_lhs -> tc_type -> unit tc_result
 and tc_ctx_contract_eqn: tc_context -> LA.contract_node_equation -> tc_context tc_result
   = fun ctx -> function
   | GhostConst c -> tc_ctx_const_decl ctx c 
-  | GhostVar c -> tc_ctx_const_decl ctx c
+  | GhostVar c -> tc_ctx_const_decl ~const_real:false ctx c
```
daniel-larraz, line 1115: Suggestion: `is_const`

apoorvingle: fixed

PR #676, opened 2020-09-08
Added functions to convert from machine integers to integers

arjunvish:


src/simplify.ml
```diff
@@ -2003,6 +2011,134 @@ let rec simplify_term_node default_of_var uf_defs model fterm args =
...
+          
+          | `UINT8_TO_INT ->
+
+            (match args with 
+            
+            | [UBV b] -> let s = (Bitvector.length_of_bitvector (Term.bitvector_of_term b)) in
+                  (match s with 
+                  | 8 -> Num (Numeral.zero, 
+                    [Numeral.one, [Term.mk_num (Bitvector.ubv8_to_num (Term.bitvector_of_term b))]])
+                  | _ -> assert false)
```
daniel-larraz, line 2023: The current code doesn’t create a constant polynomial. Expected code:
```
let bv = Term.bitvector_of_term b in
let s = Bitvector.length_of_bitvector bv in
(match s with
 | 8 -> Num (Bitvector.ubv8_to_num bv, [])
 | _ -> assert false
)
```
The same comment applies to the rest of casting operations.

PR #678, opened 2020-09-24
Lustre type checker iteration 2

apoorvingle:
- Tuple projection type checking
- Array size expressions are tracked with ast manipulations
- Contract returns a list of identifiers instead of expressions
- Anonymous enums cannot appear as types of identifiers
- Moved helper functions from lusterTypeChecker to lusterAstHelpers
- Separate name spaces for ident/consts/nodes/functions and contracts
- Updated syntax error messages to be more descriptive.


src/lustre/lustreParser.mly
```diff
@@ -828,9 +827,9 @@ pexpr(Q):
   (* An array constructor (not quantified) *)
   | e1 = pexpr(Q); CARET; e2 = expr { A.ArrayConstr (mk_pos $startpos, e1, e2) }
 
-  (* An array slice or tuple projection (not quantified) *)
-  | e = pexpr(Q); DOTPERCENT; i = expr 
-    { A.TupleProject (mk_pos $startpos, e, i) }
+  (* Tuple projection (not quantified) *)
+  | e = pexpr(Q); DOTPERCENT; i = NUMERAL 
+    { A.TupleProject (mk_pos $startpos, e, (int_of_string i)) }
```
daniel-larraz, line 832: The call to `int_of_string` may raise `Failure "int_of_string"` if if the integer represented exceeds the range of integers representable in type `int`. You should handle the exception and convert it to a proper parsing error. Alternatively, you can use a Numeral instead of a `int`.

src/lustre/lustreTypeChecker.ml
```diff
@@ -41,13 +41,42 @@ type tc_type  = LA.lustre_type
...
+  let pp_print_ty_syns: Format.formatter -> 'a t -> unit
+    = fun ppf m -> Lib.pp_print_list (pp_print_type_syn) ", " ppf (bindings m)
+
+  (** Pretty print type context *)
+  let pp_print_tymap: Format.formatter -> 'a t -> unit
+    = fun ppf m -> Lib.pp_print_list (pp_print_type_binding) ", " ppf (bindings m)
+
+  (** Pretty print value store *)
+  let pp_print_vstore: Format.formatter -> 'a t -> unit
+    = fun ppf m ->  Lib.pp_print_list (pp_print_val_binding) ", "  ppf (bindings m)
```
daniel-larraz, line 76: Keep these functions outside the IMap module.

tests/lustre/typechecking/futurePass/test_add_tuples.lus
```diff
@@ -0,0 +1,7 @@
+node n (x, y: int) returns (p,q: int);
+(*@contract
+    guarantee p = q;
+*)
+let
+(p, q) = (x, y) + (y, x);
+tel
```
daniel-larraz, line 7: This should go to futureFail.

apoorvingle: Well, this fails right now, so may be in `shouldFail` as that is the right behavior :)

daniel-larraz: Yes, you are right.

src/lustre/lustreParser.mly
apoorvingle: I see that there is [another instance](https://github.com/kind2-mc/kind2/blob/develop/src/lustre/lustreParser.mly#L1063) where we use `int_of_string`. 

Do you think we will have more than 2^62 sized tuples :)

src/lustre/lustreTypeChecker.ml
apoorvingle: Fixed.

tests/lustre/typechecking/futurePass/test_add_tuples.lus
apoorvingle: Fixed.

src/lustre/lustreParser.mly
daniel-larraz: > I see that there is another instance where we use int_of_string.

Good. Then you can fix that instance too.

> Do you think we will have more than 2^62 sized tuples :)

No, I don’t think so. That is why I suggested two alternatives instead of asking you to use a Numeral :)
However, I do think it is possible a scenario where somebody mistakenly uses a big number to index a tuple (wrong constant, bad transcompilation,...) and I would like Kind 2 to return an informative message instead of `Failure "int_of_string"`. At the bare minimum we should use a code like this one:

```
{ let idx =
    try int_of_string i 
    with _ ->
      LustreContext.fail_at_position
        (mk_pos $startpos(i)) "Tuple projection index exceeds int range"
  in
  A.TupleProject (mk_pos $startpos, e, idx) 
}

```

apoorvingle: sounds good. 
Fixed.

tests/lustre/syntax-errors/array-syntax.lus
```diff
@@ -1,5 +1,5 @@
 node slide (const n:int; s:int) returns(A: int^n);
 let
     A[i] = if i = 0 then s else (-1 -> pre A[i-1]);
-    %PROPERTY n > 1 => (true -> A[1] = pre s);
+    --%PROPERTY n > 1 => (true -> A[1] = pre s);
```
daniel-larraz, line 4: What is the point of this test if there is no syntax error after the fix?

tests/lustre/syntax-errors/contract-error.lus
```diff
@@ -0,0 +1,12 @@
...
+  guarantee out = (in > 0.0);
+tel
+
+node bli(in: real) returns (out: bool) ;
+(*@contract
+  import bla_spec (in) returns (out+out) ;
+*)
+let
+  out = in >= 0.0 ;
+tel
```
daniel-larraz, line 12: Why is this test in the **syntax-error directory** if Kind 2 returns a type error?

tests/regression/error/test-func-sliced.lus
```diff
@@ -17,21 +17,26 @@ function sincos(in: real) returns (sin: real; cos: real);
     ensure sin <= 0.0;
   );
 *)
+let
+   sin = in;
+   cos = in;
+tel
+
```
daniel-larraz, line 24: Remove this body and set the function as imported:
function **imported** sincos(in: real) returns (sin: real; cos: real);

tests/regression/error/test-func-sliced.lus
```diff
@@ -17,21 +17,26 @@ function sincos(in: real) returns (sin: real; cos: real);
...
 node X (in: real) returns (OK: bool);
 var inn: real ;
 let
   inn = in ; 
   OK =
     exp(0.0 -> inn, 1.0 -> inn) > 0.0 or
-    sincos(0.0 -> inn + pre inn) + sincos(0.0 -> inn + pre inn) = (0.0, 0.0);
+    sincos(0.0 -> inn + pre inn) + sincos(0.0 -> inn + pre inn) = [0.0, 0.0];
```
daniel-larraz, line 31: Revert this change, keep the tuple for now.

tests/regression/error/test-func-sliced.lus
```diff
@@ -48,4 +53,11 @@ function exp (x, y: real) returns (out: real);
...
    guarantee out <= sin(out);
 *)
+let
+out = x + y;
+tel
+
 function sin(in: real) returns (out: int);
+let
+   out = (int in);
+tel
```
daniel-larraz, line 63: Remove this body and set the function as imported:
function **imported** sin(in: real) returns (out: int);

tests/regression/error/test-func-sliced.lus
```diff
@@ -48,4 +53,11 @@ function exp (x, y: real) returns (out: real);
    guarantee out >= id(out);
    guarantee out <= sin(out);
 *)
+let
+out = x + y;
+tel
+
```
daniel-larraz, line 59: Remove this body and set the function as imported:
function **imported** exp (x, y: real) returns (out: real);
Replace `(* -- contract` by `(*@ contract`

tests/lustre/syntax-errors/contract-error.lus
apoorvingle: `import bla_spec (in) returns (out+out);`
This should be a syntax error... 
I get this error 
```
<Error> line 8 col. 36: Failure cause: Syntax Error! Perhaps an illegal expression in contract call.
```

PR #683, opened 2020-10-10
Removed the UBV symbol and combined BV and UBV terms in the backend.

arjunvish:


src/terms/term.ml
```diff
@@ -304,38 +285,13 @@ let bitvector_of_term t = match destruct t with
...
       Symbol.bitvector_of_symbol s
 
-  | T.Const s when Symbol.is_ubitvector s -> 
-      Symbol.ubitvector_of_symbol s
-
-  | T.Const s when Symbol.is_numeral s ->
-      Bitvector.num_to_ubv64 (Symbol.numeral_of_symbol s)
+  (*| T.Const s when Symbol.is_numeral s ->
+      Bitvector.num_to_ubv64 (Symbol.numeral_of_symbol s)*)
   (* For Yices 1 native *)
```
daniel-larraz, line 290: As the comment says, the last pattern-matching case that you commented out is for supporting Yices 1 native. Please keep the code enabled.

arjunvish: Wasn't sure whether the comment was for the case after or before :).

PR #694, opened 2020-12-08
Qualified names for Lustre Identifiers

apoorvingle:
Features
-------

- AST identifiers are now qualified
- Modes are stored as a qualified identifier and not a list of strings
- ghost variables and constants of imported contracts can be accessed by a qualified manner.
- Contracts can be imported using `as` construct i.e. aliased imports.

### DIFF
--- a/src/lustre/lustreLexer.mll
+++ b/src/lustre/lustreLexer.mll
@@ -334,7 +334,10 @@ let keyword_table = mk_hashtbl [
    C syntax: alphanumeric characters including the underscore, starting 
    with a letter or the underscore *)
 let id = ['a'-'z' 'A'-'Z' '_'] ['a'-'z' 'A'-'Z' '_' '0'-'9']*
-         
+
+let qual_id = id (':' ':' id)*
+(* a qualified id is an id or id::qual_id *)
+                
 (* Keep these separated from alphabetic characters, otherwise a->b would 
    be one token *)
 let printable = ['+' '-' '*' '/' '>' '<' '=' ]+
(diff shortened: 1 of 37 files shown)


src/utils/lib.mli
```diff
@@ -63,6 +63,9 @@ val get : 'a option -> 'a
 (** Return the min between two optional floats. Return None if both floats are None. *)
 val min_option : float option -> float option -> float option
 
+(** Maps a function over an optional value *)
+val map_option: ('a -> 'b) -> 'a option -> 'b option 
+
```
daniel-larraz, line 68: Mention this function is available in version 4.08 and later of the standard library:
https://caml.inria.fr/pub/docs/manual-ocaml/libref/Option.html

src/lustre/lustreLexer.mll
```diff
@@ -334,7 +335,10 @@ let keyword_table = mk_hashtbl [
    C syntax: alphanumeric characters including the underscore, starting 
    with a letter or the underscore *)
 let id = ['a'-'z' 'A'-'Z' '_'] ['a'-'z' 'A'-'Z' '_' '0'-'9']*
-         
+
+let qual_id = id (':' ':' id)*
+(* a qualified id is an id or id::qual_id *)
+
```
daniel-larraz, line 341: A qualified identifier shouldn’t be a single token, but a grammar rule. This forces the parser to deconstruct the qualified identifier unnecessarily.

src/utils/lib.mli
apoorvingle: yup. but we are using 4.07 right? (:

daniel-larraz: I meant please write a comment to future developers.

apoorvingle: Ah, i misread your comment..

PR #699, opened 2020-12-17
group type support

apoorvingle:
Supporting group type in dependency analysis and type checking
- Groups are like tuples but flattened
- Functions and contracts have type `GroupType -> GroupType` instead of `TupleType -> TupleType` 
- Adjust graph creation for groups especially for temporal operators -- condact, arrow, restart etc.


src/ivcMcs.ml
```diff
@@ -149,6 +149,7 @@ let rec unannot_pos = function
   | A.UserType (_,id) -> A.UserType (dpos,id)
   | A.AbstractType (_, id) -> A.AbstractType (dpos,id)
   | A.TupleType (_,ts) -> A.TupleType (dpos, List.map unannot_pos ts)
+  | A.GroupType (_,ts) -> A.TupleType (dpos, List.map unannot_pos ts)
```
daniel-larraz, line 152: Shouldn’t the rhs be a GroupType?

> A.GroupType (_,ts) -> **A.GroupType** (dpos, List.map unannot_pos ts)

src/lustre/lustreAstDependencies.ml
```diff
@@ -478,7 +481,7 @@ let rec extract_decls: ('a IMap.t * id_pos_map) -> LA.ident list -> ('a list) gr
   | i :: is ->
      (match (IMap.find_opt i decl_map) with
       | None -> (match (find_id_pos i_pos_map i) with
-                 | None -> failwith ("Identifier " ^ i ^ " not found. This should not happen")
+                 | None -> fail_no_position ("Identifier " ^ i ^ " not found. This should not happen")
```
daniel-larraz, line 484: Why did you replace `failwith` with `fail_no_position` which has the side-effect of printing a message? One of the goals of the new design is to make the code free of side effects so that messages are only printed at the top-level.

daniel-larraz: All `failwith` exceptions should be caught within the module and be notified through the error part of the `graph_result`

src/ivcMcs.ml
apoorvingle: oops. Fixed.

src/lustre/lustreAstDependencies.ml
apoorvingle: Getting a None here is a panic state and we do not expect it to happen.
currently `failwith` raises an obscure error which gets caught in `Kind2.ml` generic exception handler.
I had changed it to `fail_no_pos` so that it fails fast but prints a more pretty error.

I have changed it to `graph_error` now.

tests/lustre/typechecking/shouldPass/test_const_prop.lus
```diff
@@ -17,7 +17,7 @@ const f = arr[fc]; --  should not inline
 const f2 = arr[1]; -- should inline to false
 const f3 = [true, true, true][2];
 
-const t3:[int, int, bool] = (t, 3, true); -- should inline to (1, 3, true)
+const t3:[int, int, bool] = {t, 3, true}; -- should inline to (1, 3, true)
 const arr3 = [1, 1+1, t1+c, 5]; -- inline all t's and simplify
 
 const p = t3.%1; -- should inline to 3
```
daniel-larraz, line 23: Add a dummy node at the bottom of the file so that Kind 2 does not complain about:

> No main node defined in input

daniel-larraz: I remembered you granted me permission to modify your fork. I will fix it myself.

PR #723, opened 2021-03-25
Two small array bugs

amosr:
Hi Daniel,

I'm looking into some bugs in the array support at the moment. I have some fixes for two small bugs. I'm looking into another one now, which is that you can't call nodes or functions inside an array definition like `x[i] = f(y[i])`. This one seems a bit hairier though.

## Array indexing reuse:
Creating a new state var for an array select would reuse the existing
definition, but with the array indices from the environment rather
than what the select expression used.

The nested select in the following equation:
```
  prevs2[i] = if i = 0 then ins else (0.0 -> pre (prevs2[i - 1]));
```
was translated as `prevs2[i]`, somehow losing the `i-1`.

## Array constructor typechecking:
Array constructors (of the form `x^n`) were causing type errors in
arrows.

For a definition like:
```
const_zero = 0.0^5 -> pre const_zero;
```

This was being simplified to the following in LustreExpr:
```
abs_0 = 0.0;
const_zero[i0] = (abs_0 -> pre const_zero);
```

But this is a type error, because one side of the arrow is a real
and the other is an array. This commit inserts array selects to
make the types line up like so:
```
abs_0 = 0.0;
const_zero[i0] = (abs_0 -> pre const_zero[i0]);
```

### DIFF
--- a/src/lustre/lustreContext.ml
+++ b/src/lustre/lustreContext.ml
@@ -1051,9 +1051,10 @@ let fresh_state_var_for_expr
 
   else if reuse && E.is_select_array_var expr && present_bounds <> [] then
 
-      let v = E.var_of_array_select expr in
+      let (v, ixes) = E.indexes_and_var_of_array_select expr in
       let sv = Var.state_var_of_state_var_instance v in
-      (sv, present_bounds), ctx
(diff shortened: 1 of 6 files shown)


daniel-larraz:
Hi Amos. Thank you very much for the fixes!

PR #724, opened 2021-03-26
Array support for function and node calls

amosr:
Hi Daniel,

A little bit more array stuff: support calling functions with quantified array arguments, such as:
```
x[i] = f(y[i]);
```

This is implemented by expanding the array to one call for each element
of the array. This only works for arrays of known constant size. It effectively desugars the above into:

```
x[i] = if i = 0 then f(y[0]) else if i = 1 then f(y[1]) else …;
```

This translation isn't ideal, but for the use cases I am interested in, I expect to only really deal with arrays of 5 or maybe 10 elements at most. One neat thing is that the existing reuse logic still works for degenerate cases like `x[i] = f(a)` where the call's input arguments do not depend on the array index, so it does not introduce any extra calls there.

I guess that the 'right' way to implement this would be to somehow extend the node_call to keep track of quantified array bounds there. This seemed like too large a change for me to tackle right now. I think that this translation is a simple trade-off that gives most of the benefits, but let me know if there's a better way to do it.

Amos


daniel-larraz:
Hi Amos. Thank you for your contribution.

Issue #707, opened 2021-01-18
Type error with subrange int in arrays


Vivi-yd:
Hi,
I am getting a type-checking error that seems to be a bug. Any idea what is causing it?

`theta_in` is a percentage and `angles` is an array of percentages. 

```
type percent = subrange [0, 100] of int;
node test_array2( 
    theta_in: percent; 
) returns (
    result: bool;
)
var 
    angles    : percent^3;
let

    angles = [theta_in, theta_in, theta_in] -> [theta_in, pre (angles[0]), pre (angles[1])];
    result = false;
tel
```
Even though both should be integers, we get this error:

```
Failure cause: Type mismatch in equation: int and int 
```

Whereas this one works fine:
```
type percent = subrange [0, 100] of int;
node test_array2( 
    theta_in: percent; 
) returns (
    result: bool;
)
var 
    angles    : percent^3; 
let
    angles = [theta_in, theta_in, theta_in] -> [theta_in, (pre angles)[0], (pre angles)[1]];
    result = false;
tel
```
Thanks!
Vivian

Vivi-yd:
Found a different error, might be related to the issues mentioned above:
```
type percent = subrange [0, 100] of int; 

node test_invalid_arg(
    theta_in: percent;

) returns (
    angles    : percent^1; 
)
let
    angles = [theta_in] ->
    [2 * (pre angles)[0]]; 
tel
```

This code produce an invariant manager error:
```
Runtime error in invariant manager: File "src/terms/term.ml", line 2105, characters 12-18: Assertion failed
```

Issue #721, opened 2021-03-19
Soundness bug with subrange types


xaaronc:
With the following Lustre program:

```
type T = subrange [0,1] of int;

function B(x: T) returns (y: T)
let
    y = x;
tel

function A(x: int) returns (y, z: T)
let
    y = x;
    z = B(y);
tel
```

kind2 correctly identifies the bounds violation in the `y = x` assignment when using `--slice_nodes true`:
```
$ kind2 --slice_nodes true foo.lus
[..]
<Failure> Property y.bound is invalid by property directed reachability for k=0 after 0.051s.
```
but with `--slice_nodes false`, it succeeds erroneously:
```
$ kind2 --slice_nodes false foo.lus 
[..]
<Success> Property y.bound is valid by property directed reachability after 0.060s.
```

Version `kind2 36140a5a`

xaaronc:
I’ve been able to reproduce this bug with the default setting for `--slice_nodes`, though so far haven’t been able to produce a small example input.

xaaronc:
Found a simple reproducer:

```
type U = subrange [0,2] of int;
type T = subrange [-1,1] of int;

node A(x: U) returns (y: U)
let
    y = x;
tel

node B(x: T) returns (z: T; y: U)
let
    y = x;
    z = A(y);
tel
```

```
$ kind2 foo2.lus
[..]
<Success> Property z.bound is valid by property directed reachability after 0.050s.
<Success> Property y.bound is valid by property directed reachability after 0.050s.
```

daniel-larraz:
Hi Aaron,

Thank you for reporting this bug and providing the small reproducer. It seems the current mechanism for checking subrange bounds is broken. I will review it and make the appropriate changes to fix it, but it will take some time. I'll keep you posted.

Thanks,
Daniel

Issue #701, opened 2020-12-21
Abstract nodes forget that their return types are subranges


amosr:
Hi Daniel, hope you're well.

We found a small bug where the abstract node ignores the subrange information. If you have a node that returns a subrange int and give it a contract, then when you use that node, the abstract version can return all ints rather than just the subrange. It's not a big deal, and the workaround is just to add the constraints manually.

Here is an example:

```
-- test_bounds.lus
-- This example shows where abstract nodes do not respect the declared subranges.
type SUB = subrange [0, 100] of int;

-- Node that returns a subrange type with a very boring contract:
node subrange_contract(i : SUB) returns (o : SUB)
(*@contract
  guarantee true;
*)
let
  -- The trick here is just to make it less obvious that it's within the range [0,100]
  -- (but it is within the range, and Kind2 can prove it)
  o = i / (if i <= 0 then 1 else i);
tel

-- Use of the subrange. Try verifying with:
--$ kind2 --modular true --compositional true test_bounds.lus --lus_main eg1_fails
-- We get the following failure:
--
-- Analyzing eg1
--   with First top: 'eg1'
--              subsystems
--                | abstract: subrange_contract
--              assumptions:
--                subrange_contract: 2 one-state, 0 two-state
-- 
-- <Failure> Property b.bound is invalid by property directed reachability for k=0 after 0.035s.
-- 
-- Counterexample:
--   Node eg1 ()
--     == Inputs ==
--     i -1
--     == Outputs ==
--     a  1
-- 
--   Node subrange_contract (eg1[l16c6])
--     == Inputs ==
--     i   1
--     == Outputs ==
--     o 101
--
--  Note that the output "o" for subrange_contract is not within the range [0,100].
node eg1_fails(i : int) returns (a, b : SUB)
let
  a = 1;
  b = subrange_contract(a) * ((if i <= 0 then 1 else i) / (if i <= 0 then 1 else i)) ;
tel


-- For now, we can get away with adding explicit ranges on the contract:
node subrange_contract_explicit_range(i : SUB) returns (o : SUB)
(*@contract
  assume 0 <= i and i <= 100;
  guarantee 0 <= o and o <= 100;
  guarantee true;
*)
let
  o = i / (if i <= 0 then 1 else i);
tel

-- Verifying this one passes, as expected.
node eg2_works(i : int) returns (a, b : SUB)
let
  a = 1;
[...]

PR #725, opened 2021-03-29
Fix subrange revert to ints inside arrays

amosr:
Hi Daniel,

More small array fixes.

When defining an array where the expression has a subrange type, if
the bounds are not obviously satisfied then the type is reverted to
`int` and the bounds are added as properties. However, for arrays
this reversion was replacing the whole type with `int` rather than
just the subrange - so `(subrange [0,5] of int)^10` had its type
changed to just `int` rather than `int^10`.

I also added some pretty-printing functions, which I found useful
while debugging.

This fixes issue #707.

### DIFF
--- a/src/lustre/lustreContext.ml
+++ b/src/lustre/lustreContext.ml
@@ -1931,12 +1931,21 @@ let add_node_equation ctx pos state_var bounds indexes expr =
                | E.Unbound _ ->
                  if Type.is_array t then Type.elem_type_of_array t
                  else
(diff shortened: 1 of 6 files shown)


daniel-larraz:
Hi Amos. I'm currently working on rewriting the code that checks bounds for subrange streams. The new code will fix this issue, #701, and #721. I will publish it soon.

amosr:
Hi Daniel, that's great! Thanks

PR #729, opened 2021-03-30
Add span information to declarations and file name to program

amarmaduke:
For now span information is added to the declarations of a Lustre program (for outlining in the LSP under development) and the file name is recorded at the top level (replacing the top level list of declarations with a record, also for the LSP under development).

### DIFF
--- a/src/inputSystem.ml
+++ b/src/inputSystem.ml
@@ -27,7 +27,7 @@ module SVS = SVar.StateVarSet
 module SVM = SVar.StateVarMap
 
 type _ t =
-| Lustre : (LustreNode.t S.t * LustreGlobals.t * LustreAst.declaration list) -> LustreNode.t t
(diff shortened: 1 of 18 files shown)


amarmaduke:
Currently waiting on reviewing how positions are being filled in to the position record (are they relative? absolute? what exactly is happening in the old code?) We want to know so that the LSP can construct absolute paths to positions in included files.

amarmaduke:
The user supplied string to the `include` directive is what ends up in the `pos_fname` field of the `Lib.position` record.

The code to find the file is: `lustreLexer.mll`:408-450, in particular this snippet: `Lib.find_file p (include_curdir :: Flags.include_dirs ())` which will fail if the file can't be found. This means that constructing a absolute path to the file should always be achievable with the current directory, the supplied relative path, and any included directories from `include_dirs`.

PR #731, opened 2021-03-31
Add a Lustre AST Normalizer pass

amarmaduke:
What the normalizer is suppose to accomplish is commented at the top of the interface file.

At the moment, the actual pass in `lustreInput` is commented out because the normalizer modifies the AST in such a way that if the code that compiles the AST to a `lustreNode` does not incorporate the `generated_identifiers` information it wont work.

### DIFF
--- a/src/lustre/lustreAstNormalizer.ml
+++ b/src/lustre/lustreAstNormalizer.ml
@@ -1,0 +1,235 @@
+(* This file is part of the Kind 2 model checker.
+
+   Copyright (c) 2020 by the Board of Trustees of the University of Iowa
+
+   Licensed under the Apache License, Version 2.0 (the "License"); you
+   may not use this file except in compliance with the License.  You
+   may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0 
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+   implied. See the License for the specific language governing
+   permissions and limitations under the License. 
+
+ *)
+(** Normalize a Lustre AST to ease in translation to a transition system
+
+  The two main requirements of this normalization are to:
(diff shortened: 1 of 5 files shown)


src/lustre/lustreAstNormalizer.ml
```diff
@@ -0,0 +1,346 @@
...
+
+open Lib
+
+module A = LustreAst
+
+(** Map for types with identifiers as keys *)
+module IMap = struct
+  include Map.Make(struct
+    type t = LustreAst.ident
+    let compare i1 i2 = Stdlib.compare i1 i2
```
daniel-larraz, line 61: Do not use the polymorphic compare function `Stdlib.compare`, it is expensive. Use `String.compare` instead.

src/lustre/lustreAstNormalizer.ml
```diff
@@ -0,0 +1,346 @@
...
+  let gids = { locals = IMap.empty; oracles = [name] } in
+  nexpr, gids
+
+let expr_is_id : A.expr -> bool = function
+  | Ident (_, _) -> true
+  | _ -> false
+
+let expr_is_const : A.expr -> bool = function
+  | Const (_, _) -> true
+  | _ -> false
```
daniel-larraz, line 127: Move these functions to `lustreAstHelpers` so that they can be reused.

src/lustre/lustreAstNormalizer.ml
```diff
@@ -0,0 +1,346 @@
...
+
+and normalize_equation = function
+  | Assert (pos, expr) ->
+    let nexpr, gids = normalize_expr expr in
+    Assert (pos, nexpr), gids
+  | Equation (pos, lhs, expr) ->
+    let nexpr, gids = normalize_expr expr in
+    Equation (pos, lhs, nexpr), gids
+  | Automaton (pos, id, states, auto) ->
+    A.Automaton (pos, id, states, auto), empty
```
daniel-larraz, line 202: Automaton requires normalization, it contains Lustre expressions too. Use a `Lib.todo __LOC__` call if you do not want to implement this now for some reason.

PR #737, opened 2021-04-08
Interpreter accepts fractions in JSON input

daniel-larraz:


doc/usr/source/9_other/8_interpreter.rst
```diff
@@ -33,7 +33,7 @@ The overall structure is as follows:
     {
       "var1": "24",
       "var2": false,
-      "var3": "0.5"
```
abdoo8080, line 36: Instead of replacing floating point with fractions, why not add a fourth variable?

doc/usr/source/9_other/8_interpreter.rst
```diff
@@ -33,7 +33,7 @@ The overall structure is as follows:
...
+      "var3": "1/2"
```
abdoo8080, line 36: Shouldn't this be `1.0/2.0`?

abdoo8080 (approved):
LGTM! Left few comments.

doc/usr/source/9_other/8_interpreter.rst
daniel-larraz: We are showing input values for two steps. I replaced one floating point instance with a fraction instance.

doc/usr/source/9_other/8_interpreter.rst
daniel-larraz: We were supporting both in the CSV format, so I kept both in the JSON format.

abdoo8080: I see. I prefer consistency with Lustre though.

daniel-larraz: Ok. I will update the documentation after the merge.

PR #740, opened 2021-04-22
First pass for new node generation

amarmaduke:
This node generation pass does not include structured data or Daniel's new properties generated for subranges or contracts. The structured data code is already written but I wasn't able to test it yet and having a smaller surface to review seemed like a good idea anyway.

Normalization needed to change slightly to record the node where the ids where generated. It may make sense to annotate the generated ids with additional information (such as whether it should be constant).

~~One thing I am unsure about: Node Calls in the old code will abstract new state variables for outputs and oracles and create equations to the effect of `abs_0 = o1` (where `abs_0` is the new abstraction and `o1` is the old state variable for the node output). It is not clear to me why this is the case or needed and in the current code I do not do this abstraction.~~

### DIFF
--- a/src/lustre/lustreAstNormalizer.ml
+++ b/src/lustre/lustreAstNormalizer.ml
@@ -77,10 +77,12 @@ end
 
 type generated_identifiers = {
     node_args : (string * LustreAst.expr) list;
-    pre_args : (string * LustreAst.expr) list;
-    oracles : (LustreAst.ident * LustreAst.expr) list;
+    locals : LustreAst.expr StringMap.t;
+    oracles : (string * LustreAst.expr) list;
+    propagated_oracles : (string * string) list;
     calls : (Lib.position (* node call position *)
-      * (string list) (* abstraction variables *)
+      * (string list) (* oracle inputs *)
+      * (string list) (* abstracted inputs *)
       * LustreAst.expr (* condition expression *)
       * LustreAst.expr (* restart expression *)
       * string (* node name *)
@@ -90,6 +92,7 @@ type generated_identifiers = {
 }
 
 let pp_print_generated_identifiers ppf gids =
+  let locals_list = StringMap.bindings gids.locals in
   let pp_print_local = pp_print_pair
     Format.pp_print_string
     A.pp_print_expr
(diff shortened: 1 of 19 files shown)


src/utils/lib.ml
```diff
@@ -404,6 +404,22 @@ let list_join equal l1 l2 =
     (* Call recursive function with initial accumulator *)
     | _ -> list_join' equal [] l1 l2
 
+let list_filter_map f =
+  let rec rev_append l1 l2 =
+    match l1 with
+      [] -> l2
+    | a :: l -> rev_append l (a :: l2)
+  in
+  let rev l = rev_append l [] in
```
daniel-larraz, line 413: `List.rev` is available in OCaml 4.07

src/lustre/lustreExpr.ml
```diff
@@ -3278,11 +3284,40 @@ let mk_arrow expr1 expr2 =
 
 
 (* Pre expression *)
-let mk_pre mk_abs_for_expr mk_lhs_term ctx unguarded
+(* We don't need to abstract the RHS because the AST normalization pass 
+  already has. *)
+let mk_pre mk_lhs_term ({ expr_init; expr_step; expr_type } as expr) =
```
daniel-larraz, line 3289: First argument is not used.

src/lustre/lustreExpr.ml
```diff
@@ -233,6 +233,12 @@ let pre_offset = Numeral.(pred cur_offset)
 (* Pretty-printing                                                        *)
 (* ********************************************************************** *)
 
+(** Pretty-print a bound or fixed annotation *)
+let pp_print_bound_or_fixed pp ppf = function
+  | Bound x -> pp ppf x
+  | Fixed x -> pp ppf x
+  | Unbound x -> pp ppf x
```
daniel-larraz, line 240: It seems that `pp` is always `LustreExpr.pp_print_expr`. Use that function directly.

src/lustre/lustreAstNormalizer.ml
```diff
@@ -48,194 +50,383 @@
...
     (pp_print_list pp_print_oracle "\n") gids.oracles
     (pp_print_list pp_print_local "\n") locals_list
+    (pp_print_list pp_print_local "\n") gids.node_args
+    (pp_print_list pp_print_call "\n") gids.calls
 
 
 (* [i] is module state used to guarantee newly created identifiers are unique *)
 let i = ref 0
 
+let is_input_const = ref StringMap.empty
```
daniel-larraz, line 142: You should reset `is_input_const` at the beginning of `normalize` so that the function can be called more times if needed.

daniel-larraz: Well, since `normalize` is a recursive function right now, you may need to split it into two first.

amarmaduke: I don't think it is needed because `is_input_const` is only tracking if the the inputs to a node are constant (information I hoped could be extracted from the type checker context, but it is not being tracked). This information should be static across calls to `normalize`

amarmaduke: Although it would be better to write an auxiliary function that populates it once and only once

daniel-larraz: > This information should be static across calls to normalize

Provided that the input program is the same.

Although the current pipeline reads only one input program and then process it, I would like to have the flexibility of compiling different lustre programs in the same run of the tool.

amarmaduke: Gotcha, will do

src/lustre/lustreAstNormalizer.ml
```diff
@@ -48,194 +50,383 @@
...
+      (match output_type with
+      | A.GroupType (_, list) -> List.length list
+      | _ -> 1)
+    | _ -> assert false (* Nodes must have TArr type *)
+
+let unwrap result = match result with
+  | Ok x -> x
+  | Error (p, s) -> 
+    Format.eprintf "%s\n" s;
+    assert false
```
daniel-larraz, line 219: There is already an `unwrap` function in the `Res` module.

PR #752, opened 2021-05-28
Allow checking multiple top-levels in modular mode

amosr:
Hi Daniel,

We have a use-case where we have multiple top-level nodes that are
called by different pieces of C code. These top-level nodes share
some common utility functions, so re-checking them separately adds
some extra time.

This PR adds the ability to annotate multiple nodes with `--%MAIN;`
and check all of the transitively reachable nodes just once.
I've restricted this to only work in modular mode, because
non-modular mode will re-check the properties multiple times anyway.

If you think this restriction to only work in modular mode is too severe, I can probably figure out how to make non-modular mode work. I think it's a little more effort as I'd need to synthesise a definition that uses all of the top-level nodes.

### DIFF
--- a/src/lustre/lustreInput.ml
+++ b/src/lustre/lustreInput.ml
@@ -186,11 +186,11 @@ let of_channel in_ch =
   in
     
     (* Name of main node *)
-    let main_node = 
+    let main_nodes =
       (* Command-line flag for main node given? *)
       match Flags.lus_main () with 
       (* Use given identifier to choose main node *)
-      | Some s -> LustreIdent.mk_string_ident s
+      | Some s -> [LustreIdent.mk_string_ident s]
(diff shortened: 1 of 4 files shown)


daniel-larraz:
Hi Amos,

Thank you for your PR. Your proposed feature seems useful to us, but we prefer a solution that works across all the functions provided by Kind 2. I created a new PR (#756) that lifts your PR's restriction. Unfortunately, I couldn't find a way to reuse your synthetic-main-node solution in all cases, so I changed the Lustre inputSystem data structure instead.

Let me know if my PR works fine for you.

Thanks,
Daniel

amosr:
Hi Daniel,
Thanks for this, I just tested out your PR and it works great.

Amos

PR #785, opened 2021-09-21
Misc error checks

amarmaduke:


daniel-larraz:
Hi @amarmaduke. The check for node calls to nodes with a contract in contracts need to be refined. It used to be that way, but it was changed recently to accept two cases: a node call to an imported node, and a node call to a node with a contract that **only** contains assumptions. Same rule applies to functions. The main point is: calls to nodes/functions that are not subject to refinement during analysis should be allowed, which are basically the ones described above. I added some [tests](https://github.com/kind2-mc/kind2/commit/918a881895881af4b4b281bb76099aae7621d439) to develop which cover the different cases.

daniel-larraz:
BTW, I think there is a bug in the new translator because [one](https://github.com/kind2-mc/kind2/blob/918a881895881af4b4b281bb76099aae7621d439/tests/experimental/success/valid_node_calls_in_contract.lus) of the new tests is classified as falsifiable with the new translator but all properties are valid.

amarmaduke:
To confirm, this means that a contract for a node can have ghost variables, ghost constants, imports, and assumptions, and be fine? (Assuming that the imports are all also consisting of only ghost variables, ghost constants, imports, and assumptions?)

That is how I have currently relaxed it.

daniel-larraz:
> To confirm, this means that a contract for a node can have ghost variables, ghost constants, imports, and assumptions, and be fine? (Assuming that the imports are all also consisting of only ghost variables, ghost constants, imports, and assumptions?)
> 
> That is how I have currently relaxed it.

That is right. Thanks.

PR #789, opened 2021-10-01
Terminate the log before exiting with the new front end.

abdoo8080:
#695 for the new front-end.


daniel-larraz:
I created a new PR (#790) that solves the issue this PR aims to fix, but terminating the log in the kind2 (main) module.

Issue #538, opened 2018-07-26
Nodes declaration order in Contracts


hbourbouh:
Hi Daniel,
It seems that the order of nodes is important for contracts, any node used by a contract should be declared before it in the lustre file.
Example:
`
node mod_real(x, y: real;)
returns( z: real );
(*@contract
 guarantee abs_real(z) < abs_real(y); 
 *)
let
tel

node abs_real (x: real)
returns(y:real);
let
	y= if x >= 0.0 then x 
	else -x;
tel
`
Kind2 error: unknown node or function "abs_real"
Source: parse 

Is it possible to fix this in Kind2, or it's for implementation reasons we can not?

Thanks,
Hamza

daniel-larraz:
Hi Hamza,

It is a technical limitation of the current implementation. We have plans to address this issue and other ones related to contracts in a future rewriting of that part of the code.  

Best,
Daniel

PR #809, opened 2021-11-04
Improve performance by switching to HStrings immediately (at parsing time)

amarmaduke:
The bad news:

This change only shaves off about 2 seconds, leaving the old front-end at 0.02s and the new front-end at ~2s for the `cruise_controller` example

The good news:

After some experimentation and profiling, I decided to try commenting out the dependency sorting passes (there are only two). If no dependency sorting is done at all, the new front-end is **as fast as the old front end** (for parsing). This means that all of the remaining slow down is in the dependency sorting code.

Also, there are places in the code base where `Stdlib.compare` is used that have nothing to do with either front-end. I did not change these, should they be changed?

### DIFF
--- a/src/ivcMcs.ml
+++ b/src/ivcMcs.ml
@@ -181,7 +181,8 @@ let undef_expr pos_sv_map const_expr typ expr =
         let n = (List.length typ) in
         if n > !max_nb_args then max_nb_args := n ;
         A.Call(*Param*)
-          (pos, rand_function_name_for n typ, (*typ,*) [Const (dpos, Num (string_of_int i))])
+          (pos, HString.mk_hstring (rand_function_name_for n typ),
+            (*typ,*) [Const (dpos, Num (HString.mk_hstring (string_of_int i)))])
       end else begin
         try Hashtbl.find previous_rands svs
         with Not_found ->
@@ -189,7 +190,8 @@ let undef_expr pos_sv_map const_expr typ expr =
           let n = (List.length typ) in
           if n > !max_nb_args then max_nb_args := n ;
           let res = A.Call(*Param*)
-            (pos, rand_function_name_for n typ, (*typ,*) [Const (dpos, Num (string_of_int i))])
+            (pos, HString.mk_hstring (rand_function_name_for n typ),
+              (*typ,*) [Const (dpos, Num (HString.mk_hstring (string_of_int i)))])
           in Hashtbl.replace previous_rands svs res ; res
       end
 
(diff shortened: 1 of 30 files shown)


daniel-larraz:
> Also, there are places in the code base where `Stdlib.compare` is used that have nothing to do with either front-end. I did not change these, should they be changed?

There are still a couple of places that are related to the old/new front-ends where `Stdlib.compare` is used:
lustre/lustreAst.ml
lustre/lustreAstHelpers.ml
lustre/lustreIndex.ml
lustre/lustreIdent.ml
utils/ident.ml
utils/ident.mli

Please address just the one above. We will address the rest in a later PR. Thanks.

amarmaduke:
I didn't see any `Stdlib.compare` in `utils/ident.ml`

daniel-larraz:
> I didn't see any `Stdlib.compare` in `utils/ident.ml`

You are right. I had `utils/ident.ml` in my notes because of a different question. Since `LustreIdent` depends on `Ident`, shouldn't we change the type of `Ident.t` from `string` to `HString`?

amarmaduke:
I held off on changing `Ident.t` because its use doesn't show up in the new front end parsing phase and it would cause `HString` changes further down the stack (like in the transition systems). I guess the leads to the larger question of should HStrings just be used _everywhere_? Including in StateVar etc?

daniel-larraz:
> I held off on changing `Ident.t` because its use doesn't show up in the new front end parsing phase and it would cause `HString` changes further down the stack (like in the transition systems). 

Fair enough. Let's keep the `string` type then.

> I guess the leads to the larger question of should HStrings just be used _everywhere_? Including in StateVar etc?

We should think about this at some point, but clearly this is out of scope of this PR.

apoorvingle:
A bit too late; but as a user of the system, 
it may be beneficial to have a flag for enabling disabling dependency analysis, rather than disabling it entirely.

This affects VS code extension users and from what I understand they use the nightly build instead of the stable version.

daniel-larraz:
This PR does not disable dependency analysis. That was only a local test.

apoorvingle:
Ah!
So was it my past (dumb-ass) self who never enabled reordering of contracts and nodes in the dependency analysis? :)

daniel-larraz:
> Reason being I bumped myself into #538, which i thought I did solve.

The dependency analysis you developed for the new front-end handles correctly the example in #538. However, the issue remains open because Kind 2 still uses the old front-end by default (`--old_frontend true`). It will be closed once the new front-end becomes the default option. The new front-end is only used for now in the VS Code extension for providing language support in the editor.

daniel-larraz:
I just tested it. The model is parsed correctly in the editor. Only when the Check button is pressed, an error occurs. This is the expected behavior until both editor and checker use the same front-end.

apoorvingle:
okay, good to know! Thanks for taking a look.
(In hindsight I could have checked this myself. :))

PR #818, opened 2021-11-22
Add abstract interpretation to infer tighter subrange bounds of outputs

amarmaduke:


daniel-larraz:
The inferred subrange properties are not classified as candidate properties but as normal properties. For instance, compare the output of the old front-end and the new front-end on the `tests/experimental/success/infer_tight_subrange.lus` example.

src/lustre/lustreAbstractInterpretation.ml
```diff
@@ -260,9 +268,15 @@ and interpret_eqn node_id ctx ty_ctx lhs rhs =
...
         let ty1 = Ctx.expand_nested_type_syn ty_ctx ty1 in
-        let ty2 = interpret_expr_by_type node_id ctx ty_ctx ty1 p expr in
-        let ty = restrict_type_by ty1 ty2 in
-        add_type acc node_id id ty
+        let restrict_ty = interpret_expr_by_type node_id ctx ty_ctx ty1 p expr in
+        let ty, is_restricted = restrict_type_by ty1 restrict_ty in
+        Format.eprintf "id: %a@. ty1: %a@. restrict_ty: %a@."
+          HString.pp_print_hstring id
+          LA.pp_print_lustre_type ty1
+          LA.pp_print_lustre_type restrict_ty;
```
daniel-larraz, line 276: You forgot to disable this debugging message :)

daniel-larraz:
For node `N3` of  `tests/experimental/success/infer_tight_subrange.lus`, it creates a candidate property for the whole record value (`o._bounds`) instead of individual fields (`o.n._bounds`, `o.d._bounds`), which doesn't seem to make a lot of sense.

amarmaduke:
It was an intentional choice to have `o._bounds` for a record type because the abstracted variable has the correct condition: `((0 <= o.n) and (o.n <= 1)) and ((-1 <= o.d) and (o.d <= 0))` This matched how arrays were handled (with it just being `o._bounds` but the condition itself quantifying over all indices). If the record is mixed, the condition will only include the subrange fields. Doing it that way also made all of the structured expressions fit more nicely into one code flow.

src/lustre/lustreAbstractInterpretation.ml
amarmaduke: Oops

daniel-larraz:
>  If the record is mixed, the condition will only include the subrange fields.

When the fields of the record are subranges, I can see the user inferring what are the conditions checked by the property. I also see reasonable inferring the conditions of a candidate property like the one in the example, where a record with integer fields is made equal to a record with subrange fields. But consider a more complicated case where the fields of a record with integer fields are defined individually in terms of a mix of integer and subrange values. In that case, the conditions checked by the candidate property depends on the individual assignments done to each field, and therefore, inferring what are the conditions checked by the candidate property doesn't seem straightforward anymore. Moreover, if one of the conditions doesn't hold for one field, this may jeopardize discovering that the conditions for other fields in the candidate property are satisfied (since they are checked altogether). 

In any case, I'm going to merge this PR because I consider it makes good progress. But we should discuss the question above further in the next team meeting.

PR #825, opened 2021-12-14
Generate subrange constraint abstractions for each nested type independently

amarmaduke:


daniel-larraz:
Internal variables are leaked in generated properties. For instance, consider `tests/experimental/success/infer_tight_subrange.lus`. When analyzing `N4`, the new front-end generates `Candidate ((1 <= 1_contract_v1) and (1_contract_v1 <= 2))`. The property should be expressed in terms of the original names. If the simple name is ambiguous, we should add some prefix like we do for other contract elements.

PR #835, opened 2022-02-10
Remove automaton DSL

amarmaduke:
Note that there are a lot of warnings in the old frontend after this change. I tried to just disable them with `[@@@ocaml.warning "-.."]` but maybe did something wrong. Since all that code is getting deleted soon I didn't bother investigating further. The only code I did remove from the old frontend was just what was needed to make kind2 compile.


daniel-larraz:
The temporal operator `last` should be removed too. It was only used within the definition of an automaton.

I also found functions to print automata in `lustre/lustrePath.ml`, and a reference to automata in a comment in `lustre/lustreDeclarations.ml`. In addition, Ocaml 4.13 reports the following warning:
```
File "src/lustre/lustreAstInlineConstants.ml", line 358, characters 4-5:
358 |   | e -> e
          ^
Warning 11 [redundant-case]: this match case is unused.
```
@amarmaduke, could you please address these issues?
Thanks.

rxzephyr:
Hi, the recent release seems to have removed support for the automata. In my rough understanding, the automaton structure is more like a syntactic sugar but in some cases it may be possible to simplify the modeling. I am just curious about the decision and could you please tell me the reason?

daniel-larraz:
Thank you for your feedback.

PR #865, opened 2022-03-19
Fix current value of output streams in assumptions

amarmaduke:
- revert change that made imported node outputs not depend on any inputs (each output stream now depends on every input stream)
- conservatively check a contract import for not depending on current value of output stream (could relax this restriction by adding contract node summaries)
- add tests


daniel-larraz:
Shouldn't `ContractOutputContainsContractArguments` be removed since `ContractDependencyOnCurrentOutput` supersedes it and is more general?

doc/usr/source/2_input/1_lustre.rst
```diff
@@ -750,7 +751,7 @@ Benefits
 ^^^^^^^^
 
 Functions are interesting in the model-checking context of Kind 2 mainly as
-a mean to make an abstraction more precise. A realistic use-case is when one
+a mean to make an abstraction more precise. A realistic use-caseimported is when one
```
daniel-larraz, line 754: Typo: use-caseimported

PR #879, opened 2022-04-01
Fix consts in contract

### LINKED ISSUE #878: Referring to a constant in call inside contract gives Not_found
Hi Daniel, 
I just found this small issue with the current develop branch: if you have a contract like `guarantee f(CONST)` you get a Not_found error. It's easy to work around though.

```
node id_bool(x: bool) returns (out: bool)
let
  out = x;
tel

const ok: bool = true;
node x(in: bool) returns (out: bool)
(*@contract
  guarantee id_bool(ok);
*)
let
  out = in;
  --%PROPERTY id_bool(ok);
tel
```
checking this gives me
```
$ bin/kind2 test-mention-const.lus
 kind2 v1.4.0-440-g891f1908

<Error> Error opening input file 'test-mention-const.lus': Not_found
```

amarmaduke:
Fixes #878, also adds constant inlining for node/fn contracts, I assume this was accidentally missed?


daniel-larraz:
> also adds constant inlining for node/fn contracts, I assume this was accidentally missed?

Yes, constant inlining must be done for node/fn contracts too. Thanks.

PR #882, opened 2022-04-08
Add inductive array well-foundedness (dependency) check to new frontend

amarmaduke:


src/lustre/lustreArrayDependencies.ml
```diff
@@ -0,0 +1,252 @@
...
+module Ctx = TypeCheckerContext
+module Chk = LustreTypeChecker
+
+module StringMap = struct
+  include Map.Make(struct
+    type t = HString.t
+    let compare i1 i2 = HString.compare i1 i2
+  end)
+  let keys: 'a t -> key list = fun m -> List.map fst (bindings m)
+end
```
daniel-larraz, line 34: I would replace this declaration with just: `module StringMap = HString.HStringMap`

The `keys` function is not really necessary (see my previous comment).

src/lustre/lustreArrayDependencies.ml
```diff
@@ -0,0 +1,252 @@
+(* This file is part of the Kind 2 model checker.
+
+   Copyright (c) 2020 by the Board of Trustees of the University of Iowa
```
daniel-larraz, line 3: Last time I looked my calendar it was 2022 :-P

src/lustre/lustreArrayDependencies.mli
```diff
@@ -0,0 +1,36 @@
+(* This file is part of the Kind 2 model checker.
+
+   Copyright (c) 2020 by the Board of Trustees of the University of Iowa
```
daniel-larraz, line 3: Please update the year (2020 -> 2022).

daniel-larraz:
The following example seems to send Kind 2 to an infinite loop:
```
const N = 5;

node arrays3(y:int) returns (A,B,C: int^N);
let
  A[k] = B[k+1] + y;
  B[k] = C[k-1] - 2;
  C[k] = ( A[k-1] + B[k] ) * k ;
  --%PROPERTY C[0]>=A[0];
tel
```
@amarmaduke, could you please take a look at it?

Thanks.

daniel-larraz:
The check does not detect any cycle for this example:
```
node top() returns (A,C,B:int^3);
let
  C[i] = if i=0 then 0 else B[i];
  A=C;
  B[i] = if i=0 then 0 else A[i];
tel
```

src/lustre/lustreArrayDependencies.ml
daniel-larraz: My point was not only that the `keys` function could be removed but that doing so you could use directly the already defined types `HString.HStringMap` and `HString.HStringSet`. You only need to introduce an alias for convenience:

```
module StringMap = HString.HStringMap
module StringSet = HString.HStringSet
```

daniel-larraz:
There are still ways to elude the detection:
```
type my_record = struct {
  A: int^3;
};

node top() returns (r: my_record);
var C,B:int^3;
let
  C[i] = if i=0 then 0 else B[i];
  r = my_record { A=C; };
  B[i] = if i=0 then 0 else r.A[i];
tel
```

daniel-larraz:
The check shouldn't reject a model if different index names are used in different inductive definitions. However, Kind 2 rejects this model:
```
const N = 5;

node arrays3(y:int) returns (A,B,C: int^N);
let
  A[k] = B[k+1] + y;
  B[i] = C[i-1] - 2;
  C[k] = ( A[k-1] + B[k] ) * k ;
  --%PROPERTY C[0]>=A[0];
tel
```

amarmaduke:
At the moment inductive definitions with more than one index are not checked (only the first index is checked), but this is enough to make sure that the approach is good and I need to confirm exactly what we expect the checks to be for multiple indices.

daniel-larraz:
The check incorrectly detects a cycle for this example:
```
const N = 5;

node f(C: int^N) returns (B: int^N);
let
  B[k] = C[k-1] - 2;
tel

node arrays3(y:int) returns (A,B,C: int^N);
let
  A[k] = B[k+1] + y;
  B = f(C);
  --B[k] = C[k-1] - 2;
  C[k] = ( A[k-1] + B[k] ) * k ;
  --%PROPERTY C[0]>=A[0];
tel
```

PR #900, opened 2022-09-20
Ghost variable multiple assignment

lorchrob:
Add ability to declare and define multiple ghost variables in the same line.


src/lustre/lustreAstDependencies.ml
```diff
@@ -665,22 +670,44 @@ let rec vars_with_flattened_nodes: node_summary -> int -> LA.expr -> LA.SI.t
 (** get all the variables and flatten node calls using 
     the node summary for an expression *)
              
-let rec mk_contract_eqn_map: LA.contract_node_equation IMap.t -> LA.contract -> ((LA.contract_node_equation IMap.t), [> error]) result
-  = fun m ->
+(* The first map is a map from identifiers to contract equations, while the
+   second map only keeps track of the identifiers. We need both because
+   we may have multiple identifiers associated with the same equation, but
+   we still need to check for redeclaration. *)
```
daniel-larraz, line 676: Although this solution works, I would prefer a solution where only one map and one check are used. I suggest to use a map from identifiers to optional equations, i.e. `(LA.contract_node_equation option) IMap.t`, so you pass `Some gc` to  `check_and_add` for all cases except for the elements in the tail of the list of a ghost variable declaration. Then, in `extract_decls` you only add a declaration if `find_opt` returns `Some (Some i')`.

src/lustre/lustreDeclarations.ml
```diff
@@ -1544,10 +1544,14 @@ and eval_node_contract_item
...
 
-  (* Add ghost variables to context *)
-  | A.GhostVar v ->
+  (* Single ghost var assignment *)
+  | A.GhostVars (pos, (GhostVarDec (_, [_, i, ty])), expr) -> 
+    let v = A.TypedConst (pos, i, expr, ty) in 
     eval_ghost_var is_postponed inputs outputs locals ctx v, cpt_a, cpt_g
 
+  (* Multiple ghost var assignment *)
+  | A.GhostVars (pos, _, _) -> fail_at_position pos "Not supported in old frontend"
```
daniel-larraz, line 1553: You should say what is not supported. Something like "Multiple ghost variable declaration is not supported in old frontend".

PR #927, opened 2022-11-10
If statements and frame blocks

lorchrob:
Implement if statements and frame blocks


src/lustre/lustreParser.mly
```diff
@@ -593,12 +595,39 @@ check:
...
   | a = main_annot { a }
   | p = property { p }
   | p = check { p }
 
 
+
+node_if_block:
+  | IF; e = expr; 
+    THEN; 
+      l1 = list(node_item);
```
daniel-larraz, line 610: Should we accept an empty list here? I would reject it. Same applies to the case without `else`.

src/lustre/lustreInput.mli
```diff
@@ -116,6 +116,10 @@ type error = [
   | `LustreTypeCheckerError of Lib.position * LustreTypeChecker.error_kind
   | `LustreUnguardedPreError of Lib.position * LustreAst.expr
   | `LustreParserError of Lib.position * string
+  | `LustreDesugarIfBlocksError of Lib.position * LustreDesugarIfBlocks.error_kind
+  | `LustreDesugarFrameBlocksError of Lib.position * LustreDesugarFrameBlocks.error_kind
+  | `LustreRemoveMultAssignError of Lib.position * LustreRemoveMultAssign.error_kind
+  | `LustreRemoveMultAssignError of Lib.position * LustreRemoveMultAssign.error_kind
```
daniel-larraz, line 122: `LustreRemoveMultAssignError` is duplicate

src/lustre/lustreDesugarFrameBlocks.ml
```diff
@@ -0,0 +1,289 @@
...
+type error_kind = Unknown of string
+  | MisplacedNodeItemError of A.node_item
+  | InitializationNotFoundError of A.node_item
+  | MisplacedFrameBlockError of A.node_item
+
+let error_message error = match error with
+  | Unknown s -> s
+  | MisplacedNodeItemError ni -> "Node item " ^ Lib.string_of_t A.pp_print_node_item ni ^ " is not allowed in frame block."
+  | MisplacedFrameBlockError ni -> "FrameBlock " ^ Lib.string_of_t A.pp_print_node_item ni ^ " is not allowed in function."
+  | InitializationNotFoundError ni -> "Node item " ^ Lib.string_of_t A.pp_print_node_item ni ^ " does not have a corresponding initialization in the frame block."
```
daniel-larraz, line 45: This error is not used anymore.

src/lustre/lustreSyntaxChecks.ml
```diff
@@ -91,6 +98,7 @@ let error_message kind = match kind with
   | UnsupportedAssignment -> "Assignment not supported"
   | AssumptionVariablesInContractNode -> "Assumption variables not supported in contract nodes"
   | ClockMismatchInMerge -> "Clock mismatch for argument of merge"
+  | MisplacedVarInFrameBlock id -> "Misplaced var '" ^ HString.string_of_hstring id ^ "' is not declared in the frame block header"
```
daniel-larraz, line 101: I would replace this message with the following message:
`Variable 'x' is defined in the frame block but not declared in the frame block header`

src/lustre/lustreDesugarIfBlocks.ml
```diff
@@ -0,0 +1,337 @@
...
+module Chk = LustreTypeChecker
+module Ctx = TypeCheckerContext
+
+
+type error_kind = Unknown of string
+  | MisplacedNodeItemError of A.node_item
+
+let error_message error = match error with
+  | Unknown s -> s
+  | MisplacedNodeItemError ni -> "Node item " ^ Lib.string_of_t A.pp_print_node_item ni ^ " is not allowed in if block"
```
daniel-larraz, line 44: I would say what kind of node item we are referring to but I would not print the whole item. It doesn't provide additional information and it might be quite big. For instance, if an assert is found, I would say: `Asserts are not allowed in if block`

src/lustre/lustreDesugarFrameBlocks.ml
```diff
@@ -0,0 +1,289 @@
...
+
+type error_kind = Unknown of string
+  | MisplacedNodeItemError of A.node_item
+  | InitializationNotFoundError of A.node_item
+  | MisplacedFrameBlockError of A.node_item
+
+let error_message error = match error with
+  | Unknown s -> s
+  | MisplacedNodeItemError ni -> "Node item " ^ Lib.string_of_t A.pp_print_node_item ni ^ " is not allowed in frame block."
+  | MisplacedFrameBlockError ni -> "FrameBlock " ^ Lib.string_of_t A.pp_print_node_item ni ^ " is not allowed in function."
```
daniel-larraz, line 44: See comment for `MisplacedNodeItemError` in `lustreDesugarIfBlocks.ml`

src/lustre/lustreAstHelpers.ml
```diff
@@ -1086,15 +1116,18 @@ let rec abstract_pre_subexpressions: expr -> expr = function
...
+    | Body eqn -> [eqn]
+    | IfBlock (_, _, l1, l2) -> 
+      List.flatten (List.map extract_equation_helper l1) @ 
+      List.flatten (List.map extract_equation_helper l2)
+    | FrameBlock (_, _, nes, nis) -> 
+      nes @ List.flatten (List.map extract_equation_helper nis)
+    | AnnotMain _ -> []
+    | AnnotProperty _ -> [])
+  in 
+  fun items -> List.flatten (List.map extract_equation_helper items)
```
daniel-larraz, line 1130: This function is never called. I have my doubts about the usefulness of this function in future code since we are extracting equations without taking into consideration `IfBlocks` and `FrameBlock`. I would remove it.

src/lustre/lustreSyntaxChecks.ml
```diff
@@ -612,6 +658,17 @@ and check_struct_items ctx items =
...
+(* Within a frame block, make sure vars in the LHS of 'ni' appear somehwere in 'vars'  *)
+and check_frame_vars pos vars ni = 
+  let vars_of_ni = List.map snd (LAH.vars_lhs_of_eqn_with_pos ni) in
+  (* 'find_vars' stores the original variables from 'vars_of_ni' and their corresponding
+     variables found in 'vars', if they exist *)
+  let find_vars = List.map (fun var1 -> (var1, (List.find_opt (fun var2 -> var2 = var1) vars))) vars_of_ni in
+  let no_match = List.filter (fun x -> match x with | (_, None) -> true | (_, Some _) -> false) find_vars in
+  match no_match with
+    | [] -> Res.ok ()
+    | (var, _) :: _ -> syntax_error pos (MisplacedVarInFrameBlock var)
```
daniel-larraz, line 670: This could be more concisely expressed with set operations:
```
let open HString in
let unlisted =
  HStringSet.diff (HStringSet.of_list vars_of_ni) (HStringSet.of_list vars)
in
match HStringSet.choose_opt unlisted with
| None -> Res.ok ()
| Some var -> syntax_error pos (MisplacedVarInFrameBlock var)
```

doc/usr/source/2_input/1_lustre.rst
```diff
@@ -793,3 +793,155 @@ a linear one.
...
+present in the list of variables ``frame ( y1, y2 )``.
+
+.. code-block:: none
+
+   frame ( y1, y2 )
+   y1 = init1;
+   let
+      y1 = expr1;
+      y2 = expr2;
+   tel
```
daniel-larraz, line 912: I would clarify what is the initial value of a variable when it does not have an initialization (perhaps with a concrete example). Namely, it is the initial value of the expression assigned in the frame block body or undefined otherwise.

doc/usr/source/2_input/1_lustre.rst
```diff
@@ -793,3 +793,155 @@ a linear one.
...
+the ``fi`` token. As with other mainstream programming languages, Kind 2 allows for arbitrary nesting of ``if`` statements,
+as well as writing ``if`` statements that do not have an ``else`` block. 
+
+Frame conditions
+^^^^^^^^^^^^^^^^
+Kind 2 also has support for code blocks with frame conditions. At the beginning of the block
+(denoted by the ``frame`` keyword), the user specifies a list of variables that they wish to 
+define within the frame block. All variables defined within the frame block must be present in
+this list. Then, initial values are optionally specified for these variables. 
+Variables are defined within the frame block body (denoted by the ``let`` and ``tel`` kewords).
```
daniel-larraz, line 832: Typo: kewords -> keywords

tests/ounit/lustre/lustreSyntaxChecks/misplaced_frame_block.lus
```diff
@@ -0,0 +1,24 @@
...
+   y1 = 2;
+else
+   y1 = 3;
+fi
+tel;
+
+node oscillate() returns (Z: bool)
+let
+   Z = true -> not pre Z;
+tel;
```
daniel-larraz, line 24: Kind 2 returns the following error for this input: `Frame blocks are not allowed inside other frame blocks`

The message should say something like:
`Frame blocks are not allowed inside other frame blocks or if statements`

tests/ounit/lustre/lustreSyntaxChecks/misplaced_node_item_frame.lus
```diff
@@ -0,0 +1,18 @@
...
+   let
+      assert x > 0;
+      y2 = pre x;
+   tel
+tel;
+
+node oscillate() returns (Z: bool)
+let
+   Z = true -> not pre Z;
+tel;
```
daniel-larraz, line 18: Kind 2 returns the following error for this input: `Node item assert (x > 0); is not allowed in if block`

According to a previous discussion, Kind 2 should report:
`Asserts are not allowed in if block`

src/lustre/lustreDesugarIfBlocks.ml
```diff
@@ -0,0 +1,344 @@
...
+type error_kind = Unknown of string
+  | MisplacedNodeItemError of A.node_item
+
+let error_message error = match error with
+  | Unknown s -> s
+  | MisplacedNodeItemError ni -> (match ni with
+    | Body (Assert _) -> "Asserts are not allowed inside frame blocks."
+    | FrameBlock _ -> "Frame blocks are not allowed inside other frame blocks."
+    | AnnotMain _ -> "Main annotations are not allowed inside frame blocks."
+    | AnnotProperty _ -> "Property annotations are not allowed inside frame blocks."
```
daniel-larraz, line 48: All these error messages should say: `[...] inside if blocks` instead of `[...] inside frame blocks`

tests/regression/success/frame_block_no_init.lus
```diff
@@ -0,0 +1,21 @@
...
+      if c2
+      then
+        x1 = 37;
+      else
+        x1 = 34;
+      fi
+    fi
+  tel
+
+  check (true -> (x1 = 24 or x1 = 37 or x1 = 34));
```
daniel-larraz, line 19: I would replace this check with this stronger version:
`check (x1 = 24 or x1 = 37 or x1 = 34);`

tests/regression/success/frame_block_unguarded_pres.lus
```diff
@@ -0,0 +1,37 @@
+node test(x: int) returns (y1: int; y2: bool; p1, p2, p3: bool);
+let 
+  p1 = (y1 >= 0);
+  p2 = (true);
```
daniel-larraz, line 4: @lorchrob I think you forgot to define property `p2` :)

tests/regression/success/imperative_if_oracle3.lus
```diff
@@ -0,0 +1,36 @@
...
+  --%PROPERTY p1;
+  --%PROPERTY p2;
+  --%PROPERTY p3;
+  --%PROPERTY p4;
+  
+
+
+  if (x > 0)
+  then 
+    -- Can we split these definitions? Like y1[0] = 23? Don't think so.
```
daniel-larraz, line 30: The answer is no. Deprecated comment.

src/lustre/lustreDesugarFrameBlocks.ml
```diff
@@ -0,0 +1,276 @@
...
+ *)
+
+module A = LustreAst
+module R = Res
+module LDI = LustreDesugarIfBlocks
+module Chk = LustreTypeChecker
+
+let (let*) = R.(>>=)
+
+type error_kind = Unknown of string
```
daniel-larraz, line 36: The `Unknown` error is not used. Please remove it.

src/lustre/lustreDesugarIfBlocks.ml
```diff
@@ -0,0 +1,344 @@
...
+
+module R = Res
+module A = LustreAst
+module AH = LustreAstHelpers
+module GI = GeneratedIdentifiers
+module Chk = LustreTypeChecker
+module Ctx = TypeCheckerContext
+
+
+type error_kind = Unknown of string
```
daniel-larraz, line 39: The `Unknown` error is not used. Please remove it.

src/lustre/lustreRemoveMultAssign.mli
```diff
@@ -0,0 +1,34 @@
...
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+   implied. See the License for the specific language governing
+   permissions and limitations under the License. 
+*)
+
+(** @author Rob Lorch *)
+
+type error_kind = Unknown of string
```
daniel-larraz, line 20: The `Unknown` error is not used. Please remove it.

src/lustre/lustreRemoveMultAssign.ml
```diff
@@ -0,0 +1,299 @@
...
+  | A.NodeDecl (s, ((node_id, b, nps, cctds, ctds, nlds, nis, co) as d)) -> 
+    let* ctx = Chk.get_node_ctx ctx d in
+    let* res = R.seq (List.map (desugar_node_item ctx) nis) in
+    let (new_decls, nis) = List.split res in
+    R.ok (A.NodeDecl (s, (node_id, b, nps, cctds, ctds, (List.flatten new_decls) @ nlds, (List.flatten nis), co))) 
+  | _ -> R.ok (decl)
+
+
+(** Desugars a declaration list to remove IfBlocks. Converts IfBlocks to
+    declarative ITEs, filling in oracles if branches are undefined. *)
```
daniel-larraz, line 296: This comment does not belong here. Update comment to reflect the behavior of this function.

src/lustre/lustreRemoveMultAssign.ml
```diff
@@ -0,0 +1,299 @@
...
+let desugar_node_item ctx ni = match ni with
+  | A.IfBlock _ -> 
+    let* (nis, ib, decls1, _) = remove_mult_assign_from_ni ctx ni in 
+    R.ok (decls1, nis @ [ib])
+  | A.FrameBlock _ -> 
+    let* (nis2, fb, decls1, _) = remove_mult_assign_from_ni ctx ni in 
+    R.ok (decls1, nis2 @ [fb])
+  | _ -> R.ok ([], [ni])
+
+(** Desugars an individual node declaration (removing IfBlocks). *)
```
daniel-larraz, line 280: Update this comment.

src/lustre/lustreRemoveMultAssign.ml
```diff
@@ -0,0 +1,299 @@
...
+    | A.Ident (pos, i1) -> 
+      let comb = List.combine arraydefs_original arraydefs_new in
+      let update = Lib.find_map (fun pair -> match pair with 
+        | (A.ArrayDef(_, i2, _), A.ArrayDef (_, i3, _)) when i1 = i2 -> 
+          Some (A.Ident(pos, i3))
+        | _ -> None
+      ) comb in (match update with 
+        | Some value -> value 
+        | None -> A.Ident(pos, i1)
+      )
```
daniel-larraz, line 108: Notice that you are combining `arraydefs_original` and `arraydefs_new` for each identifier despite their combination does not depend on the identifier. This suggests you should compute their combination once before calling `modify_arraydefs_in_expr` and pass it to the function. Moreover, the call to `Lib.find_map` is basically using `comb` as an association list. So I suggest to replace this code with the following code:
```
| A.Ident (pos, i1) -> (
  let update = List.assoc_opt i1 array_assoc_list in
  match update with 
  | Some id -> A.Ident(pos, id)
  | None -> A.Ident(pos, i1)
)
```
where `array_assoc_list` is a new argument that replaces arguments `arraydefs_original` and `arraydefs_new`, and can be computed in `create_new_eqs` before the call to `modify_arraydefs_in_expr`  as follows:
```
let get_array_ids =
   List.filter_map (function
     | A.ArrayDef (_, id, _) -> Some id
     | _ -> None)
in
let array_assoc_list =
  let arrayids_original = get_array_ids ss in
  let arrayids_new = get_array_ids sis in
  List.combine arrayids_original arrayids_new
in
let expr = modify_arraydefs_in_expr array_assoc_list expr in
```

src/lustre/lustreDesugarIfBlocks.mli
```diff
@@ -0,0 +1,50 @@
...
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+   implied. See the License for the specific language governing
+   permissions and limitations under the License. 
+ *)
+
+ (** 
+   Code for desugaring imperative-style if blocks to functional ITEs.
+
+   The code has a few steps.
+    1. Remove multiple assignment from if blocks using temp variables.
```
daniel-larraz, line 22: Step 1 is performed outside this module. Update comment.

src/lustre/lustreDesugarIfBlocks.ml
```diff
@@ -0,0 +1,343 @@
...
+let split_and_flatten3 ls =
+  let xs = List.map (fun (x, _, _) -> x) ls |> List.flatten in
+  let ys = List.map (fun (_, y, _) -> y) ls |> List.flatten in
+  let zs = List.map (fun (_, _, z) -> z) ls |> List.flatten in
+  xs, ys, zs
+
+
+(** Helper function for 'desugar_node_item' that converts IfBlocks to a list
+    of ITEs. There are a number of steps in this process.
+    1. Removing multiple assignment in if blocks
```
daniel-larraz, line 279: Replace Step 1 with Pre-condition.

src/lustre/lustreDesugarIfBlocks.ml
```diff
@@ -0,0 +1,330 @@
...
+      Node (i, str, Leaf j)
+    |	Node (Leaf i, str, ((Node _) as j)) ->  
+      let j = simplify_tree j in
+      if trees_eq (Leaf i) j then j
+    else Node (Leaf i, str, j)
+    | Node (((Node _) as i), str, ((Node _) as j)) -> 
+      let i = simplify_tree i in
+      let j = simplify_tree j in
+      if trees_eq i j then i else
+      Node (i, str, j)
```
daniel-larraz, line 254: Can't you simplify all these cases to just two?
```
let rec simplify_tree node = 
  match node with
    | Leaf _ -> node
    | Node (i, str, j) -> 
      let i = simplify_tree i in
      let j = simplify_tree j in
      if trees_eq i j then i else
      Node (i, str, j)
```

src/lustre/lustreDesugarIfBlocks.ml
```diff
@@ -0,0 +1,330 @@
...
+    | Error _ -> false
+  )
+  | Node (l1, _, r1), Node (l2, _, r2) -> 
+    trees_eq l1 l2 && trees_eq r1 r2
+  | _ -> false
+
+  
+(** Removes redundancy from a binary tree. 
+   Currently, redundancy check doesn't quite work because
+   different positions mess with equality of leaves. *)
```
daniel-larraz, line 236: Is that right? The description of `AH.syn_expr_equal` (used in `trees_eq`) claims positions are ignored.

src/lustre/lustreDesugarIfBlocks.ml
```diff
@@ -0,0 +1,330 @@
...
+  let* tree_map = if_block_to_trees ib in
+  let (lhss, trees) = LhsMap.bindings (tree_map) |> List.split in
+  let trees = List.map simplify_tree trees in  
+  let poss = List.map (fun (A.StructDef (a, _)) -> a) lhss in
+  let ites = List.map2 tree_to_ite poss trees in
+  let tys = (List.map (get_tree_type ctx) lhss) in 
+  let tys = (List.map (fun x -> match x with | Some y -> y | None -> assert false (* not possible *)) 
+                       tys) in
+  let res = List.map2 fill_ite_with_oracles ites tys |>
+                               List.map (fun (a, b, c) -> a, b, c) in
```
daniel-larraz, line 283: `|> List.map (fun (a, b, c) -> a, b, c)` is the identity function. Redundant.

src/lustre/lustreDesugarIfBlocks.ml
```diff
@@ -0,0 +1,330 @@
...
+      (* shouldn't be possible *)
+      | _ -> assert false
+  ) in
+  (helper ib LhsMap.empty [])
+  
+(** Converts a tree of conditions/expressions to an ITE expression. *)
+let rec tree_to_ite pos node =
+  match node with
+    | Leaf Some expr -> expr
+    | Leaf None -> A.Ident(pos, HString.mk_hstring "ib_oracle")
```
daniel-larraz, line 176: Define `HString.mk_hstring "ib_oracle"` as global constant in the module and use the constant everywhere. `fill_ite_with_oracles` depends on this constant.

src/lustre/lustreDesugarIfBlocks.ml
lorchrob: You're right. This is a deprecated comment.

src/lustre/lustreDesugarIfBlocks.ml
```diff
@@ -0,0 +1,330 @@
...
+      let e1, gids1, decls1 = (
+      match e1 with
+        | TernaryOp _ as top -> fill_ite_with_oracles top ty
+        | Ident(p, s) when s = HString.mk_hstring "ib_oracle" -> 
+          let (expr, gids) = (mk_fresh_ib_oracle ty) in
+          (match expr with
+            | A.Ident (_, name) ->  expr, gids, [A.NodeVarDecl(p, (p, name, ty, ClockTrue))]
+            | _ -> assert false (* not possible*))
+        | _ -> e1, GI.empty (), []
+      ) in
```
daniel-larraz, line 207: Since the same code is applied to both branches, factor out this code into a local function, and call the function on each branch. It makes the symmetry clear, keeping the code synchronized is easier, etc.

lorchrob: I chose a different refactoring that I think makes a bit more sense. Let me know if you want me to switch back to your suggestion.

daniel-larraz: LGTM. Thanks.

src/lustre/lustreDesugarFrameBlocks.mli
```diff
@@ -0,0 +1,51 @@
...
+
+type error_kind = 
+  | MisplacedNodeItemError of A.node_item
+  | MisplacedFrameBlockError of A.node_item
+
+val error_message : error_kind -> string
+
+type error = [
+  | `LustreDesugarFrameBlocksError of Lib.position * error_kind
+  | `LustreDesugarIfBlocksError of Lib.position * LustreDesugarIfBlocks.error_kind
```
daniel-larraz, line 39: This error is not used in this module. Remove it.

src/lustre/lustreDesugarFrameBlocks.ml
```diff
@@ -0,0 +1,264 @@
...
+  For each node item in frame block body:
+    Fill in ITE oracles and initialize equations (RHS) when an initialization
+    value is specified.
+  For each initialization:
+    Fill in an equation if one doesn't exist.
+  For each variable that is neither initialized nor defined:
+    Fill in an equation of the form 'y = pre y' (initially undefined)
+*)
+let desugar_node_item _ ni = match ni with
+    (* All multiple assignment is removed in lustreDesugarIfBlocks.ml *)
```
daniel-larraz, line 230: Update comment.

src/lustre/lustreDesugarFrameBlocks.ml
```diff
@@ -0,0 +1,264 @@
...
+    ) in
+
+    (match res with
+      (* Already defined in frame block *)
+      | Some _ -> R.ok []
+      (* Fill in equation in frame block body *)
+      | None -> R.ok [A.Body(A.Equation(pos, lhs, Arrow(pos, init, Pre(pos, build_array_index (List.rev id2)))))]
+    )
+  (* Assert in frame block guard *)
+  | A.Equation(pos, _, _) -> mk_error pos (MisplacedNodeItemError (A.Body ne))
```
daniel-larraz, line 149: My understanding is that this case is impossible because multiple assignments in frame blocks have been desugared. If that is the case, I would replace the call to `mk_error` with `assert false` and I would explain why the case is impossible with a comment.

src/lustre/lustreDesugarFrameBlocks.ml
```diff
@@ -0,0 +1,264 @@
...
+    expression (which is stuttering, ie, 'pre variable').
+*)
+let rec fill_ite_helper fill = function
+  (* Replace all oracles with 'fill' *)
+  | A.Ident (pos, i) -> 
+    (* See if 'i' is of the form "n_iboracle" *)
+    let str = String.split_on_char '_' (HString.string_of_hstring i) |>
+      List.rev |> List.hd
+    in
+    if (str = "iboracle") then fill else A.Ident(pos, i)
```
daniel-larraz, line 60: The check `str = "iboracle"` creates a logic dependency between modules (`LustreDesugarFrameBlocks` and `LustreDesugarIfBlocks`). You should put the definition of the constant `"iboracle"` and this check in a common place, outside both modules, so that the dependency is clear. Perhaps `GeneratedIdentifiers` is a good place. If not, choose a different module which makes sense or create a new module.

src/lustre/lustreParser.messages
```diff
@@ -2648,3 +2648,51 @@ main: FUNCTION ASSUME LPAREN RPAREN RETURNS LPAREN RPAREN CONTRACT_PSATBLOCK VAR
...
+
+<YOUR SYNTAX ERROR MESSAGE HERE>
+
+main: FUNCTION ASSUME LPAREN RPAREN RETURNS LPAREN RPAREN LET IF ASSUME THEN MAIN_PSBLOCKSTART PSBLOCKEND ELSE MAIN_PSBLOCKSTART PSBLOCKEND ELSE
+
+<YOUR SYNTAX ERROR MESSAGE HERE>
+
+main: FUNCTION ASSUME LPAREN RPAREN RETURNS LPAREN RPAREN LET IF ASSUME THEN MAIN_PSBLOCKSTART PSBLOCKEND XOR
+
+<YOUR SYNTAX ERROR MESSAGE HERE>
```
daniel-larraz, line 2698: Replace `<YOUR SYNTAX ERROR MESSAGE HERE>` with generic error message `Syntax Error!`

src/lustre/lustreAstNormalizer.mli
```diff
@@ -72,9 +72,24 @@ type error = [
...
+  inductive_variables : LustreAst.lustre_type GeneratedIdentifiers.StringMap.t;
+  quantified_variables : LustreAst.typed_ident list;
+  node_is_input_const : (bool list) GeneratedIdentifiers.StringMap.t;
+  contract_calls_info : LustreAst.contract_node_decl GeneratedIdentifiers.StringMap.t;
+  contract_scope : (Lib.position * HString.t) list;
+  contract_ref : HString.t;
+  interpretation : HString.t GeneratedIdentifiers.StringMap.t;
+  local_group_projection : int
+}
+
```
daniel-larraz, line 88: Type `info` is not used outside this module. The declaration in the interface file is not required.

PR #934, opened 2022-12-14
Fix version number

kit-ty-kate:


daniel-larraz:
Hello @kit-ty-kate. Thank you for your patch, but my next planned commit is going to override your change so I am not going to merge your PR. Specifically, every time after I publish a new release of Kind 2 in the opam repositories, I remove the version field from the `dune-project` file. This also has the effect of removing the version field from the opam file (after running dune with the `generate_opam_files` option set to true).

For the next release, I will make sure the version field in the opam file does not contain the v prefix. Thank you.

PR #943, opened 2023-01-29
Reachability queries

lorchrob:
Add support for reachability queries


src/kEvent.mli
```diff
@@ -163,21 +163,21 @@ val invariant : string list -> Term.t -> Certificate.t -> bool -> unit
 (** Broadcast a step cex *)
 val step_cex :
   'a InputSystem.t -> Analysis.param -> TransSys.t -> string ->
-  (StateVar.t * Model.value list) list -> unit
+  (StateVar.t * Model.value list) list -> Property.prop_kind -> unit
 
 (** Broadcast a property status *)
-val prop_status : Property.prop_status -> 'a InputSystem.t -> Analysis.param -> TransSys.t -> string -> unit
+val prop_status : Property.prop_status -> 'a InputSystem.t -> Analysis.param -> TransSys.t -> string -> Property.prop_kind -> unit
```
daniel-larraz, line 169: I noticed you have changed a lot of the interfaces of this module to pass the kind of the property. However, in most (if not all) the cases, you are also passing the transition system and the name of the property, which is enough to retrieve the kind of the property within the function. Could you please review whether you really need to pass the new argument or you can reuse the previous interface?

src/kEvent.mli
```diff
@@ -75,7 +75,7 @@ val log_counterexample : Lib.kind_module -> Lib.log_level -> string list -> Tran
 
     Should only be used by the invariant manager, other modules must use
     {!prop_status} to send it as a message. *)
-val log_prop_status : Lib.log_level -> TransSys.t -> (string * Property.prop_status) list -> unit
+val log_prop_status : Lib.log_level -> TransSys.t -> (string * Property.prop_status) list -> (string * Property.prop_kind) list -> unit
```
daniel-larraz, line 78: The new argument is related to the third argument. Instead of adding a new argument to pass the list of property kinds separately from the list of status, I would modified the current argument to pass all info together: ` (string * Property.prop_status * Property.prop_kind) list`. In other words, I would combine the two lists outside of this function instead of within the function. This prevents calls to this function where the lengths of the two lists do not match.

src/kind2Flow.ml
```diff
@@ -151,9 +153,13 @@ let status_of_trans_sys sys =
...
         Format.eprintf "%s KU@." n;
         u+1,f
-      | (n, Property.PropFalse _) ->
+      | (n, Property.PropFalse _) when TSys.get_prop_kind sys n = Invariant ->
         Format.eprintf "%s FALSE@." n;
         u,f+1
+      | (n, Property.PropInvariant _) -> 
+        if TSys.get_prop_kind sys n = Invariant
+        then u,f
+        else (Format.eprintf "%s UNREACHABLE@." n; u,f+1)
```
daniel-larraz, line 162: You should update the description of the `status_of_trans_sys` function. Something like:
```
The exit status is
  * 0 if some properties are unknown or k-true (timeout),
  * 10 if some invariant properties are falsifiable (unsafe) or some reachability properties are unreachable,
  * 20 if all invariant properties are invariants (safe) and all reachability properties are reachable. *)
```

src/transSys.mli
```diff
@@ -506,6 +506,12 @@ val get_prop_term : t -> string -> Term.t
     system of the first property of name [n]. *)
 val get_prop_status : t -> string -> Property.prop_status 
 
+(** Return current status of the property
```
daniel-larraz, line 509: status -> kind

src/transSys.mli
```diff
@@ -523,6 +529,12 @@ val is_disproved : t -> string -> bool
     system of each property along with the name of the property. *)
 val get_prop_status_all_nocands : t -> (string * Property.prop_status) list
 
+(** Return the kind of all properties excepted candidates
+
+    [get_prop_status t] returns the status saved in the transition
```
daniel-larraz, line 534: status -> kind

src/transSys.mli
```diff
@@ -537,6 +549,12 @@ val get_prop_status_all_unknown : t -> (string * Property.prop_status) list
 (** Instantiate all properties to the bound *)
 val props_list_of_bound : t -> Numeral.t -> (string * Term.t) list 
 
+(** Instantiate all invariant properties to the bound *)
+val props_list_of_bound_no_skip : t -> Numeral.t -> (string * Term.t) list 
+
+(** Instantiate all reachable properties to the bound *)
+val props_list_of_bound_skip : t -> Numeral.t -> (string * Term.t) list
```
daniel-larraz, line 556: You should improve the description of these functions.

src/lustre/lustreAst.ml
```diff
@@ -227,13 +227,24 @@ type node_equation =
...
 
+type prop_bound =
+  | From of int
+  | Within of int
+  | At of int
+  | FromWithin of int * int
+
+type prop_kind =
+  | Invariant
+  (* The 'HString.t' refers to the timestep of the bound, eg "within 10 timesteps" *)
```
daniel-larraz, line 238: Old comment?

src/lustre/lustreAst.mli
```diff
@@ -229,13 +229,27 @@ type node_equation =
...
+type prop_bound =
+  | From of int
+  | Within of int
+  | At of int
+  | FromWithin of int * int
+
+(** Whether a property is reachable or invariant *)
+type prop_kind =
+  | Invariant
+  (* The 'HString.t' refers to the timestep of the bound, eg "within 10 timesteps" *)
```
daniel-larraz, line 243: Old comment?

src/lustre/lustreAstNormalizer.mli
```diff
@@ -65,7 +65,9 @@
   7. Restarts of node calls (if it is not a constant)
 
      @author Andrew Marmaduke *)
-     
+
+val ctr_id : HString.t
+
```
daniel-larraz, line 70: I would move this to `generatedIdentifiers`.

src/lustre/lustreParser.mly
```diff
@@ -574,25 +578,59 @@ main_annot:
...
+  | PROPERTY_SSBLOCKSTART ; REACHABLE ; name = option(STRING);
+    COLON; e = qexpr ;  bound = property_timestep; SEMICOLON; SSBLOCKEND
+    { A.AnnotProperty (mk_pos $startpos, name, e, bound) }
 
 check:
   | CHECK ; name = option(STRING) ; e = qexpr ; SEMICOLON
-    { A.AnnotProperty (mk_pos $startpos, name, e) }
+    { A.AnnotProperty (mk_pos $startpos, name, e, Invariant) }
+  | CHECK ; REACHABLE ; name = option(STRING) ; e = qexpr ; bound = property_timestep; SEMICOLON
+    { A.AnnotProperty (mk_pos $startpos, name, A.UnaryOp (mk_pos $startpos, A.Not, e), bound) }
```
daniel-larraz, line 633: For all these reachability checks, I would store the original expression in the Ast, and convert it to the negation later on.

src/lustre/lustreAstNormalizer.ml
```diff
@@ -711,7 +722,40 @@ and normalize_item info map = function
...
+        A.BinaryOp (pos, Or, 
+          A.CompOp(pos, A.Lt, Ident(dpos, ctr_id), 
+          Const (dpos, Num (HString.mk_hstring (string_of_int b1)))),
+          A.CompOp(pos, A.Gt, Ident(dpos, ctr_id), 
+                              Const (dpos, Num (HString.mk_hstring (string_of_int b2)))))
+        )
+
+      | Reachable _ -> A.UnaryOp (pos, A.Not, expr)
+      | _ -> expr
+    ) in
```
daniel-larraz, line 758: If you swap the order of this `let expr` binding and the `let name'` binding below, then reachability properties without a name will use the original expression instead of the negated expression and calls to the `rm_neg_prop_str` function in `kEvent` will not be necessary anymore.

src/induction/base.ml
```diff
@@ -193,18 +221,40 @@ let rec next (input_sys, aparam, trans, solver, k, unknowns) =
...
       nu_unknowns |> List.partition (
-        fun (name,_) -> match TransSys.get_prop_status trans name with
-        | Property.PropKTrue k -> k_int > k
+        fun (name,_) -> match (TransSys.get_prop_status trans name, TransSys.get_prop_kind trans name) with
+        | (Property.PropKTrue k, Invariant) 
+        | (Property.PropKTrue k, Reachable None) -> k_int > k
+        | (Property.PropKTrue k, Reachable Some (From b)) -> k_int > k && b <= k_int
+        | (Property.PropKTrue k, Reachable Some (Within b)) -> k_int > k && b >= k_int
+        | (Property.PropKTrue k, Reachable Some (At b)) -> k_int > k && b = k_int
+        | (Property.PropKTrue k, Reachable Some (FromWithin (b1, b2))) -> k_int > k && b1 <= k_int && b2 >= k_int
```
daniel-larraz, line 253: Kind 2 should do three things. (1) Classifying all the (negated) reachability properties with a lower bound `L` as K-true for `L-1` from the beginning, (2) classifying all the reachability properties with an upper bound `U` as unreachable as soon as they are K-true for `U`, and  (3) filtering out all the reachability properties that are out of bounds. This code only addresses the latter (and with some issues). 

If you address the first point, then you don't need to perform the first partition between `out_of_bounds` and `nu_unknowns` because the partition between `unknowns_at_k` and `k_true` would take care of reachability properties with a lower bound. Moreover, the second partition would be as simple as before (`| Property.PropKTrue k -> k_int > k`) because it would depend only on the property being K-true.

If you address the second point too, then Kind 2 will not include rechability properties out of the upper bound either (because they are already proven).

src/induction/base.ml
```diff
@@ -193,18 +221,40 @@ let rec next (input_sys, aparam, trans, solver, k, unknowns) =
...
         | _ -> true
       )
     in
-
+    let step = ref k in
     let unfalsifiable =
```
daniel-larraz, line 258: Please avoid code with side effects if possible. In this case, compute two values `let unfalsifiable, k =`. Then replace `if not invs then skip_steps_next trans solver step k;` with `k = if not invs then skip_steps_next trans solver k k else k` and `k_true @ unfalsifiable` with `k_true @ unfalsifiable, k`. It is fine to use a local reference within ` skip_steps_next` to implement the while loop. The important point is that any use of a (stateful) variable is local to the function that use it and is not shared by the caller.

src/induction/base.ml
```diff
@@ -295,16 +356,36 @@ let rec next (input_sys, aparam, trans, solver, k, unknowns) =
...
+    TransSys.trans_of_bound (Some (SMTSolver.declare_fun solver)) trans !step
+    |> SMTSolver.assert_term solver ;
+
+    step := Numeral.succ !step;
+  done;
+  num_skip
 
 (* Initializes the solver for the first check. *)
-let init input_sys aparam trans =
+let init input_sys aparam trans invs =
```
daniel-larraz, line 378: Rename the new bool argument `invs` with `no_skip` (in all functions). The current name is misleading.

src/induction/base.ml
```diff
@@ -295,16 +356,36 @@ let rec next (input_sys, aparam, trans, solver, k, unknowns) =
     else
      (* Looping. *)
      next
-      (input_sys, aparam, trans, solver, k_p_1, unfalsifiable)
+      (input_sys, aparam, trans, solver, k_p_1, out_of_bounds @ unfalsifiable, invs)
+
+and skip_steps_init trans solver =
```
daniel-larraz, line 361: For non-mutually recursive functions use `let` instead of `and`.

src/lustre/lustreParser.mly
```diff
@@ -127,8 +127,12 @@ let mk_span start_pos end_pos =
 (* Token for assertions *)
 %token ASSERT
     
-(* Token for check *)
+(* Tokens for check *)
 %token CHECK
+%token REACHABLE
```
daniel-larraz, line 132: Add `INVARIANT` token too.

src/kEvent.ml
```diff
@@ -386,13 +470,29 @@ let pp_print_counterexample_pt
...
+            "@{<red>%s@}:@,  @[<v>%a@]"
+            title
+            (InputSystem.pp_print_path_pt input_sys trans_sys disproved)
+            (Model.path_of_list cex))
+      | _ -> 
+        Format.fprintf ppf
+        "@{<red>%s@}:@,  @[<v>%a@]"
+        title
+        (InputSystem.pp_print_path_pt input_sys trans_sys disproved)
+        (Model.path_of_list cex)
```
daniel-larraz, line 493: Factor out this duplicated printing code to a local function.

src/kEvent.ml
```diff
@@ -690,6 +879,15 @@ let proved_xml mdl level trans_sys k prop_name =
          | None -> ()
          | Some msg -> Format.fprintf ppf " comment=\"%s\"" msg
       )
+      (match prop.prop_kind with
+      | Property.Invariant -> "valid"
+      | Property.Reachable Some (From ts) -> "unreachable from " ^ string_of_int ts ^ " timesteps"
+      | Property.Reachable Some (Within ts) -> "unreachable within " ^ string_of_int ts ^ " timesteps"
+      | Property.Reachable Some (At ts) -> "unreachable at " ^ string_of_int ts ^ " timesteps"
+      | Property.Reachable Some (FromWithin (ts1, ts2)) -> "unreachable from " ^ string_of_int ts1 ^ " and within " ^ string_of_int ts2 ^ " timesteps"
+      | Property.Reachable None -> "unreachable"
```
daniel-larraz, line 888: Don't use verbose values for XML/JSON. Just use `"unreachable"`.  XML/JSON values should be easily parseable. If we decide to provide info about the bounds (in addition to the result) later, we should do it in a more structured way.

lorchrob:
TODO: Identify properties by type (reach or inv) and position, not expression. Check no two property names are the same.

src/flags.ml
```diff
@@ -2974,8 +2976,12 @@ module Global = struct
         enable_values
         (string_of_enable enable_default_after)
     )
+    
   (* let enable mdl = enabled := mdl :: !enabled *)
-  let enabled () = !enabled
+  let enabled () = 
+      if (List.mem `BMCREACHABLE !enabled |> not) && (List.mem `BMC !enabled)
+      then `BMCREACHABLE :: !enabled
+      else !enabled
```
daniel-larraz, line 2984: Move this check to `kind2Flow` so that the `BMCREACHABLE` module is only added if there is at least one reachability property with a lower bound.

src/induction/base.ml
```diff
@@ -193,18 +223,26 @@ let rec next (input_sys, aparam, trans, solver, k, unknowns) =
...
     let unknowns_at_k, k_true =
       nu_unknowns |> List.partition (
         fun (name,_) -> match TransSys.get_prop_status trans name with
-        | Property.PropKTrue k -> k_int > k
+        | Property.PropKTrue k -> k_int > k 
         | _ -> true
       )
     in
-
+    let k = if not no_skip then skip_steps_next trans solver k else k in
```
daniel-larraz, line 245: This should be moved to line 216. After updating transition system but before filtering properties.

src/induction/base.ml
```diff
@@ -326,6 +396,11 @@ let init input_sys aparam trans =
     (Some (SMTSolver.declare_fun solver)) trans Numeral.zero
   |> SMTSolver.assert_term solver ;
 
+  (* Function 'skip_steps_init' has the side effect of skipping steps
+      for reachability queries if we know we have to unroll the transition
+      relation many times before we can find an example trace. *)
+  let num_skip = if no_skip then 0 else skip_steps_init trans solver in
+
```
daniel-larraz, line 403: This is not necessary, it should always start with `k=0`. The call to `skip_steps_next` in `next` should take care of skipping steps.

PR #977, opened 2023-05-14
Add support for subrange types with an open end

lorchrob:
This PR adds support for declaring partially open intervals. For example:
- `type positive = subrange [1,*] of int;`
- `type negative = subrange [*,-1] of int;`


src/lustre/lustreTypeChecker.ml
```diff
@@ -1425,6 +1429,10 @@ and check_type_well_formed: tc_context -> tc_type -> (unit, [> error]) result
             else Ok ()
       else type_error pos (SubrangeArgumentMustBeConstantInteger e2)
     else type_error pos (SubrangeArgumentMustBeConstantInteger e1)
+    | Some e1, None -> if is_expr_int_type ctx e1 then Ok () else type_error pos (SubrangeArgumentMustBeConstantInteger e1)
+    | None, Some e2 -> if is_expr_int_type ctx e2 then Ok () else type_error pos (SubrangeArgumentMustBeConstantInteger e2)
```
daniel-larraz, line 1433: The condition should also check `is_expr_of_consts ctx e1`. Similarly for `e2`.

lorchrob: Addressed

PR #1003, opened 2023-08-30
Fix bug with subranges

lorchrob:
Fixed a bug with subranges (dealing with int vs enum types, similar to a previous bug fix). For an example of the bug, see the new test file.

### DIFF
--- a/src/lustre/lustreTransSys.ml
+++ b/src/lustre/lustreTransSys.ml
@@ -489,7 +489,7 @@ let add_constraints_of_type init terms state_var =
 
     (* Get bounds of integer range *)
     let l, u = 
-      (if Type.is_array state_var_type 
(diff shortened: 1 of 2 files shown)


src/lustre/lustreTransSys.ml
```diff
@@ -489,7 +489,7 @@ let add_constraints_of_type init terms state_var =
 
     (* Get bounds of integer range *)
     let l, u = 
-      (if Type.is_array state_var_type 
+      (if (Type.is_array state_var_type || Type.is_int_range state_var_type)
```
daniel-larraz, line 492: At this location `Type.is_array state_var_type` is always false. The condition should be simply:
```
Type.is_int_range state_var_type
```

lorchrob: Addressed

PR #1026, opened 2023-10-30
Array length constraints

lorchrob:
Defer checking of array length constraints to the model checking phase


src/lustre/lustreNodeGen.ml
```diff
@@ -1587,6 +1587,28 @@ and compile_node_decl gids is_function cstate ctx i ext inputs outputs locals it
...
+          index_type
+          (Some N.KGhost)
+        in
+        match possible_state_var with
+        | Some state_var -> X.add index state_var accum
+        | None -> accum
+      in let result = X.fold over_indices index_types X.empty in
+      result :: glocals
+    in List.fold_left over_generated_locals glocals gids.GI.array_constraints
+  (* ****************************************************************** *)
```
daniel-larraz, line 1611: Why do you use ad hoc code for these variables? You should add them to the field `locals` in `generatedIdentifiers` at the same time that you add the generated equation.

src/lustre/lustreNodeGen.ml
```diff
@@ -2096,6 +2118,26 @@ and compile_node_decl gids is_function cstate ctx i ext inputs outputs locals it
...
+      let src = Property.Generated (Some pos, [sv]) in
+      a, ac, g, gc, (sv, name, src, Property.Invariant) :: p
+    in
+    let (assumes, _, guarantees, _, props) = 
+      List.fold_left over_array_constraints
+      (assumes, List.length assumes, guarantees, List.length guarantees, props)
+      gids.GI.array_constraints
+    in
+    assumes, guarantees, props
+  (* ****************************************************************** *)
```
daniel-larraz, line 2140: I think only `props` is updated, not `assumes` or `guarantees`. You should simplify the code.

tests/regression/falsifiable/arrays_multidimensional3.lus
```diff
@@ -0,0 +1,11 @@
+node M(const m: int; const n: int) returns (A: [int^m^n, int]);
+let
+  A = {0^2^3, 0};
+tel
+
+node N() returns(A: [int^2^3, int]);
+let
+  A = M(2, 3);
+  check (A = pre A);
```
daniel-larraz, line 9: This check is not legal with the latest fix in develop.

lorchrob: Addressed

src/lustre/lustreNodeGen.ml
lorchrob: Addressed

src/lustre/lustreNodeGen.ml
lorchrob: Addressed

PR #1053, opened 2024-03-05
Fix lustreAstNormalizer assert failure

lorchrob:
Fix bug illustrated in following example: 
```
node imported W(b: bool) returns (a: int; r: bool);

node main(c : bool) returns (t: int; a: int; b: bool);
let
  if c then
    a, b = W(t = 34);
  fi
tel
```


daniel-larraz:
Closed in favor of #1054.

PR #1060, opened 2024-04-03
Desugar any operators to imported function calls when possible (rather than always desugaring to imported node calls)

lorchrob:


src/lustre/lustreAstHelpers.ml
```diff
@@ -720,6 +720,44 @@ let rec vars_without_node_call_ids: expr -> iset =
...
+  | ArrayIndex (_, e1, e2) -> SI.union (calls_of_expr e1) (calls_of_expr e2)
+  (* Quantified expressions *)
+  | Quantifier (_, _, _, e) -> calls_of_expr e
+  (* Clock operators *)
+  | When (_, e, _) -> calls_of_expr e
+  | Condact (_, e1, e2, _, es1, es2) ->
+    SI.flatten (calls_of_expr e1 :: calls_of_expr e2:: (List.map calls_of_expr es1) @ (List.map calls_of_expr es2))
+  | Activate (_, _, e1, e2, es) -> SI.flatten (calls_of_expr e1 :: calls_of_expr e2 :: List.map calls_of_expr es)
+  | Merge (_, _, es) -> List.split es |> snd |> List.map calls_of_expr |> SI.flatten
+  | RestartEvery (_, _, es, e) -> SI.flatten (calls_of_expr e :: List.map calls_of_expr es)
```
daniel-larraz, line 754: `Condact (_, _, _, i, _, _)`, `Activate (_, i, _, _, _)`, and `RestartEvery (_, i, _, _)` call to a node, so you need to include the node identifier in the returned set.

lorchrob: Addressed

PR #1061, opened 2024-04-04
Add support for realizability checks for non-imported nodes

lorchrob:
Add support for realizability checks for non-imported nodes. For a non-imported node, one can now check the realizability of (i) the node's contract (augmented with type information) and (ii) the node's inputs/environment.


src/lustre/lustreGenRefTypeImpNodes.ml
```diff
@@ -0,0 +1,148 @@
...
+        if not (List.mem `CONTRACTCK (Flags.enabled ())) then decl else
+        (* To prevent slicing, we mark generated imported nodes as main nodes *)
+        A.NodeDecl(span, (id, extern, params, inputs, outputs, locals, AnnotMain(Lib.dummy_pos, true) :: node_items, contract)) 
+      in
+      (match node_decl_to_contracts decl2 with 
+      | None -> decl :: acc
+      | Some (decl2, decl3) -> decl :: decl2 :: decl3 :: acc)
+    | A.FuncDecl (span, ((id, extern, params, inputs, outputs, locals, node_items, contract) as decl2)) -> 
+      let decl = 
+        if not (List.mem `CONTRACTCK (Flags.enabled ())) then decl else
```
daniel-larraz, line 139: My understanding is that the generation of imported nodes associated with refinement types is only necessary when `List.mem CONTRACTCK (Flags.enabled ())` is true, right? If that is the case, I would remove all these checks in this file, only call to `gen_imp_nodes` in `lustreInput` if the condition is true, and assume the condition holds in the function has been called.

src/lustre/lustreGenRefTypeImpNodes.ml
```diff
@@ -0,0 +1,148 @@
...
+    | A.ConstDecl (_, TypedConst (_, _, _, ty)) -> 
+      (match (ref_type_to_contract ctx ty None) with  
+      | None -> decl :: acc
+      | Some decl2 -> decl :: decl2 :: acc)
+    | A.TypeDecl (_, FreeType _)
+    | A.ConstDecl (_, UntypedConst _) -> decl :: acc
+    | A.NodeDecl (span, ((id, extern, params, inputs, outputs, locals, node_items, contract) as decl2)) -> 
+      let decl = 
+        if not (List.mem `CONTRACTCK (Flags.enabled ())) then decl else
+        (* To prevent slicing, we mark generated imported nodes as main nodes *)
```
daniel-larraz, line 131: In fact, here you are marking the original nodes as main nodes. Is it really required? If so, update the comment to explain it, and move this comment to `node_decl_to_contracts` and `ref_type_to_contract`.

src/lustre/lustreGenRefTypeImpNodes.ml
lorchrob: Addressed

src/lustre/lustreGenRefTypeImpNodes.ml
lorchrob: Addressed. It is not necessary in this instance, but the other main annotations are necessary. I removed the main annotation and comment in this location, and included the comment in the spots where the main annotations are actually needed.

src/lustre/lustreInput.ml
```diff
@@ -297,7 +305,18 @@ let of_channel old_frontend only_parse in_ch =
...
-          | Some s -> [LustreIdent.mk_string_ident s]
+          | Some s -> 
+            let s_ident = LustreIdent.mk_string_ident s in
+            let main_lustre_node = LN.node_of_name s_ident nodes in
+            (* If checking realizability and main node is not external (imported), then 
+               we are actually checking realizability of Kind 2-generated imported nodes representing 
+               the (1) the main node's contract instrumented with type info and 
+                   (2) the main node's enviornment *)
+            if (not main_lustre_node.is_extern) && List.mem `CONTRACTCK (Flags.enabled ()) then 
+              [s_ident;
```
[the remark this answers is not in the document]
lorchrob: Addressed. The original main node can be omitted.

src/lustre/lustreGenRefTypeImpNodes.ml
```diff
@@ -0,0 +1,143 @@
...
+  | _ -> None
+
+let gen_imp_nodes: Ctx.tc_context -> A.declaration list -> A.declaration list 
+= fun ctx decls -> 
+  List.fold_left (fun acc decl -> 
+    match decl with 
+    | A.TypeDecl (_, AliasType (_, _, ty))
+    | A.ConstDecl (_, FreeConst (_, _, ty))
+    | A.ConstDecl (_, TypedConst (_, _, _, ty)) -> 
+      (match (ref_type_to_contract ctx ty None) with
```
daniel-larraz, line 122: This is the only call to `ref_type_to_contract` and the third argument is `None`. Can we remove the third parameter?

src/lustre/lustreGenRefTypeImpNodes.ml
```diff
@@ -0,0 +1,143 @@
...
+  let node_items = [A.AnnotMain(Lib.dummy_pos, true)] in 
+  let gen_node_id = HString.concat2 (HString.mk_hstring inputs_tag) id in
+  let gen_node_id2 = HString.concat2 (HString.mk_hstring contract_tag) id in
+  let inputs2, outputs2 = 
+    List.map (fun (p, id, ty, cl) -> (p, id, ty, cl, false)) outputs, 
+    List.map (fun (p, id, ty, cl, _) -> (p, id, ty, cl)) inputs 
+  in
+  (* We generate two imported nodes: One for the input node's contract (w/ type info), and another 
+     for the input node's inputs/environment *)
+  Some (A.NodeDecl (span, (gen_node_id, extern, params, inputs2, outputs2, locals, node_items, contract)),
```
daniel-larraz, line 68: Shouldn't `locals` be the empty list? Neither an assumption nor a refinement type can refer to them. Right?

lorchrob: Addressed (you are correct)

src/lustre/lustreGenRefTypeImpNodes.ml
lorchrob: Addressed

PR #1063, opened 2024-04-26
Support forwards and backwards variable references in local variable refinement type predicates

lorchrob:


lorchrob:
This PR also removes support for untyped constants.

PR #1075, opened 2024-06-11
Add --lus_main_type flag and give info about type declarations in LSP info

lorchrob:
1. Add flag --lus_main_type which allows the user to specify a main type for realizability checking
2. Augment the LSP info to include whether or not the main type contains some refinement type

### DIFF
--- a/src/lustre/lustreInput.ml
+++ b/src/lustre/lustreInput.ml
@@ -293,7 +293,19 @@ let of_channel old_frontend only_parse in_ch =
           (* Command-line flag for main node given? *)
           match Flags.lus_main () with
           (* Use given identifier to choose main node *)
-          | Some s -> [LustreIdent.mk_string_ident s]
+          | Some s -> 
+            let s_ident = LustreIdent.mk_string_ident s in (
+            try 
+              let _ = LN.node_of_name s_ident nodes in 
+              [s_ident]
+            (* User-specified main node in command-line input might not exist *)
+            with Not_found -> 
+              let msg =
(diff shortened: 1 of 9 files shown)


src/lustre/lustreInput.ml
```diff
@@ -332,21 +342,35 @@ let of_channel old_frontend only_parse in_ch =
...
-      (* Check that main nodes all exist *)
-      main_nodes |> List.iter (fun mn ->
-        if not (LN.exists_node_of_name mn nodes) then
-          (* This can only happen when the name is passed as command-line argument *)
-          let msg =
-            Format.asprintf "Main node '%a' not found in input"
-              (LustreIdent.pp_print_ident false) mn
-          in
-          raise (NoMainNode msg)
-      ) ;
```
daniel-larraz, line 349: This also covered the case where `old_frontend` is true. This is no longer the case with the new location.

lorchrob: Addressed by handling --lus_main and --lus_main_type analogously in the old frontend. 

Instead of fully repeating the checks performed in the new frontend, I did not support --lus_main_type in the old frontend, nor did I fully support checking Kind 2-generated imported nodes in the old frontend. I did this simply for brevity (which we may value more than supporting all the features, since we don't think the old front-end is being used nowadays); if you want, I can easily update the code to fully support these features.

daniel-larraz: Sounds good. Thanks.

PR #1085, opened 2024-07-31
Imported node generation bug fix

lorchrob:
Stop duplicating type declarations when generating imported nodes.

Note: `LGI.gen_imp_nodes` processes type and constant declarations included in `const_inlined_type_and_consts`, but these declarations should not be included in the result.


lorchrob:
The bug is exhibited by the following lustre file:

```
const C: int;
type UserType = subtype { x: int | x < C and x > 0 }; 

node N(x: UserType) returns (y: UserType)
let
    y = x;
tel
```

when calling with `--enable CONTRACTCK`

PR #1103, opened 2024-10-03
docs: fix references to lustre_main in docs

erooke:
The docs make mention of a `--lustre_main` flag which was replaced with `--lus_main` in 467f66ad88d7422bbef8212c6678ec5c4d7df56a


daniel-larraz:
There is one [more](https://github.com/kind2-mc/kind2/blob/develop/src/lustre/lustreInput.mli#L50) occurrence. Could you please fix that as well?

erooke:
done

PR #1127, opened 2025-01-28
Env realizability imports fix

lorchrob:
Fixes kind2-mc/kind2-dev#150

Also see below for a more in depth test case, which correctly report N's environment as unrealizable because initially, it is not possible for x+2 to be less than 0 and x + (3+l) (which is  x + (3+0) = x+3) to be greater than 0.

However, N2's environment is correctly reported as realizable.

```
contract S1(x, y: int) returns (z: bool);
let
  assume x < 0 and 0 < y;
tel

node M (x: int) returns (y: int)
let
 y = x;
tel

node imported N(x:int) returns (y: bool; y2: int);
(*@contract
  var l: int = 0 -> pre (M(4 + 2));
  var m: int = 3;
  import S1(x+2, x+ (3 + l)) returns (y);
  guarantee y2 + l > 0;
*)

node imported N2(x:int) returns (y: bool; y2: int);
(*@contract
  var m: int = 1;
  var l: int = 2 + m;
  import S1(x+0, x+ (1 + l)) returns (y);
  guarantee y2 + l > 0;
*)
```

Another related note:

Normalization of mode references is complicated (in the normalizer, we normalize every chain of contract calls separately), but this is tricky with our current infrastructure where some generated identifiers are being created earlier in the pipeline (at `lustreGenRefTypeImpNodes.ml`). Due to this, we have temporarily disabled mode references in assumptions for node environment realizability checks (see the assumption in node `N` in the following example). 

```
node H (u:bool) returns (v:bool);
let
   v = not u;
tel;

contract A (a:bool) returns (b:bool) ;
let
  assume a;
  assume H(a);
  guarantee not b;
  mode Am (
      require a;
      ensure not b;
  );
tel

contract B (x:bool) returns (y:bool) ;
let
  var v: bool = not (not x);

  assume H(x);
  import A (x) returns (y);
  mode Bm (
      require x;
      require v;
      require v;
      ensure ::A::Am;
  );
tel

contract C (n:bool) returns (m:bool);
let
    import B (n) returns (m);
    import A (n) returns (m);
tel

node N (u:bool) returns (v:bool);
(*@contract
  var h1: bool = not u;
  import C (h1) returns (v);
  assume ::C::B::A::Am;
*)
let
   v = u;
tel;
```

### DIFF
--- a/src/lustre/lustreGenRefTypeImpNodes.ml
+++ b/src/lustre/lustreGenRefTypeImpNodes.ml
@@ -28,6 +28,37 @@ let unwrap = function
   | Ok x -> x 
   | Error _ -> assert false
 
+let contract_node_decl_to_contracts
+= fun ctx (id, params, inputs, outputs, (pos, base_contract)) -> 
+  let contract' = List.filter_map (fun ci -> 
+    match ci with
+    | A.GhostConst _ | GhostVars _ -> Some ci
+    | A.Assume (pos, name, b, expr) -> Some (A.Guarantee (pos, name, b, expr))
+    | A.ContractCall (pos, name, ty_args, ips, ops) -> 
+      let name = HString.concat2 (HString.mk_hstring ".inputs_") name in
+      Some (A.ContractCall (pos, name, ty_args, ips, ops))
+    | A.Guarantee _ | A.AssumptionVars _ | A.Mode _  -> None
+  ) base_contract in
+  let gen_node_id = HString.concat2 (HString.mk_hstring inputs_tag) id in
+  let inputs2, outputs2 = 
+    List.map (fun (p, id, ty, cl) -> (p, id, ty, cl, false)) outputs, 
+    List.map (fun (p, id, ty, cl, _) -> (p, id, ty, cl)) inputs 
+  in
+  (* Since we are omitting assumptions from environment realizability checks,
+     we need to chase base types for environment inputs *)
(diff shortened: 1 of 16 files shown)


src/lustre/lustreGenRefTypeImpNodes.ml
```diff
@@ -28,15 +29,110 @@ let unwrap = function
...
+        let gen_ghost_var, gids = mk_fresh_ghost_var pos id ghost_var_ty expr in
+        gen_ghost_var,
+        gids
+      ) ips |> List.split in
+      Some (
+        A.ContractCall (pos, name, ty_args, ips', ops'), 
+        List.fold_left GI.union ( GI.empty ()) gids
+      )
+    | A.Guarantee _ | A.AssumptionVars _ | A.Mode _  -> None
+  ) base_contract |> List.split in
```
daniel-larraz, line 77: I think we can extract this code into a new function and call it both here and in `node_decl_to_contracts`. The only difference I see is that the code in `node_decl_to_contracts` is guarded by `Flags.Contracts.check_environment()`. You can keep the guard since it has minimal overhead, but this will make it clearer that the same code is being reused.

daniel-larraz (changes requested):

src/lustre/lustreGenRefTypeImpNodes.ml
lorchrob: Addressed

PR #1128, opened 2025-01-28
Fix .poly_N tag leak in property names

lorchrob:
Bandaid solution to leaking generated node names; a more robust solution will come in a future PR.

Note to Daniel- after #1127 is merged, I will double check this PR. Once I double check, I will post another comment when I think it is ready to merge.


lorchrob:
The PR is now ready to review.

PR #1139, opened 2025-04-23
Arbitrary width machine integers

lorchrob:
Add support for arbitrary-width machine integers (e.g., `int[22]` and `uint[22]`) and a concatenation operation.

There is one limitation of the PR: SMT-LIB level signed bitvector to int casting (see `GenericSMTLIBDriver.ml`). However, this is not critical for now, as the motivation for supporting arbitrary-width bitvectors lies in the moxi interface (where there is no native SMT-LIB support for this anyway), not the Lustre interface.


src/lustre/lustreExpr.ml
```diff
@@ -1574,39 +1574,21 @@ let type_of_numOrBV_numOrBV_numOrBV ?(is_div = false) op t t' =
...
-      | _ -> raise Type_mismatch)
-    | t when Type.is_int16 t -> (
-      match t' with
-      | t when Type.is_int16 t -> Type.t_bv 16
-      | _ -> raise Type_mismatch)
-    | t when Type.is_int32 t -> (
+    | t when Type.is_bitvector t -> (
       match t' with
-      | t when Type.is_int32 t -> Type.t_bv 32
+      | t when Type.is_bitvector t ->
```
daniel-larraz: This match case introduces a new variable `t` that shadows the previous one. Review the rest of cases in the file.

src/lustre/lustreAstHelpers.ml
```diff
@@ -1420,14 +1421,6 @@ and syn_type_equal depth_limit x y : (bool, unit) result =
...
     | Bool _, Bool _
     | Int _, Int _
-    | UInt8 _, UInt8 _
-    | UInt16 _, UInt16 _
-    | UInt32 _, UInt32 _
-    | UInt64 _, UInt64 _
-    | Int8 _, Int8 _
-    | Int16 _, Int16 _
-    | Int32 _, Int32 _
-    | Int64 _, Int64 _
```
daniel-larraz, line 1430: The match cases for `SBitVector` and `UBitVector` are missing.

src/lustre/lustreExpr.ml
```diff
@@ -2958,14 +2912,21 @@ let eval_bvand expr1 expr2 =
   | _ -> Term.mk_bvand [expr1; expr2]
   | exception Invalid_argument _ -> Term.mk_bvand [expr1; expr2]
 
+let eval_bvconcat expr1 expr2 = 
+
+  match Term.destruct expr1, Term.destruct expr2 with              
+    
+  | _ -> Term.mk_bvconcat [expr1; expr2]
+  | exception Invalid_argument _ -> Term.mk_bvand [expr1; expr2]
```
daniel-larraz: `Term.mk_bvand [expr1; expr2]` -> `Term.mk_bvconcat [expr1; expr2]`

src/terms/bitvector.ml
```diff
@@ -655,7 +641,7 @@ let pp_print_signed_machine_integer ppf b =
                | 16 -> bv16_to_num b
                | 32 -> bv32_to_num b
                | 64 -> bv64_to_num b
```
daniel-larraz: Eliminate the explicit match cases as you did for `pp_print_unsigned_machine_integer`.

src/lustre/lustreTypeChecker.ml
```diff
@@ -2031,9 +2046,9 @@ and check_ref_type_assumptions ctx src nname bound_var e =
 and check_type_well_formed: tc_context -> source -> NI.t option -> bool -> tc_type -> ([> warning] list, [> error]) result
   = fun ctx src nname is_const ->
   function
-  | LA.TArr (_, arg_ty, res_ty) ->
-    let* warnings1 = check_type_well_formed ctx src nname is_const arg_ty in
-    let* warnings2 = check_type_well_formed ctx src nname is_const res_ty in 
+  | LA.TArr (_, ty1, ty2) ->
+    let* warnings1 = check_type_well_formed ctx src nname is_const ty1 in
+    let* warnings2 = check_type_well_formed ctx src nname is_const ty2 in
```
daniel-larraz: I think the previous names were more informative. `TArr` is the function ("arrow") type, `arg_ty` means argument type and `res_ty` means result type.

src/lustre/lustreTypeChecker.ml
```diff
@@ -1225,17 +1234,17 @@ and infer_type_unary_op: tc_context -> NI.t option -> Lib.position -> LA.expr ->
 
 and are_args_num: tc_context -> Lib.position -> tc_type -> tc_type -> (bool, [> error]) result
   = fun ctx pos ty1 ty2 ->
+  let* base_ty1 = expand_type_syn_reftype_history ctx ty1 in 
+  let* base_ty2 = expand_type_syn_reftype_history ctx ty2 in
+  match base_ty1, base_ty2 with 
+  | LA.SBitVector _, SBitVector _ -> R.ok true
+  | LA.UBitVector _, UBitVector _ -> R.ok true
+  | LA.SBitVector _, UBitVector _ -> R.ok true
```
daniel-larraz: `are_args_num` is used for `Plus | Minus | Times | Div | Lte  | Lt  | Gte | Gt`. Do we really want to allow operands with different signs?

src/lustre/lustreExpr.ml
lorchrob: Addressed

src/terms/bitvector.ml
lorchrob: Addressed

src/lustre/lustreTypeChecker.ml
lorchrob: Addressed

src/lustre/lustreAstHelpers.ml
lorchrob: Addressed

src/lustre/lustreExpr.ml
lorchrob: Addressed

src/lustre/lustreTypeChecker.ml
lorchrob: My understanding of `are_args_num` was that it was only checking if the args are number types, and not if they were "matching" number types. If they are not "matching" number types, then (as it currently stands) there will still be a type error, but it will be an `ExpectedType` error rather than an `ExpectedNumberTypes` error. So the question is which error message we prefer: (1) `Expected type int[64] but found type uint[64]`, or (2) `Expected both arguments of operator to be of same integer type (or type real) but found uint[64] and int[64]`

lorchrob: If we prefer the second message, I can update `are_args_num` to check if the number types are indeed matching

lorchrob: I think I may have been misunderstanding the original intent of the function. Since the second error message is more informative, I will pursue that, unless you tell me otherwise.

edit -- done

PR #1143, opened 2025-05-01
Cone of influence based slicing

erooke:
The current slicing algorithm is applied at the lustre level. This pull request adds a slicing algorithm which operates at the transition system level.

A new module, `src/lustre/dependencyGraph.ml`, is added for computing the dependence between terms in the transition system.

This module is used to compute the cone of influence of all the properties of the system. Any terms in the transition system not present in the cone of influence are removed.

This changes the flag `--slice_nodes` from a boolean to a flag with three options: `off`, `on`, `experimental`. `off` corresponds to the previous choice false, `on` to the previous choice true, and `experimental` to the slicing algorithm in this pr. Currently the default is left as `on`.

### DIFF
--- a/src/lustre/dependencyGraph.mli
+++ b/src/lustre/dependencyGraph.mli
@@ -1,0 +1,57 @@
+(* This file is part of the Kind 2 model checker.
+
+   Copyright (c) 2025 by the Board of Trustees of the University of Iowa
+
+   Licensed under the Apache License, Version 2.0 (the "License"); you
+   may not use this file except in compliance with the License.  You
+   may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+   implied. See the License for the specific language governing
+   permissions and limitations under the License.
+
+*)
+
+(** Create a dependency graph from a transition system *)
+
+module Vertex : Graph.OrderedType with type t = StateVar.t
+module VarGraph : Graph.S with type vertex = Vertex.t
+
+val dependency_graph_of_system : Term.TermSet.t -> TransSys.t -> VarGraph.t
+(** [dependency_graph_of_system definition_set trans_sys] constructs a
+    dependency graph representing the dependencies between state variables in
+    the given transition system.
+
+    @param definition_set A set of terms which are to be treated as dependencies
+    @param trans_sys
+      The transition system from which dependencies are extracted.
+    @return
+      A graph representing variable dependencies. The state variable `sv1`
+      depends on `sv2` if there is the edge `sv1 -> sv2` *)
+
+val cone_of_influence :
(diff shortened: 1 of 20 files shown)


erooke:
The ci build failure is due to `Seq.find` being used. This was added in ocaml 4.14 while ci is using 4.09. Is there any stated version of ocaml we target?

daniel-larraz:
We target OCaml 4.09

src/utils/graph.ml
```diff
@@ -429,4 +435,8 @@ module Make (Ord: OrderedType) = struct
 
   let to_vertex_list: vertices -> vertex list = VSet.elements
   (** returns a list of vertex *)
+
+  let to_edge_list = fun edge -> List.map (fun element -> (fst element, snd element))  (ESet.elements edge)
```
daniel-larraz: You can simplify this definition to `let to_edge_list = ESet.elements`

erooke: Not sure how I missed that on the initial pass, fixed.

src/lustre/dependencyGraph.mli
```diff
@@ -0,0 +1,57 @@
...
+    @param trans_sys
+      The transition system from which dependencies are extracted.
+    @return
+      A graph representing variable dependencies. The state variable `sv1`
+      depends on `sv2` if there is the edge `sv1 -> sv2` *)
+
+val cone_of_influence :
+  VarGraph.t -> StateVar.StateVarSet.t -> StateVar.StateVarSet.t
+(** [cone_of_influence properties dependency_graph] constructs the set of state
+    variables which are depended upon by [properties]
```
daniel-larraz: The signature's description is no longer accurate.

src/transSys.mli
```diff
@@ -699,6 +699,16 @@ val enforce_constantness_via_equations : t -> (t * StateVar.t list)
 (** Return the global constant state variables of the transition system *)
 val global_const_state_vars : t -> StateVar.t list
 
+(** [slice_system sys vars] returns a new system obtained by restricting
+    [sys] to only the state variables in [vars].
+
+    @param sys The original system.
+    @param vars The set of state variables to retain in the sliced system.
+    @return A new system containing only the specified state variables.
```
daniel-larraz: Update this doc.

erooke: Clarified this in: https://github.com/kind2-mc/kind2/pull/1143/commits/816704f31d18b5c7b49e7da6eb52024d50443937

src/lustre/dependencyGraph.mli
erooke: I'm not sure I follow what is not accurate here. I have clarified that the dependency is dictated by the dependency graph in https://github.com/kind2-mc/kind2/pull/1143/commits/e2fbba09dd515106635434e05bb0b4425f3d6ed7

daniel-larraz: The doc starts with:
> [cone_of_influence properties dependency_graph] constructs the set of state variables which are depended upon by [properties]

The `dependency_graph` parameter is now the first parameter and the `properties` parameter is now named `roots`.

erooke: Woops, missed that. Fixed: https://github.com/kind2-mc/kind2/pull/1143/commits/6eeba69464ddfcf2e0aee4d629fb118f6115adfb

PR #1160, opened 2025-05-21
Generalize bitvector invariant generation

lorchrob:
Previously, there was a hardcoded invariant generation engine for each hardcoded bitwidth (int8, uint8, int16, uint16, and so on). However, this is not sufficient to support arbitrary-width bitvectors. In this PR, we 
1) parameterize the invariant generation engine by the bit width
2) scan the input problem for a list of relevant bit widths
3) spawn copies of the invariant generation engine for only these relevant bit widths


src/inputSystem.ml
```diff
@@ -990,6 +992,82 @@ function
...
+    let subsystems = List.map fst checks in 
+    let sources = List.map (fun subsys -> subsys.S.source) subsystems in 
+    let state_vars = Lib.concat_map TransSys.state_vars sources in 
+    List.fold_left over_svar IntSet.empty state_vars
+  | Native sub -> 
+    let subsystems = S.all_subsystems sub in
+    let sources = List.map (fun subsys -> subsys.S.source) subsystems in 
+    let state_vars = Lib.concat_map TransSys.state_vars sources in 
+    List.fold_left over_svar IntSet.empty state_vars
+  | Horn _ -> IntSet.empty
```
daniel-larraz: This code could be factored out. Could you create an auxiliary function parameterized with `over_svar`, and reduce `get_bv_sizes` and `get_ubv_sizes` to a call to this new function?

src/invgen/invGenDomain.ml
```diff
@@ -228,55 +228,21 @@ module MakeMachineInteger(M: MachineIntegerParam): Domain = struct
   )
 end
 
-module Int8Miner: MachineIntegerMiner = struct
-  let name = "Int8"
-  let mine = InvGenMiner.Int8.mine
+module BVMiner(IS : sig val length : int end): MachineIntegerMiner = struct
+  let name = "BV"
```
daniel-larraz: How about adding the length value to the name?

src/inputSystem.ml
lorchrob: Addressed

src/invgen/invGenDomain.ml
lorchrob: Addressed

lorchrob:
Most recent commit addresses kind module ID clash issue

PR #1167, opened 2025-06-02
Machine integer casts

lorchrob:
Support cast operators with arbitrary-width bitvector types. This includes generalizing the SMT-LIB encoding of cast operators. Notably, we transition to using the new SMT-LIB operator `sbv_to_int` rather than hardcoding `intN_to_int` for fixed values of `N`. However, not all solvers/solver versions support the new cast operators, so we use a different encoding for those versions that don't support it.


src/flags.ml
```diff
@@ -3920,7 +3923,15 @@ let solver_dependant_actions solver =
...
-      Some (major_rev, minor_rev, patch_rev)
+      let result = Some (major_rev, minor_rev, patch_rev) in 
+      (match Smt.solver () with
+        | `Z3_SMTLIB -> 
+          if
+            (major_rev > 4) ||
+            (major_rev = 4 && (minor_rev > 14 || (minor_rev = 14 && patch_rev >= 1)))
+          then support_new_bv_cast_operators' := true;
+        | _ -> ());
+      result
```
daniel-larraz: I would keep `get_version` independent of a particular solver and instead set `support_new_bv_cast_operators'` within the corresponding `match` block below.

src/flags.ml
```diff
@@ -3958,7 +3969,7 @@ let solver_dependant_actions solver =
   )
   | `Z3_SMTLIB -> (
     let cmd = Format.asprintf "%s -version" (Smt.z3_bin ()) in
-    match get_version false cmd with
+    match get_version true cmd with
```
daniel-larraz, line 3964: Changing the argument from `false` to `true` has no effect unless you also use `patch_rev` to set `support_new_bv_cast_operators'` here, which I assume was your intention.

src/lustre/lustreParser.messages
```diff
@@ -3253,3 +3145,239 @@ Syntax Error!
 main: FUNCTION ASSUME LPAREN RPAREN RETURNS LPAREN RPAREN LET ASSERT ASSUME IN DECIMAL WEAKLY
 
 Syntax Error!
+
+one_expr: UINT8 XOR
+
+<YOUR SYNTAX ERROR MESSAGE HERE>
```
daniel-larraz: Can you replace `<YOUR SYNTAX ERROR MESSAGE HERE>` by `Syntax Error!` in the file?

src/lustre/lustreSimplify.ml
```diff
@@ -210,16 +210,10 @@ let rec eval_ast_expr bounds ctx =
     | A.ConvOp (pos, A.ToInt, expr) -> eval_unary_ast_expr bounds ctx pos E.mk_to_int expr 
 
   (* Conversion to unsigned fixed-width integer numbers [uint8-uint64 expr] *)
```
daniel-larraz: Update comment

src/lustre/lustreSimplify.ml
```diff
@@ -210,16 +210,10 @@ let rec eval_ast_expr bounds ctx =
...
-    | A.ConvOp (pos, A.ToUInt8, expr) -> eval_unary_ast_expr bounds ctx pos E.mk_to_uint8 expr
-    | A.ConvOp (pos, A.ToUInt16, expr) -> eval_unary_ast_expr bounds ctx pos E.mk_to_uint16 expr
-    | A.ConvOp (pos, A.ToUInt32, expr) -> eval_unary_ast_expr bounds ctx pos E.mk_to_uint32 expr
-    | A.ConvOp (pos, A.ToUInt64, expr) -> eval_unary_ast_expr bounds ctx pos E.mk_to_uint64 expr
+    | A.ConvOp (pos, A.ToUBV n, expr) -> eval_unary_ast_expr bounds ctx pos (E.mk_to_ubv n) expr
 
   (* Conversion to signed fixed-width integer numbers [int8-int64 expr] *)
```
daniel-larraz: Update comment

src/induction/compress.ml
```diff
@@ -282,7 +282,8 @@ let equal_mod_input only_bv unc_inputs accum s1 s2 =
             let term_of_state s =
               let n =
                 if only_bv then
-                  Term.mk_ubv (Bitvector.num_to_ubv64 (aux s))
+                  let width = Term.sufficiently_large_bit_width (aux s |> Numeral.to_int) |> Numeral.of_int in
+                  Term.mk_ubv (Bitvector.num_to_ubv (Numeral.of_int 64) width)
```
daniel-larraz:
```suggestion
                  let n' = aux s in
                  let width = Term.sufficiently_large_bit_width (n' |> Numeral.to_int) |> Numeral.of_int in
                  Term.mk_ubv (Bitvector.num_to_ubv width n')
```

src/terms/termLib.ml
```diff
@@ -198,9 +198,9 @@ let logic_of_flat fun_symbols t acc =
   | App (s, _) when Symbol.(s == s_to_int || s == s_to_real || is_divisible s) ->
     sup_logics acc |> add LA |> add IA |> add RA
 
-  | App (s, _) when Symbol.(is_to_uint8 s || is_to_uint16 s || is_to_uint32 s || is_to_uint64 s ||
+  (* | App (s, _) when Symbol.(is_to_uint8 s || is_to_uint16 s || is_to_uint32 s || is_to_uint64 s ||
```
daniel-larraz: The `logic_of_flat` function is used to determine the required SMT-LIB logic for a given expression. We should include a match case for conversions from arbitrary-width bitvectors, adding the BV logic when such a conversion is detected.

src/terms/term.ml
```diff
@@ -323,7 +334,9 @@ let bitvector_of_term t = match destruct t with
       Symbol.ubitvector_of_symbol s
 
   | T.Const s when Symbol.is_numeral s ->
-      Bitvector.num_to_ubv64 (Symbol.numeral_of_symbol s)
+      let num = Symbol.numeral_of_symbol s in
+      let width = sufficiently_large_bit_width (num |> Numeral.to_int) |> Numeral.of_int in
+      Bitvector.num_to_ubv (Numeral.of_int 64) width
```
daniel-larraz:
```suggestion
      Bitvector.num_to_ubv width num
```

src/terms/term.ml
```diff
@@ -307,6 +307,17 @@ let rec numeral_of_term t = match destruct t with
   | _ -> invalid_arg "numeral_of_term"
 
 
+(* Given input n, return an integer representing a bit width sufficiently 
+   large to store n as an unsigned machine integer. *)
+let sufficiently_large_bit_width (n : int) : int =
+  if n = 0 then 1
```
daniel-larraz:
```suggestion
  if n < 0 then invalid_arg "sufficiently_large_bit_width: input must be non-negative"
  else if n = 0 then 1
```

daniel-larraz (changes requested):

src/terms/term.ml
lorchrob: Addressed

src/lustre/lustreSimplify.ml
lorchrob: Addressed

src/lustre/lustreSimplify.ml
lorchrob: Addressed

src/lustre/lustreParser.messages
lorchrob: Addressed

src/flags.ml
lorchrob: Addressed

src/flags.ml
lorchrob: I'm confused -- I did use `patch_rev` to update `support_new_bv_cast_operators` (see [this line](https://github.com/kind2-mc/kind2/pull/1167/files#diff-c386885891e71387d92566e8fda6741b1329a8206339a6241eda229bc6eb8a51R3973)), no?

src/terms/term.ml
lorchrob: Addressed

src/induction/compress.ml
lorchrob: Addressed

src/flags.ml
daniel-larraz: After your recent commit bbce41c addressing the other issue, this change now makes sense :)

src/terms/termLib.ml
lorchrob: Here, do you mean conversions *to* arbitrary-width bitvectors? The old code (and in the integer case) is only checking casts *to* the type of interest, not *from*. 

Addressed under that assumption, but let me know if otherwise.

daniel-larraz: Yes, I meant _to_. But actually, I think we need to cover _from_ as well.

lorchrob: Actually, the *from* case seems a bit unclear -- for example, the `TO_BV` symbol casts ints to bitvectors, but it could also cast a bitvector to a bitvector. My guess is that if we are coming *from* an integer, then that will be detected from the logics of the subterms (in `acc`), so maybe no change is required?

daniel-larraz: We should add `LA` when converting from machine integers to mathematical integers. For example, consider the following input:
```
function N(x,y: uint8) returns ();
let
  check (int x) < (int y);
tel
``` 
Kind 2 fails on this input.

lorchrob: Addressed

src/lustre/lustreExpr.ml
```diff
@@ -1905,46 +1772,23 @@ let eval_to_int expr =
         (Numeral.of_big_int
            (Decimal.to_big_int
               (Symbol.decimal_of_symbol s)))
-    | Term.T.Const s when Symbol.is_ubv8 s ->
-      Term.mk_num
-        (Bitvector.ubv8_to_num
-          (Symbol.ubitvector_of_symbol s))
```
daniel-larraz, line 1911: This code performs a conversion from an 8-bit unsigned integer literal (a known concrete value) to an integer, instead of delegating the conversion to the solver. We should retain this optimization for arbitrary bit widths (for signed and unsigned integers).

lorchrob: Addressed

daniel-larraz (approved):

PR #1177, opened 2025-06-16
Empty maps

lorchrob:
Support the empty map constructor.

Note to Daniel: There is one internal comment marked `(*!! ... *)` with a minor question to be checked before merging.


lorchrob:
The aforementioned internal comment was resolved and removed.

src/lustre/lustreAbstractInterpretation.ml
```diff
@@ -537,6 +538,7 @@ and interpret_int_expr node_id ctx ty_ctx proj expr =
       | _ -> assert false
     in
     extract_bounds_from_type output_ty
+  | EmptyMap _
```
daniel-larraz, line 541: Shouldn't we extract the bounds of the map value type?

daniel-larraz: I was mistaken. Please disregard this.

PR #1179, opened 2025-06-18
Fix lus_main and lus_main_type flag bugs

lorchrob:
After the changes to node IDs, `--lus_main` and `--lus_main_type` (before this PR) exhibited some buggy behaviors. 

1. Line 402 (before changes) accidentally omitted `~node_type:Type`, so in this case, the constructed node ID could not be found in the list of lustre nodes, throwing an error. 
2. Say you have a refinement type `T` but pass `--lus_main T` (rather than `--lus_main_type T`) to Kind 2. Previously, an internal error was thrown in `subsystem_of_nodes` rather than a nice error message. This is because the check on line 354 (calling `node_of_input_name`) only looks to see if there is some component with a matching name, and not whether the component type (node or type declaration) is correct. To address this, I insert calls to `node_of_node_id` to trigger the `Not_found` exception in all necessary cases. 

You can compare the behavior of calling `--lus_main T`, `--lus_main_type T`, `--lus_main N`, and `--lus_main_type N` before and after changes in the following example: 

```
type T = int;

node N (x: T) returns (y: T)
let
    y = x;
    check x = y;
tel
```

### DIFF
--- a/src/lustre/lustreInput.ml
+++ b/src/lustre/lustreInput.ml
@@ -358,12 +358,16 @@ let of_channel old_frontend only_parse in_ch =
                     (2) the main node's enviornment, if environment checking is enabled *)
               let main_nodes = 
                 if (not main_lustre_node.is_extern) && List.mem `CONTRACTCK (Flags.enabled ()) then 
-                  [NI.mk_node_id ~node_type:Contract (HString.mk_hstring s)]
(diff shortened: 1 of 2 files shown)


lorchrob:
For context: This issue is related to the code block Josh mentioned in the Kind 2 meeting for his work on the extension

PR #1182, opened 2025-06-30
Map element update

lorchrob:
Note to Daniel: we should talk about this PR on Tuesday (July 1) before you review


src/lustre/lustreExpr.ml
```diff
@@ -1251,6 +1251,24 @@ let mk_index_var i =
...
+          :: I.reserved_scope)
+       |> HString.mk_hstring)
+     kt 
+    |> Term.mk_var
+  in
+
+  (* create lustre expression for free variable*)
+  { expr_init = v;
+    expr_step = v;
+    expr_type = Type.t_int }
```
daniel-larraz: Shouldn't the expression be of type `kt`?

lorchrob: Yes. Addressed

PR #1186, opened 2025-07-11
Map element update 2

lorchrob:
Reimplement map element update in terms of 'inductive' equation


src/lustre/lustreAst.mli
```diff
@@ -207,7 +207,7 @@ type struct_item =
   | TupleSelection of position * ident * expr
   | FieldSelection of position * ident * ident
   | ArraySliceStructItem of position * ident * (expr * expr) list
-  | ArrayDef of position * ident * ident list
+  | ArrayDef of position * ident * ident list * lustre_type (* key type *)
```
lorchrob: Note to self -- undo this change as discussed with Daniel on Tuesday

PR #1193, opened 2025-07-30
Generate slicing certificates

erooke:
Adds a new flag `--certif_slicing` which will generate and verify a certificate that the slicing we performed did not change the behavior of the system. This branch is based off of https://github.com/kind2-mc/kind2/pull/1143


erooke:
~~Build error has to do with a bad rebase, figuring that out now.~~ Fixed

PR #1205, opened 2025-09-02
Update frame block initialization semantics

lorchrob:
Before this PR, frame block initializations were always supplied to the desugared equation at the top level of the RHS expression. In this PR, we update the semantics to only supply the initialization value in branches of if blocks where variables are left undefined. Consider the following example.
```
node N (c: bool) returns (y: int)
let
  frame (y)
  y = 0;
  let 
    if c then 
      y = 1;
    fi
  tel
tel
```
Previously, this would desugar to `y = 0 -> if c then 1 else pre y`. With the updated semantics, this desugars to `y = if c then 1 else 0 -> pre y`. (Notice the initialization is now only supplied to the missing `else` case of `y`'s definition.)


daniel-larraz:
@lorchrob There is a conflict with recent merges, could you please resolve it? Also, please update the user documentation as well as the documentation for the `desugar_frame_blocks` and `desugar_node_item` functions to reflect the new semantics.

lorchrob:
> @lorchrob There is a conflict with recent merges, could you please resolve it? Also, please update the user documentation as well as the documentation for the `desugar_frame_blocks` and `desugar_node_item` functions to reflect the new semantics.

Addressed both

lorchrob:
> @lorchrob ... as well as the documentation for the `desugar_frame_blocks` and `desugar_node_item` functions to reflect the new semantics.

Addressed this portion of the comment, which I missed in the previous commit

PR #1207, opened 2025-09-04
Supported new element update syntax for tuples, records, and arrays

lorchrob:
There is only one obstacle with supporting the new syntax. Consider expression `e[k := v]`. Here, we do not know at parse time whether to treat `k` as a `MapIndex`, `Index`, or `Label`. So, I added a new index constructor `GenericIndex` for use at parse time. At type checking, we use the inferred type of `e` to determine the correct index type. At normalization time, the index is replaced within the AST (from `GenericIndex` to one of the preexisting constructors).


lorchrob:
UPDATE -- support for the old syntax (with `with`) has been removed

src/lustre/lustreDesugarFrameBlocks.ml
```diff
@@ -155,6 +155,7 @@ let rec fill_ite_helper frame_pos node_id lhs id fill = function
               | A.Label (a, b) -> A.Label (a, b)
               | MapIndex (a, e) -> MapIndex (a, fill_ite_helper frame_pos node_id lhs id fill e)
               | Index (a, e) -> Index (a, fill_ite_helper frame_pos node_id lhs id fill e)
+              | GenericIndex _ -> assert false
```
daniel-larraz: This is reachable since desugaring happens before normalization.

src/lustre/lustreTypeChecker.ml
```diff
@@ -1394,6 +1398,24 @@ and check_type_expr: tc_context -> NI.t option -> LA.expr -> tc_type -> ([> warn
...
+    match ty with 
+    | LA.TupleType _ 
+    | LA.ArrayType _ -> Ok (LA.Index (pos, e2))
+    | LA.Map _ -> Ok (LA.MapIndex (pos, e2))
+    | LA.RecordType _ -> (
+      match e2 with 
+      | LA.Ident (pos, id) -> Ok (LA.Label (pos, id))
+      | _ -> type_error pos (IlltypedUpdateWithIndex ty) 
+      ) 
+    | _ -> type_error pos (IlltypedUpdateWithLabel ty)
```
daniel-larraz: I believe these two type errors need to be swapped: `IlltypedUpdateWithIndex` is the generic one, while `IlltypedUpdateWithLabel` applies specifically to records.

daniel-larraz (changes requested):

src/lustre/lustreTypeChecker.ml
lorchrob: I think this comment is half correct. I believe the first error message should stay as is. Consider the following.
```
type R = struct { f1: int; f2: real; };

node N(inp: R; x: int) returns (r: R);
let
  r = inp[(x+3) := 3]; 
  check r.f1 = 3;
tel
```
The inferred type of `inp` is a record type, and yet `(x+3)` cannot be interpreted as a record field value. So, we correctly produce the error `Expected a tuple, array, or map type but found R { f1: int; f2: real }` (with a position mapping to `inp`) (ie, expression `x+3` leads us to expect `inp` to have a non-record type, but indeed it is a record type). If we swapped the messages, we would get `Expected a record type but found R { f1: int; f2: real }`, which doesn't make sense.

However, consider the latter case. Since `x`'s inferred type is `int`, it cannot be updated. So we should expect a more generic error message: `Expected a tuple, array, map, or record type but found int`. I added this error message and applied it in this case.
```
type R = struct { f1: int; f2: real; };

node N(inp: R; x: int) returns (r: R);
let
  r = x[f1 := 3]; 
  check r.f1 = 3;
tel
```

src/lustre/lustreDesugarFrameBlocks.ml
lorchrob: Addressed

PR #1216, opened 2025-09-15
Use default values for out of bounds accesses in arrays

lorchrob:


lorchrob:
`hash_ty` is currently unused, but I figured it may be useful in the future. Let me know if I should remove it from this PR.

PR #1223, opened 2025-10-03
Sets

lorchrob:
Support minimal feature set required for the new test cases `empty_set.lus`. This includes 
1. The `set` lustre type
2. The empty set term `{}@<lustre_type>`
3. The set membership predicate `element in set_expression`

Some notes about this PR
* Sets compile identically to maps, but without the array of map values (only the boolean flag array denoting presence or absense)
* Some test cases had variables named "set" which I had to alter since it is now a keyword
* I chose to overload the `in` predicate between sets and maps. Similarly to index accesses (which are now overloaded between maps and arrays), the types of the operands are (now) needed during compilation. So, that information is now stored in the AST. However, the types are not known at parse time, so they are filled later in the pipeline once the types are known.
* There were cases, particularly in `lustreAstHelpers.ml`, where the `EmptyMap` terms (and now, analogously, `EmptySet` terms) were not handled properly -- the types should have been recursed over, but they were ignored (an example is collecting all the variable names in an expression). I extended these functions to recurse over the types for both the `EmptySet` and `EmptyMap` cases. However, I suspect the treatment is still not perfect -- none of these types take type checking context as input, but that is necessary to resolve type aliases and so on. So, if there are `UserType`s, they cannot be chased for now. I think understanding and addressing this issue in full generality is outside the scope of this PR.


src/lustre/typeCheckerContext.mli
```diff
@@ -292,7 +292,7 @@ val type_contains_abstract : tc_context -> tc_type -> bool
 val type_contains_array: tc_context -> tc_type -> bool
 (** Returns true if the lustre type expression contains an array *)
 
-val type_contains_map: tc_context -> tc_type -> bool
+val type_contains_map_or_set: tc_context -> tc_type -> bool
 (** Returns true if the lustre type expression contains a map *)
```
daniel-larraz: Update documentation line: `contains a map or a set`

src/lustre/lustreAstHelpers.mli
```diff
@@ -193,3 +193,7 @@ val name_of_prop : Lib.position -> HString.t option -> LustreAst.prop_kind -> HS
 (** Get the name associated with a property *)
 
 val get_const_num_value : expr -> int option
+
+val fold_lustre_ty : (expr -> 'a) -> 'a -> ('a -> 'a -> 'a) -> lustre_type -> 'a
+
+val map_lustre_ty : (expr -> expr) -> lustre_type -> lustre_type
```
daniel-larraz, line 202: Can you document this functions?

src/lustre/lustreNodeGen.ml
```diff
@@ -2217,6 +2243,33 @@ and compile_node_decl gids_map is_function opac cstate ctx node_id ext params in
...
+    let over_empty_sets acc (id, _) =
+      let eq_lhs, _ = compile_struct_item (A.SingleIdent (Lib.dummy_pos, id)) in 
+      let eq_lhs = flatten_list_indexes eq_lhs in
+      (* extract index for boolean flag denoting presence or absence of map item *)
+      (*!! No longer necessary? *)
+      (*let eq_lhs = X.fold (fun k sv acc -> match k with 
+      | X.TupleIndex 0 :: _ -> X.add k sv acc 
+      | _ -> acc 
+      ) eq_lhs X.empty 
+      in*)
```
daniel-larraz: Should we remove this?

daniel-larraz (changes requested):

src/lustre/lustreNodeGen.ml
lorchrob: Addressed

src/lustre/lustreAstHelpers.mli
lorchrob: Addressed

src/lustre/typeCheckerContext.mli
lorchrob: Addressed

PR #1225, opened 2025-10-15
Non-empty set literals

lorchrob:
Support singletons (more generically, support set literals with cardinality > 0).

The implementation is analogous to the implementation of map element updates applied to the empty map (e.g., syntax `{ x, y, z }@<K>` is supported analogously to (pseudocode) `map[x := (), y := (), z := ()]@<K, unit>`. In other words, the implementation can be viewed as set insertions to the empty set.

Note: The `Union` binary operator appears in this PR as an artifact of my earlier "optimization" (which is now removed), but nothing interesting happens with it -- it is unsupported and simply triggers an assertion failure in `lustreAstNormalizer` and `lustreNodeGen` (though the latter should be unreachable)


src/lustre/lustreTypeChecker.ml
```diff
@@ -1553,11 +1570,22 @@ and infer_type_binary_op: tc_context -> NI.t option -> Lib.position
           (type_error pos (UnificationFailed (ty1, ty2))))
       | Ok _, Ok _ -> (type_error pos (ExpectedIntegerTypes (ty1, ty2)))
       | Error id, _ | _, Error id -> (type_error pos (UnboundIdentifier id)))
-  | LA.Plus | LA.Minus | LA.Times | LA.Div ->
+  | LA.Plus -> 
+    let* b1 = are_args_num ctx pos ty1 ty2 in 
+    let* ty1' = expand_type_syn_reftype_history_subrange ctx ty1 in 
+    let* ty2' = expand_type_syn_reftype_history_subrange ctx ty1 in
```
daniel-larraz: Typo?
```suggestion
    let* ty2' = expand_type_syn_reftype_history_subrange ctx ty2 in 
```

daniel-larraz (changes requested):

src/lustre/lustreTypeChecker.ml
lorchrob: Yes, it was a typo. Now addressed

daniel-larraz (approved):

PR #1227, opened 2025-10-17
Set unions

lorchrob:
Support set unions (overloading `+`)

### DIFF
--- a/src/lustre/lustreNodeGen.ml
+++ b/src/lustre/lustreNodeGen.ml
@@ -1216,8 +1216,7 @@ and compile_ast_expr
     compile_binary bounds E.mk_minus expr1 expr2
   | A.BinaryOp (_, A.Plus, expr1, expr2) ->
     compile_binary bounds E.mk_plus expr1 expr2
-  | A.BinaryOp (p, A.Union, _, _) ->
-    fail_at_position p "Set unions not yet supported"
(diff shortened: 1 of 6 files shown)


src/lustre/lustreNodeGen.ml
```diff
@@ -2358,6 +2357,28 @@ and compile_node_decl gids_map is_function opac cstate ctx node_id ext params in
     List.fold_left over_set_insertions [] gids.GI.set_insertions
   in
   let gequations = gequations @ set_insertions_eqs in
+  let set_union_eqs = 
+    let over_set_unions acc (id, nexpr1, nexpr2, fresh_idx_name, _) =
+      (* Desugar to lhs[i] = if i = nexpr2 then true else i in nexpr1 *)
+      (* Desugar to lhs[i] = i in nexpr1 or i in nexpr2 *)
```
daniel-larraz: I believe only the second comment about desugaring makes sense.

daniel-larraz (changes requested):

src/lustre/lustreNodeGen.ml
lorchrob: You are correct; line 2363 was meant to replace line 2362. Addressed

daniel-larraz (approved):

PR #1244, opened 2025-11-06
Global assumptions

lorchrob:
Partial support for global assumptions with the `assume` keyword. 

For full support, we need to (at least)
1. Make sure no generated names appear in counterexamples
2. Make sure we aren't breaking other features, e.g. inductive validity cores 

For the above, it may be necessary to introduce a new `declaration` constructor in the Lustre AST.

Note -- to avoid shift/reduce conflicts, I had to promote `assume` to a proper keyword (it can't be used as a variable/constant/function/node name). We have not yet determined whether the same treatment should be applied to the other contract "keywords"


lorchrob:
We decided not to pursue global assumptions

PR #1245, opened 2025-11-07
Add support for global assumptions

daniel-larraz:
This PR supersedes PR #1244 by addressing some of its limitations.


lorchrob (approved):

lorchrob:
We decided not to pursue global assumptions

lorchrob:
We will try to implement something analogous here with lemmas (as in Dafny)

PR #1252, opened 2025-11-17
Use base type to instantiate polymorphic nodes

daniel-larraz:


daniel-larraz:
Closing this PR because we decided that instantiation should occur on the original type. However, one possible improvement in compositional mode would be to use the guarantee `G1` unconditionally in the included test, since this guarantee is independent of the refinement-type constraints of the arguments.

PR #1258, opened 2025-12-08
New contract syntax

lorchrob:
Support the new contract syntax that doesn't require `(*@contract ... *)` where contract items are included along with (strictly before) the locals. To avoid shift-reduce conflicts, we now use `gvar` and `gconst` (rather than the previously-overloaded `var` and `const`) for ghost constants and variables, respectively.

See `tests/regression/success/contract_syntax.lus` and `tests/regression/error/contract_syntax.lus`


lorchrob:
Previous commit addresses the migration of the user docs to the new contract syntax. (We still mention the old syntax, and note that it is still available but no longer actively supported)

PR #1260, opened 2025-12-10
Update tuple indexing syntax

lorchrob:
Update tuple indexing syntax from `t.%idx` to `t[idx]`, where `idx` must be a concrete natural number in range.

The bracket syntax is now shared for tuples, maps, and arrays, so we don't know at parse time the access type. So, we have to discriminate these cases later in the pipeline. (We were already doing this for maps and arrays.)

I also noticed we didn't have a user doc pages on tuples. Since we have our own syntax that is probably not the same as Lustre V4 or V6, I added a page.

### DIFF
--- a/doc/usr/source/2_input/9_tuples.rst
+++ b/doc/usr/source/2_input/9_tuples.rst
@@ -1,4 +1,4 @@
-Enumeration types
+Tuple types
 ==================
 
 Tuples are constructed with the syntax ``'(x1, ..., xn)`` and destructed with the syntax ``t[idx]``, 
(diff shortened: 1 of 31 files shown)


doc/usr/source/2_input/9_tuples.rst
```diff
@@ -0,0 +1,14 @@
+Enumeration types
```
daniel-larraz:
```suggestion
Tuple types
```

lorchrob: Addressed

PR #1288, opened 2026-02-02
Re-introduce `choose`

lorchrob:
Before this PR, we inferred whether to desugar `any` operators to either imported nodes or imported functions based on whether or not the input type was temporal (respectively). After the change to make lustre functions mathematical by default (deterministic), this is no longer ideal. For example, a contract definition with an `any` operator may be imported in both a function context and in a node context; however, automatically desugaring to the right case is nontrivial (for the former, we would need an imported function; for the latter, it should be up to the user). 

To address this, we added back `choose`, and now the `any` operator always desugars to an imported node, and the `choose` operator to an imported function. If an `any` operator is used in the context of a function, then an error is thrown.


doc/usr/source/2_input/1_lustre.rst
```diff
@@ -1276,19 +1281,21 @@ the system model. As a result, Kind 2 proves the property P1 valid.
...
      check "P1" z > 0 and z < 0;
    tel
 
 This problem is mitigated by the possibility for
 the user to check that the predicate ``P(x)`` in
-the ``any`` expression is realizable.
-This is possible because, for each ``any`` expression occurring in
+the ``any`` (or ``choose``) expression is realizable.
+This is possible because, for each ``any`` (or ``choose``) expression occurring in
 a model, Kind 2 introduces an internal imported node whose
```
daniel-larraz: Clarify Kind 2 introduces an imported _function_ in the case of a `choose` expression.

src/lustre/lustreSyntaxChecks.ml
```diff
@@ -641,6 +646,7 @@ let rec expr_only_supported_in_merge observer expr =
   | Quantifier (_, _, _, e) 
   | StructUpdate (_, e, _, None) -> r observer e
   | AnyOp (_, _, e) -> r false e
+  | ChooseOp (_, _, e) -> r false e (*!! Why was this `false` with `AnyOp`? *)
```
daniel-larraz: I think this was a mistake. We should use `r observer e` instead.

src/lustre/lustreDesugarAnyChooseOps.ml
```diff
@@ -61,28 +62,25 @@ fun ctx node_name fun_ids expr ->
       | None -> assert false
     ) inputs in
     let name = mk_fresh_fn_name pos node_name in
```
daniel-larraz: The name and node_type should be different depending on whether it is a `choose` or an `any` expression.

lorchrob:
When reporting results, we need to use `choose` for choose operators.
Update node name, node type, and printing.

src/lustre/lustreSyntaxChecks.ml
lorchrob: Addressed

doc/usr/source/2_input/1_lustre.rst
lorchrob: Addressed

lorchrob:
> When reporting results, we need to use `choose` for choose operators. Update node name, node type, and printing.

Addressed

PR #1290, opened 2026-02-10
Remove the option to compile to Rust

daniel-larraz:
After revamping test generation to stop producing Rust oracles, with that functionality replaced by a new contract monitor, we decided to remove the option to compile to Rust.


omrischwarz:
The Rust compilation option is what has had me follow this project. I take it there has not been much interest? 
That surprises me because there's a company (Picturus.com) whose product is a Simulink-like interface for producing Rust code, and they are doing quite well

daniel-larraz:
@omrischwarz As a matter of fact, there hasn't been much interest in this feature among other Kind 2 users. Are you an active user of this feature? Are you interested in transpiling to Rust specifically, or to any compilable programming language?

In this project, we are more focused on the formal verification of high-level specifications than on compiling them into executable code. Since our resources are limited, we must allocate our efforts strategically. Moreover, because Kind 2 is open-source, removing this feature should not prevent others from taking the current code to adapt and extend the transpiler to meet their needs. If the interest is not specifically in Rust, there are other tools available that can produce executable code from a Lustre model, such as the [Lustre v6](https://www-verimag.imag.fr/lustre-v6.html) compiler, the [Vélus](https://velus.inria.fr/) compiler, or [Heptagon](https://gitlab.inria.fr/synchrone/heptagon).

omrischwarz:
My interest was mercenary. I was looking into paths for migrating a large Simulink code base to another environment as a way of addressing vendor-lock-in at my company.

daniel-larraz:
Thank you for your feedback.

As mentioned, our current priorities lie in developing other parts of the tool. However, previous versions of Kind 2 that include this feature will remain available for anyone who wishes to use them.

PR #1294, opened 2026-02-19
MCS Cut Property Position Output

Jabe03:
Added:
- Line, column, and file of the property invalidated by the MCS.
- Function to get the position of a property.

Updated JSON schema to reflect these changes.


daniel-larraz (changes requested):
Could you also update the XML output and schema file?

daniel-larraz (approved):

PR #1295, opened 2026-02-20
Remove old tuple syntax

Jabe03:
Removed old tuple syntax support

- Tuple input like `"t": ("36", false, "5.0")` is no longer supported
- Only accept tuples of the form `t: {"0" : 36, "1" : false, "2": 5.0}`
- Lifted the dependency constraint for Yojson: now compatible with any version
- Removed mention of the old syntax from the docs


Jabe03:
It was requested that, when we see an input file with the old tuple syntax with the new version of Yojson that rejects it, we give an error message saying which variable this error is associated with. This is not a value interpretation error, but an invalid JSON syntax error, so Kind 2 does not make it to the place where we are creating the values for variables when the old syntax is used; Yojson rejects it when you load it from the filename. So I gave an error message that does not expose the internal Yojson error, but gives the same information.

PR #1298, opened 2026-02-25
Support type annotation for the `pre` operator

lorchrob:
1. There are two internal comments intentionally left in the code (starting with `(*!!`). They are both places where the existing code looked weird to me. Could you take a look? They are both pretty simple questions.
2. In the normalizer, I added a boolean option argument `force_prop` to `mk_fresh_subrange_constraint`. If set to true, Kind 2 will necessarily treat the generated constraint as non-candidate (which is what we want if we have a type annotation with a subrange type)


src/lustre/lustreAstHelpers.ml
```diff
@@ -1278,8 +1307,8 @@ let rec replace_with_constants: expr -> expr =
    | RestartEvery (p, i, es, e) ->
       RestartEvery (p, i, List.map replace_with_constants es, replace_with_constants e)
 
-  (* Temporal operators *)
-  | Pre (_, e) -> replace_with_constants e
+  (* Temporal operators *) (*!! Was this case supposed to peel off the `Pre`? *)
```
daniel-larraz: `replace_with_constants` is only called from `abstract_pre_subexpressions`. For that use case, it does not matter whether the `Pre` is peeled off. However, the function should not peel it off because that behavior goes beyond simply replacing identifiers with constants and prevents the function from being reused in cases where only replacement is required.

src/lustre/lustreAstDependencies.ml
```diff
@@ -709,8 +714,9 @@ let rec vars_with_flattened_nodes: node_summary -> int -> LA.expr -> LA.SI.t
     let call_vars = r (Call (pos, [], node_id, es)) in
     SI.union (r clk_exp) call_vars
 
-  (* Temporal operators *)
-  | Pre (_, _) -> SI.empty
+  (* Temporal operators *) 
+  (*!! Was this case supposed to not recurse on the subexpression? *)
```
daniel-larraz: This may be intentional, or at least it has an impact. The `vars_with_flattened_nodes` function is only used to collect variables that are not under a `Pre`, in order to detect dependencies on the current values of variables. All calls to `vars_with_flattened_nodes`, except one, take as an argument an expression that is the result of applying `abstract_pre_subexpressions`. For those cases, it does not matter. However, there is [one call](https://github.com/lorchrob/kind2/blob/90bf903ed3829bdc6ed7bd2946ba051e1e2e608b/src/lustre/lustreAstDependencies.ml#L1193) where the expression is passed directly, and thus the subexpression of a `Pre` could contain a variable identifier. In that particular case, simply not recursing can achieve the same behavior as passing the result of `abstract_pre_subexpressions`. But this raises the question of whether we should recurse in `vars_with_flattened_nodes` and ensure we pass the result of calling `abstract_pre_subexpressions` to that call, so that both functions achieve their goals independently, or whether we should keep the current behavior and verify if `abstract_pre_subexpressions` is really required in the other cases. For now, we can keep it as is, add a comment, and investigate this question later.

src/lustre/lustreAstHelpers.ml
lorchrob: Addressed

src/lustre/lustreAstDependencies.ml
lorchrob: Addressed

src/lustre/lustreAstHelpers.ml
```diff
@@ -1368,7 +1398,10 @@ let rec abstract_pre_subexpressions: expr -> expr = function
       RestartEvery (p, i, List.map abstract_pre_subexpressions es, abstract_pre_subexpressions e)
 
   (* Temporal operators *)
-  | Pre (p, e) -> Pre(p, replace_with_constants e)
+  | Pre (p, e, None) -> Pre (p, replace_with_constants e, None)
+  | Pre (p, e, Some ty) -> 
+    let ty = map_lustre_ty abstract_pre_subexpressions ty in
+    Pre (p, replace_with_constants e, Some (map_lustre_ty abstract_pre_subexpressions ty))
```
daniel-larraz, line 1404: You are calling `abstract_pre_subexpressions` over `ty` twice.

lorchrob:
We decided to use a general type ascription operator and not support type annotations for all built-ins (unless there is demand to actually put in this effort)

PR #1303, opened 2026-02-27
Disallow undefined output variables

lorchrob:
Add check to disallow undefined node and function outputs. 

After adding this check (and the corresponding `output_var = any@<ty>;` equations to test cases), test file `tests/regression/falsifiable/double_mono.lus` was throwing an exception. The root cause was that monomorphizations were being generated in such an order that some node was called before it was defined in the declaration list. 

We have faced the same problem before, and I spent a good deal of time trying to find the error in the order of the generated declarations. After enough time, I decided that the most robust solution is to simply re-sort the declarations after monomorphization (after these bugs, I don't know if there is any straightforward way to generate all the monomorphizations with the correct order by construction). 

Re-sorting simply involves a call to `sort_and_check_nodes_contracts` from the dependency analysis module. However, I also extended the module to check the generated equations for dependencies, as they may be hidden there.

### DIFF
--- a/src/lustre/lustreAstDependencies.ml
+++ b/src/lustre/lustreAstDependencies.ml
@@ -606,7 +606,6 @@ let rec  mk_decl_map: LA.declaration option IMap.t -> LA.t -> ((LA.declaration o
   | LA.NodeParamInst _ :: _-> Lib.todo __LOC__
 (** builds an id :-> decl map  *)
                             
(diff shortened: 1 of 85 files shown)


lorchrob:
UPDATE: Rather than re-sorting the declarations (due to the monomorphization step bug), we take another strategy: Initially separate generated monomorphizations from the user declarations, then merge them together. While merging, insert each monomorphization directly before its polymorphic base node (e.g., `N<int>` before `N`).

PR #1304, opened 2026-03-04
Type ascription operator

lorchrob:
Support type ascription operator `(e : ty)`. 

1. Kind 2 will check that `e` satisfies the base type of `ty` in the type checker
2. If `ty` contains some refinement type or subrange type, the normalizer will generate the corresponding constraint and then erase the ascription.

In principle, Kind 2's language theoretically supports type annotations for all built-in polymorphic operators (`pre`, `arrow`, and so on). However, implementing support for all these type annotations would require major changes (in terms of LOC), and we don't envision that they will be very practically useful. As an alternative, we support this general type ascription operator to check that a given expression has a given type. 

For example, instead of `pre@<int> x` (which we don't currently support), one can do `pre (x: int)`.


daniel-larraz:
Kind 2 reports `5:35: Unbound identifier 'y'`  for the following example:
```
node N(x,y: int) returns (z: int);
let
  z = (x : subtype { v: int | v < y });
  check z < y;
tel
```

lorchrob:
> Kind 2 reports `5:35: Unbound identifier 'y'` for the following example:
> 
> ```
> node N(x,y: int) returns (z: int);
> let
>   z = (x : subtype { v: int | v < y });
>   check z < y;
> tel
> ```

Addressed

PR #1305, opened 2026-03-05
Check map index accesses respect map key types

lorchrob:


lorchrob:
Closed (for now) because semantically we decided to infer base types for non-built-in operators, so we don't do checks here to be consistent

PR #1307, opened 2026-03-10
Correctly handle refinement types in ghost constants and variables

lorchrob:


lorchrob:
From Zulip convo with Daniel

"The issue is that slicing does not preserve properties when the node/function is abstract in the analysis. This behavior made sense until now. However, we would now like certain properties to be preserved, specifically, those generated from proof obligations in the context of the contract, but not those from the body. We will need to distinguish between these two cases.

I pushed a change with an approximate solution to the problem. It is not the actual solution because, I explained above, for functions/nodes that have contract and body, we only want to keep the properties during slicing when using compositional analysis if they were generated as a result of a proof obligation generated in the contract, but not if it is a check in the body of the function/node."

To be clear, the current fix is over-conservative in the sense that it does not slice some properties that it should.

I will mark this PR as a draft until an exact fix is implemented.

More advice from Daniel:
"Yes, it [the failing test case] is related to the issue I mentioned above. A property in the body is included in the analysis (which checks whether the contract is exhaustive), but it should not be. As a consequence, in this case the property is proved invalid because there is no body in that check.

BTW, regarding the use of the Local source: this helps us obtain an approximate solution. However, for the actual solution we may keep Ghost as the source. Then, when we generate the actual properties for the Lustre node, if we see that a property comes from an imported node (and it is Ghost), we treat it as a proof obligation.

I believe the actual change should be made in setting the source of the Property module. Classifying the property simply as Generated is not sufficient; we also need to distinguish the context in which it appears: contract or body.

Once you have the right classification, you filter properties during slicing. Contract -> new behavior. Body -> old behavior."

lorchrob:
Another comment (of mine) from Zulip:
"we should mark the source as Ghost, but then in LustreNodeGen, if source is Ghost and the node is imported, then create a Generated property, where the Generated constructor includes more information to distinguish between generated-from-body and generated-from-contract"

lorchrob:
TODO: Revert changes to XML/JSON output to not show source of Generated

lorchrob:
> TODO: Revert changes to XML/JSON output to not show source of Generated

Addressed

PR #1319, opened 2026-03-31
Fix interaction between type annotations and laziness

lorchrob:


lorchrob:
Not necessary due to #1325

PR #1320, opened 2026-04-01
Refinement type generated quantified variable checks

lorchrob:
Disallow cases where refinement type bound variables desugar to quantified variables in ways we can't handle
1. Quantified variable under pre
2. Quantified variable in call to non-inlinable node/function

Refinement type bound variables become quantified when under set/array/map elements


lorchrob:
This PR will be subsumed by a more general PR that disallows temporal operators (or maybe just `pre`) and quantifiers over refinement type bound variables in nested types

lorchrob:
Rather than having this PR be subsumed by another PR, I simply generalized the checks. 

1. Refinement type bound variables in nested types under pre or arrow
2. Refinement type bound variables in nested types being passed to non-inlinable nodes or functions

lorchrob:
Most of the code in this PR was generated by AI, with my guidance (it took a handful of iterations of me giving feedback). 

The generalization of the checks was performed manually.

lorchrob:
After discussion with Daniel, we updated the checks again.

1. Refinement type bound variables in nested types under pre or arrow
2. Refinement type bound variables in nested types being passed to non-inlinable calls
3. *Any* node call under nested types 

Points 1 and 3 are due to our semantics of container/nested types -- they are streams of containers, not containers of streams. So within the container are scalar values, not streams, and hence temporal operators and node calls (stream transformers) are not type correct. I could imagine lifting some of these restrictions in the future though, as one can view (eg) `[subtype {x: int | x > pre x}, int]` as "desugaring" to `subtype { t: [int, int] | t[0] > pre t[0] }`. However, for now, we do not perform this implicit desugaring.

Functions are mathematical, so they can be applied to streams or scalars. Point 2 is a purely technical limitation that our infrastructure cannot currently handle.

lorchrob:
The GitHub UI shows a very large diff for the `check_type_well_formed` function. Most of this diff is introducing a local helper recursive function `check_type_well_formed_rec` to track whether the type is nested, and in most places the only difference is an extra level of indentation.

PR #1323, opened 2026-04-02
Type ascription fixes

lorchrob:
I noticed two problems with type ascriptions. Both are addressed in this PR.

1. The check for when to generate a node (rather than a function) was overly conservative. The `expr_is_const` function (which we used to use) only returns true if the expression is literally some `Const _`; really, we want to check that the type does not contain any expression with a temporal operator or node (non-function) call. (If there is a temporal operator, a node is required because, for now, we require that function interfaces be non-temporal.)

2. Since type ascriptions may generate node (not function) calls, the generated nodes may not be inlinable. So, we need to handle cases where symbolic array indices or quantified variables are passed to type ascription generating non-inlinable calls. For now (in this PR), these cases are rejected.

The test case `falsifiable/type_ascription.lus` (modified version in this PR) helped me identify both these problems. Upon modification, the over-conservative check (point 1) led to a node being generated rather than a function. Then, we did not notice that a quantified variable was being passed to a non-inlinable node (point 2), and an exception was thrown. Addressing point 1 allows us to support the input, as we can just generate an inlinable function. However, for the new test cases, we must generate a node, so an error is thrown.


lorchrob:
Updates to the above description:

* The above problems can be addressed by *always* generating nodes for type ascriptions, but then inlining them anyway. We know we can inline because we always have a definition.
* We need to reject temporal operators in type ascriptions in the context of a function (will do in this PR after this comment)
* We need to filter type ascription nodes from realizability checks (will do in this PR after this comment)

Issue #1229, opened 2025-10-20
Wrongly detected Cyclic dependency when using a `pre` inside an `any`


elipparini:
```
node C () returns();
var v1 : int;
let
    v1 = any {v1_tmp: int |v1_tmp >= (0 -> pre v1)}; 
    check not(v1=1);
tel
```
         
produces the following error

 `Cyclic dependency detected in definition of identifiers: v1 -> v1`

although we are simply imposing monotonicity on v1 and there is no cyclic dependency.

daniel-larraz:
The expression `any {v1_tmp: int | v1_tmp >= (0 -> pre v1)}` is syntactic sugar for a call to an imported node defined as follows:

```
node imported AnyV1(v1: int) returns (v1_tmp: int);
(*@contract
  guarantee v1_tmp >= (0 -> pre v1);
*)
```

Since version 2.2.0, Kind 2's causality analysis *conservatively* assumes that each output stream of an *imported* node or function depends on the current values of all its inputs, and rejects any model that does not satisfy this condition. This stricter check facilitates modular reasoning about the realizability of the full system, at the cost of rejecting specifications that actually have a valid executable realization. Although this may seem very stringent, in practice it is usually possible to refine the model to avoid such false positives.

In your case, replacing the equation for `v1` with:
```
v1 = any {v1_tmp: int | v1_tmp >= (0 -> pre v1_tmp)};
```
does the trick.

That said, I will update the error message to clarify that it indicates a *potential* cycle.

elipparini:
Hi @daniel-larraz , thanks for the explanation.

Regarding the trick you suggested, I see that in this simple case it works. But if we have something slightly more complicated, e.g.

```
node C (b : bool) returns();
var v1 : int;
let
    if b then
        v1 = any {v1_tmp: int |v1_tmp >= (0 -> pre v1)}; 
    else
        v1 = (0 -> pre v1) - 1; 
    fi
    check b => v1>=0;
tel
```

then it does not, as `v1` could reach say `-10`, but whenever `b` is `true` we would be imposing `v1` to be (at least) `>= 0` (because `pre v1_tmp` is not bounded to be equal to `pre v1`), hence missing the values between `-10` and `-1`.

The tool indeed will return that `b => v1>=0` is valid, although b = (false, true, ...) gives a counterexample on the original code.

Is there any simple way to make it work?

daniel-larraz:
Yes, you can modify the model as follows:
```
[...]

PR #1327, opened 2026-04-10
Cyclic dependency message

Jabe03:
Changed:
- When doing dependency analysis, we now check if an imported node caused one of the transitions in the cycle, if there is one
- If there is an imported node, mark the error as a "potential" cyclic dependency
- This error may be caused by Kind 2's necessarily conservative assumption that all outputs of an imported node depend on all of the inputs
- Try to explain the "cause" of the error
  - If this node was auto generated by the any or choose operator, keep the name anonymous and report the error `use of the any/choose operator`
  - If not, report the name of the imported node whom may have been overlooked by the user when writing the model
- "cause" here means the potential disconnect between the user's understanding and the actual semantics of Kind 2

This provides a more informative error message for the example reported in issue #1229.


src/lustre/lustreAstDependencies.ml
```diff
@@ -55,6 +56,19 @@ let error_message error = match error with
     ^ (Lib.string_of_t (Lib.pp_print_list LA.pp_print_ident ", ") (SI.elements ids))
   | CyclicDependency ids -> "Cyclic dependency detected in definition of identifiers: "
     ^ (Lib.string_of_t (Lib.pp_print_list LA.pp_print_ident " -> ") ids)
+  | ImportedCyclicDependency (ids, imported_id) -> 
+    let imported_node_type = NI.get_node_type imported_id in
+    let imported_node_name = NI.get_user_name imported_id in
+
+    let helping_message = match imported_node_type with 
+      | Any -> "Cyclic dependency is likely caused by the use of the any operator" 
+      | Choose -> "Cyclic dependency is likely caused by the use of the choose operator"
```
daniel-larraz: the 'any' operator
the 'choose' operator

src/lustre/lustreAstDependencies.ml
```diff
@@ -55,6 +56,19 @@ let error_message error = match error with
...
+    let imported_node_type = NI.get_node_type imported_id in
+    let imported_node_name = NI.get_user_name imported_id in
+
+    let helping_message = match imported_node_type with 
+      | Any -> "Cyclic dependency is likely caused by the use of the any operator" 
+      | Choose -> "Cyclic dependency is likely caused by the use of the choose operator" 
+      | _ ->  "Cyclic dependency is likely caused by a call to the imported node " 
+        ^ Lib.string_of_t LA.pp_print_ident imported_node_name in
+
+    "Potential Cyclic dependency detected in definition of identifiers: "
```
daniel-larraz: Potential Cyclic dependency -> Potential cyclic dependency

daniel-larraz (changes requested):
Some minor fixes in the error message

daniel-larraz (approved):

PR #1328, opened 2026-04-10
Type ascription user doc and syntax-test.lus

lorchrob:
Unrelated changes in this PR: Requiring outputs to be defined broke `syntax-test.lus`, so I also fixed that here. Analogously, I updated `tests/conftest.py` to include `./examples/syntax-test.lus` in the regression suite so a similar problem does not happen in the future.


daniel-larraz:
Can you make the test script check syntax-test.lus to avoid this problem in the future?

lorchrob:
> Can you make the test script check syntax-test.lus to avoid this problem in the future?

Addressed

PR #1331, opened 2026-04-27
Disallow temporal operators and node calls in map and set literals

lorchrob:


lorchrob:
This is too restrictive.

We can view (eg) set literal constructors as taking streams of scalars as input and producing as output streams of sets (and analogously for other nested types).

PR #1332, opened 2026-04-30
Remove subranges

lorchrob:


src/lustre/lustreInput.ml
```diff
@@ -237,9 +231,9 @@ let type_check declarations =
       LCF.constants_to_calls new_func_ids const_inlined_nodes_and_contracts
     in
 
-    (* Step 20. Normalize AST: guard pres, abstract to locals where appropriate *)
+    (* Step 11. Normalize AST: guard pres, abstract to locals where appropriate *)
```
daniel-larraz: This should be `Step 19. Normalize AST: ...`

daniel-larraz (changes requested):

src/lustre/lustreInput.ml
lorchrob: Addressed

daniel-larraz (approved):

PR #1334, opened 2026-05-05
Stack crash normalizer bug

lorchrob:
I learned about the crash bug (see new test case). It involves a bug with the normalization of lazy operators. I implemented a potential fix, but I don't know enough to tell if the fix is correct. So, I added comments explaining what I learned, and the fix I implemented. 

The comments should be read in order (start with `(*!! 1. <comment> *)`, then `(*!! 2. <comment> *)`, then `(*!! 3. <comment> *)`. Hopefully they provide enough description for you to understand the bug, how I fixed it, and why I am unsure whether the fix is correct.

For now, the PR still contains debug printing. I will remove it once we are confident in the fix (either this fix, or a different one).

If it doesn't make sense, let me know, and we can discuss it over Zoom.


daniel-larraz:
Issue addressed in PR #1335.

PR #1345, opened 2026-05-15
Show property constraints in property summary

Jabe03:
Added:

- Printing of property constraints when printing summary of properties to the command line
- New JSON field `expr` showing the constraint of a property when its JSON object is outputted
- New command line argument `--show_props (true|false)` that enables/disables property constraint printing

Changed: 

- Type ascription node input names now gets "unfolded" when printed as part of a property
   -  `(.inp >= 1)` gets unfolded to `((tN + 1) >= 1)` when generated by `(tN + 1 : Pos)`
- Type ascription properties get renamed from their internal name when they are printed
   -  `type_ascription_l68c7[l68c7]` gets renamed to `TypeAscription[l68c7]`


src/flags.ml
```diff
@@ -3662,8 +3662,21 @@ let check_nonvacuity_default = true
...
   let log_format_json () = Log.get_log_format () = Log.F_json
 
-  (** ************************************************************ **)
+  (* Use add_format_spec instead of add_spec *)
+
+  (* JSON log. *)
+  let show_props_default = Log.get_show_props ()
 
+  let _ = add_spec
+    "--show-props"
```
daniel-larraz: We use `_` to separate words in option names. Please rename this option to `--show_props`.

daniel-larraz (changes requested):

src/lustre/lustreNodeGen.ml
```diff
@@ -2653,39 +2668,57 @@ and compile_node_decl gids_map is_function opac cstate ctx node_id ext params in
   (* Generate Contract Constraints for Refinement Type Constraints      *)
   (* ****************************************************************** *)
   in let (assumes, guarantees, props) =
-  let create_constraint_name rexpr = 
-    Format.asprintf "@[<h>%a@]" A.pp_print_expr rexpr
+  let create_constraint_name_pos node_id (pos : position)= 
+    match NI.get_node_type node_id with
+    | _ -> Format.asprintf "@[<h>SubType%a@]" pp_print_line_and_column pos
```
daniel-larraz: Remove the unnecessary `match` statement.

src/lustre/lustreNodeGen.ml
```diff
@@ -2617,31 +2631,32 @@ and compile_node_decl gids_map is_function opac cstate ctx node_id ext params in
...
-          a, ac, g, gc, (sv, name, src, Property.Invariant) :: p
+          a, ac, g, gc, (sv, name, src, Property.Invariant, rexpr) :: p
         | _ -> assert false
       else
-        let name = create_constraint_name rexpr in
+        let name = create_constraint_name "Generated" pos in
         let src = Property.Generated (Some pos, [sv], Property.Body) in
         let src = Property.Candidate (Some src) in
-        a, ac, g, gc, (sv, name, src, Property.Invariant) :: p
+        a, ac, g, gc, (sv, name, src, Property.Invariant, rexpr) :: p
```
daniel-larraz: All these constraints should be named `SubType` because they are generated from subranges that are subtypes.

src/lustre/lustreNodeGen.mli
```diff
@@ -19,6 +19,7 @@
   
   @author Andrew Marmaduke *)
 
+
```
daniel-larraz: Clean up whitespace.

src/lustre/lustreNodeGen.ml
```diff
@@ -2653,39 +2668,57 @@ and compile_node_decl gids_map is_function opac cstate ctx node_id ext params in
...
+  | Some nid -> NI.Map.find_opt nid global_teas
+  | None -> None
+  in
+
+  let rexpr = match node_id_opt, replace_expr with
+    | Some _, Some expr ->
+      (* Format.printf "Found type ascription with a mapped RHS for %a: %a\n" NI.pp_print_node_id_input_name nid A.pp_print_expr expr ; *)
+      LustreAstHelpers.substitute_naive (HString.mk_hstring ".inp") expr rexpr
+    | Some _, None -> 
+      (* Format.printf "Didn't find mapped type ascription for %a. \n" NI.pp_print_node_id_input_name nid;
```
daniel-larraz: Clean up debug comments.

src/lustre/lustreInput.mli
```diff
@@ -148,3 +148,4 @@ val ast_of_file : string -> (LustreAst.t, [> error]) result
    indent-tabs-mode: nil
    End: 
 *)
+
```
daniel-larraz: Clean up whitespace

daniel-larraz (changes requested):

src/lustre/lustreNodeGen.ml
daniel-larraz: The "temporary fix" (using `SubRange` instead of `SubType`) is the correct solution until PR #1332 is merged.

src/lustre/lustreNodeGen.ml
```diff
@@ -2653,39 +2668,52 @@ and compile_node_decl gids_map is_function opac cstate ctx node_id ext params in
...
+  in
+
+  let rexpr = match node_id_opt, replace_expr with
+    | Some _, Some expr ->
+      LustreAstHelpers.substitute_naive (HString.mk_hstring ".inp") expr rexpr
+    | Some _, None -> 
+      rexpr
+    | None, _ ->
+      rexpr
+    in
```
daniel-larraz: Given the definition of `replace_expr`, you can simplify this code:

```suggestion
  let rexpr = match replace_expr with
    | Some expr ->
      LustreAstHelpers.substitute_naive (HString.mk_hstring ".inp") expr rexpr
    | None -> rexpr
  in
```

src/lustre/lustreNodeGen.ml
```diff
@@ -1688,22 +1693,25 @@ and compile_contract cstate gids ctx map contract_scope node_scope contract =
...
     in List.map over_guarantees guarantees
       |> List.map (fun g -> g, false)
   in assumes @ assumes2,
     guarantees @ guarantees2
 
 and compile_node_decl gids_map is_function opac cstate ctx node_id ext params inputs outputs locals items contract =
   let gids = NI.Map.find node_id gids_map in
+  let global_teas = 
+    List.fold_left (fun acc (_,gds) -> NI.Map.union (fun _ a _ -> Some a) acc gds.GeneratedIdentifiers.type_ascription_exprs) NI.Map.empty (NI.Map.bindings gids_map) 
+  in
```
daniel-larraz: Do we really need to compute the union of the `type_ascription_exprs` of all nodes? Shouldn't the looked-up node ID always be in `gids.GI.type_ascription_exprs`?

daniel-larraz (changes requested):

src/lustre/lustreNodeGen.ml
Jabe03: It seems like we do. The issue is that we are not aware of which node contains the type ascription call. That is the node that would have the type_ascription_exprs that contains the original expression that has the ascription. I have made a change so that we now look up which node has the expression substitution, instead of always merging the type_ascription_exprs together.

daniel-larraz (approved):

PR #1395, opened 2026-07-12
Fix internal Type_mismatch in equality over maps with collection values

daniel-larraz:
## Problem

Structural equality over a value of type `map<K, set<E>>` makes Kind 2 exit with an internal error, misleadingly reported on the input-file channel:

```
<Error> Error opening input file '...': LustreExpr.Type_mismatch
```

A typical trigger is a map-of-sets state variable assigned by a node call under a `when` clock inside a frame block: the clocked-call machinery generates held-value consistency definitions containing exactly such an equality (see the added regression test for a 13-line reproduction).

## Cause

`LustreNodeGen.compile_equality` compiles equality over an indexed type leaf-by-leaf, and guards map-value leaves with the map's domain (presence) arrays collected in `acc_guard`. Each guard was selected with the **full** index-variable list of the current leaf. For `map<K, set<E>>` the value leaf has two index positions (key and set element) while the domain guard has one, so the second select is applied to a `Bool` term and `LustreExpr.type_of_select` raises `Type_mismatch`.

## Fix

Select each guard with only the prefix of index variables matching its own arity. Leaves whose guards have the same number of index positions (all previously working cases) compile exactly as before.

## Testing

- Added `tests/regression/success/repro_map_of_sets_clocked.lus`; it fails with the internal error on unpatched `develop` and verifies with the fix (both checked locally against this base).
- Exercised at scale on a set of distributed-protocol models (quorum-based leader election with `map<HostId, set<HostId>>` state, two-phase commit, sharded KV, Paxos): all verify with the fix, with results identical to equivalent formulations that avoid map-of-sets state.

### DIFF
--- a/src/lustre/lustreNodeGen.ml
+++ b/src/lustre/lustreNodeGen.ml
@@ -1174,13 +1174,19 @@ and compile_ast_expr
           | E.Unbound _ -> 
             acc
         ) E.t_true (List.combine idx_vars bounds) in
-        (* For map value equality we only consider m1[k] = m2[k] for k in the maps. 
-           `acc_guard` collects the constraints that k is in the map (if the equality is over maps) *)
-        let acc_guard' = List.fold_left (fun acc e -> 
-          let e = List.fold_left (fun acc arr_i -> 
+        (* For map value equality we only consider m1[k] = m2[k] for k in the maps.
+           `acc_guard` collects the constraints that k is in the map (if the equality is over maps).
+           A guard array may have fewer index positions than the current leaf (e.g. the
+           domain array of a map whose values are sets, which adds an element index), so
+           select each guard with only the prefix of index variables matching its arity. *)
+        let acc_guard' = List.fold_left (fun acc e ->
+          let guard_arity =
+            List.length (Type.all_index_types_of_array (E.type_of_lustre_expr e))
+          in
+          let e = List.fold_left (fun acc arr_i ->
             E.mk_select_and_push acc arr_i
-          ) e arr_is in
-          E.mk_and acc e 
+          ) e (fst (list_split guard_arity arr_is)) in
+          E.mk_and acc e
         ) E.t_true acc_guard in
         (* For equality:    forall (x: K) conditions =>  arr1[x]  = arr2[x] 
            For disequality: exists (x: K) conditions and arr1[x] <> arr2[x]. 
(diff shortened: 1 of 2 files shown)


daniel-larraz:
The test `tests/regression/success/map_value_equality_nested_guards.lus`, included in PR #1396, also reproduces this issue without using a `when` block.

src/lustre/lustreNodeGen.ml
```diff
@@ -1174,13 +1174,19 @@ and compile_ast_expr
           | E.Unbound _ -> 
             acc
         ) E.t_true (List.combine idx_vars bounds) in
-        (* For map value equality we only consider m1[k] = m2[k] for k in the maps. 
-           `acc_guard` collects the constraints that k is in the map (if the equality is over maps) *)
-        let acc_guard' = List.fold_left (fun acc e -> 
-          let e = List.fold_left (fun acc arr_i -> 
+        (* For map value equality we only consider m1[k] = m2[k] for k in the maps.
```
lorchrob, line 1177: To make the comment more human-understandable, I would just say 
`(* For map value equality we only consider m1[k] = m2[k] for k in the maps.
           "acc_guard" collects the constraints that k is in the map 
           (if the equality is over maps).
           "guard_arity" (and its usage) ensures that we only use indices associated 
           with the map key type and not its value type (eg, consider map<int, set<int>>). *)`

tests/regression/success/repro_map_of_sets_clocked.lus
```diff
@@ -0,0 +1,32 @@
+-- Minimal reproduction of a frontend bug (now fixed): a map-of-sets state
```
lorchrob: I think the comment above (in `lustreNodeGen.ml`) explains the test case well enough, and we can remove the comment here (probably it would confuse me if I were try to read it later). But I do not feel strongly.

src/lustre/lustreNodeGen.ml
daniel-larraz: I agree. Changed.

tests/regression/success/repro_map_of_sets_clocked.lus
daniel-larraz: Removed.

lorchrob (approved):

PR #1396, opened 2026-07-12
Fix dropped domain guards in equality over nested map values

daniel-larraz:
Stacked on #1395 (it relies on the guard-arity handling introduced there); this branch is based on `fix-map-value-equality-guard-arity` and GitHub should retarget it to `main` automatically when #1395 merges.

## Problem

Structural equality over maps whose values contain further maps evaluates incorrectly: two structurally equal values of type `map<int, map<int,int>>` compare as **unequal** (falsifiable at k=0). The same applies to maps nested deeper inside tuples, records or ADT payloads in a map's value type.

## Cause

`LustreNodeGen.compile_equality` guards each map-value leaf so that entries are compared only at keys present in the map. The guards are collected by the trie fold, which can recognize only the **first** leaf as a domain (presence) array — a map's domain/value pair is encoded with plain tuple indices, which are ambiguous with actual tuples, so deeper domain arrays cannot be identified from trie indices alone. Inner value leaves were therefore compared at keys outside the inner domains, where the SMT arrays contain unconstrained junk.

## Fix

Compute the guard association from the **source type** of the equality operands: for every map in the (alias-expanded) type, collect the trie prefix of its value subtree together with the prefix of its domain leaf; a leaf under a value prefix is guarded by the domain arrays of all its enclosing maps. Actual tuples are distinguished from the map pair encoding by the type itself, so e.g. a `bool` field of a real tuple inside a map value is never mistaken for a domain guard. To make operand types inferable also in normalizer-generated equations (e.g. the held-value consistency definitions of clocked calls), the typing context is extended with the node interface, locals, and generated locals. If the type of neither operand can be inferred, the previous first-leaf guard is kept as a fallback.

## Testing

- Added `tests/regression/success/map_value_equality_nested_guards.lus` covering nested-map equality and disequality plus the real-tuple-in-map-value disambiguation; it fails on this base without the fix (`nested map equality` falsifiable at k=0) and passes with it. ADT payloads containing maps are also fixed but not expressible with the current `main` grammar; equivalent probes pass on a branch with the extended ADT payload syntax.
- The regression suite cases from #1395 still pass, and a set of distributed-protocol models exercising map/set-heavy state (two-phase commit, quorum leader election, sharded KV, Paxos) verify with results unchanged.

### DIFF
--- a/src/lustre/lustreNodeGen.ml
+++ b/src/lustre/lustreNodeGen.ml
@@ -1174,13 +1174,19 @@ and compile_ast_expr
           | E.Unbound _ -> 
             acc
         ) E.t_true (List.combine idx_vars bounds) in
-        (* For map value equality we only consider m1[k] = m2[k] for k in the maps. 
-           `acc_guard` collects the constraints that k is in the map (if the equality is over maps) *)
-        let acc_guard' = List.fold_left (fun acc e -> 
-          let e = List.fold_left (fun acc arr_i -> 
+        (* For map value equality we only consider m1[k] = m2[k] for k in the maps.
+           `acc_guard` collects the constraints that k is in the map (if the equality is over maps).
+           A guard array may have fewer index positions than the current leaf (e.g. the
+           domain array of a map whose values are sets, which adds an element index), so
+           select each guard with only the prefix of index variables matching its arity. *)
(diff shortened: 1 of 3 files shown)


daniel-larraz:
Closing in favor of PR #1404.

PR #1400, opened 2026-07-14
Recursive ADTs

lorchrob:
TODO: Proper support for refinement types
TODO: Support mutual recursion
TODO: Non-scalar leaf types in ADTs
TODO: Proper support for polymorphism


daniel-larraz (commented):
## Overview

This PR makes self-recursive `datatype` declarations survive the frontend as genuine SMT-LIB datatypes instead of being desugared into tagged records. The approach is a clean split:

- **Non-recursive ADTs** keep the existing `lustreDesugarADTs` record encoding (discriminant enum + per-constructor payload fields).
- **Recursive ADTs** (`is_recursive` flag threaded through `LDAT.adt_info` → `G.adt_map`) skip desugaring; `ADTTerm`/`ADTTester`/`FieldProject` flow all the way to `lustreNodeGen`, which emits constructor applications, `(_ is C)` testers, and `C_i` selectors.
- New `Type.Datatype of string * (string * t list) list`, two new `Symbol.interpreted_symbol` cases (`IsConstructor`, `Selector`), a new `DT` feature in `TermLib`, `declare-datatypes` emission in `SMTLIBSolver`, and explicit `DT` rejection in every solver driver that can't handle it (Bitwuzla, MathSAT, OpenSMT, SMTInterpol, Yices2, yicesNative, IC3IA).
- A well-foundedness check (`NonWellFoundedDatatype`), cycle-guards in `unify_types` / `type_contains_enum` / `type_contains_ref`, and self-edge removal in `lustreAstDependencies`.
- Separately, ADT *monomorphization* (`instantiate_polymorphic_adts`) is added so `Option<Msg>` works.

The architecture is sound and the layering is the right one. I built the branch and all 7 new regression tests pass. My concerns are about **robustness outside the tested paths** — several of the lookups are name-based rather than type-based, and I was able to break the compiler with small variations on the PR's own test files.

Everything below was verified against a local build at `08fc042` (the PR head).

---

## Correctness issues

### 1. `find_recursive_selector` resolves field projections by name alone — breaks unrelated code (critical)

`src/lustre/lustreNodeGen.ml:835` is called unconditionally at `src/lustre/lustreNodeGen.ml:1769`, *before* ordinary record projection, and matches on the user-visible field name across **every** recursive ADT and **every** constructor, returning the first hit in `HStringMap` order. Two confirmed failures:

**(a) A plain record field collides with an ADT field name.** Declaring an unrelated recursive ADT breaks a previously-working record projection:

```
datatype Msg = Atomic (s1: int) | Enc (s2: Msg, s3: Msg);
type R = struct { s1: bool };
node main(r: R) returns (ok: bool);
let ok = r.s1; check ok; tel
```

→ `Error opening input file: Not_found`. Removing the `datatype` line makes it work. Any user file that happens to declare a record field sharing a name with an ADT constructor field now fails to compile.

**(b) Field names reused across constructors pick the wrong selector.** This is not hypothetical — the PR's own `msg_atomic_injective.lus` and `msg_ctor_disjoint.lus` declare `s1` in three constructors; they only pass because they never *project* it:

```
datatype Msg = Atomic (s1: int) | Pair (s1: Msg, s2: Msg);
node main(m1: Msg; m2: Msg) returns (ok: bool);
var f: Msg;
let
  f = match Pair(m1,m2) with
      | Pair(x, y) : x
      | Atomic(n) : choose@<Msg>
      end;
  ok = f = m1; check ok;
tel
```

→ `Error opening input file: LustreExpr.Type_mismatch`. `Atomic` sorts before `Pair`, so `s1` resolves to `Atomic_0 : Msg -> int`.

The root cause is that `src/lustre/lustreDesugarADTs.ml:179-193` and `:428-434` deliberately *strip* the `ctor_` prefix so that nodeGen can re-derive it — throwing away the one piece of information needed to disambiguate. Two options:

- **Preferred:** keep the internal `Ctor_field` name in the `FieldProject` for recursive ADTs too, and have `find_recursive_selector` split on the known constructor rather than search. That makes the lookup exact and removes the record collision entirely.
- Or: compile the scrutinee first, and use `Type.is_datatype` / `Type.name_of_datatype` on its type to select the right ADT before matching the field name — which also gives you a natural `None` for records.

### 2. Multiple recursive datatypes are declared to the solver in the wrong order

`src/transSys.ml:1038` and `:1097` iterate `Type.get_all_datatype_types ()`, i.e. hashcons-table order, with one `declare-datatypes` per type. When one datatype's field references another, the reference can be emitted first:

```
datatype A = LeafA (a1: int) | NodeA (a2: A);
datatype B = LeafB (b1: int) | NodeB (b2: B, b3: A);
node main(x: B) returns (ok: bool);
let ok = (x = LeafB(0)); check ok; tel
```

→ `SMT solver failed: "line 4 column 19: invalid datatype declaration, unknown sort 'A'"`; BMC and IC3QE both exit 1 and the property comes back `unknown` instead of `invalid`. Fix by topologically sorting on field dependencies, or by emitting all datatypes in a single `declare-datatypes` block (which also buys you mutual recursion for free later).

### 3. Recursion under a non-`UserType` head hangs the frontend

`is_directly_recursive` (`src/lustre/lustreDesugarADTs.ml:95`) only matches a bare top-level `UserType`. Anything else is classified non-recursive and handed to the record desugaring, which unfolds forever:

```
datatype T = Leaf (v: int) | Node (kids: T^2);
```

→ hangs until `Lib.TimeoutWall`. Detection should be structural (any occurrence of `type_name` anywhere inside the field type), with an explicit "unsupported" error for the cases the datatype encoding can't express, rather than falling through to the record path.

### 4. `-json` output for datatype streams is not valid JSON

`Model.pp_print_value_term_json` (`src/model.ml:280`) has no datatype case, so it falls through to `pp_print_term` and emits a bare term:

```
"instantValues" : [ [0, Atomic(0)] ]
```

`json.load` on the full `-json` output of `tests/regression/falsifiable/msg_self_enc.lus` fails with `Expecting value`. The enum case one branch above quotes its value — datatypes need the same (or a structured `{"constructor": ..., "args": [...]}` encoding, which would be more useful for nested values). The schema change in `schemas/kind2-output.json` covers `typeInfo` but not the value encoding.

### 5. `-xml` output emits a duplicate attribute

`src/lustre/lustrePath.ml:1641` writes `name="%s"` for the datatype name, but `Stream` already has a required `name`:

```xml
<Stream name="m" type="datatype" name="Msg" class="input">
```

→ expat: `duplicate attribute: line 19, column 39`. Compare the existing conventions `enumName` / `abstrName`. `schemas/kind2-output.xsd` also only gained the `datatype` enumeration value — it needs a matching `datatypeName` (and optionally `constructors`) attribute declaration, mirroring the JSON schema's `datatypeTypeInfo`.

---

## Known gaps that currently fail silently

The PR description lists refinement types and mutual recursion as TODOs. Two of these produce wrong answers rather than errors, which is the more dangerous failure mode.

### 6. Refinement types on recursive-ADT fields are silently dropped

`src/lustre/lustreFlattenRefinementTypes.ml:118` and `:129` return `[]` / `ty` with a bare `(* TODO *)`. Confirmed divergence:

```
type Pos = subtype { x: int | x > 0 };
datatype Msg = Atomic (s1: Pos) | Enc (s2: Msg, s3: Msg);   -- recursive     -> invalid
datatype Msg = Atomic (s1: Pos) | Two (a: int, b: int);     -- non-recursive -> valid (k=1)
node main(m: Msg) returns (ok: bool);
let ok = match m with | Atomic(n) : n > 0 | _ : true end; check ok; tel
```

Here it yields a spurious counterexample, but a dropped refinement on an *output* position drops a proof obligation instead. I'd raise an explicit unsupported-feature error for refinement types reachable from a recursive ADT field until this is implemented.

### 7. Polymorphic recursive ADTs give a misleading parse error

```
datatype L<T> = Nil | Cons (hd: T, tl: L<T>);
```

→ `Parser error: Call to node, contract, or user type 'L' passes an incorrect number of type parameters`.

At `src/lustre/lustreTypeChecker.ml:2582` the new `add_ty_syn ctx' i ty` registers the recursive type in `ctx'`, but the type variables were added to `ctx` (`add_ty_vars_ty ctx i ps`), not `ctx'` — so the self-reference `L<T>` fails its arity check. Relatedly, `instantiate_polymorphic_adts` hardcodes `~is_recursive:false` (`src/lustre/lustreInstantiatePolyNodes.ml:799`), so even if the arity check passed, a polymorphic recursive ADT would be monomorphized into a record. Either support it or reject it with a message that says what's actually unsupported.

### 8. Two different definitions of "recursive field"

`is_recursive_field` in the well-foundedness check (`src/lustre/lustreTypeChecker.ml:3097`) matches `UserType (_, [], id)`; `is_directly_recursive` (`src/lustre/lustreDesugarADTs.ml:95`) matches `UserType (_, _, n)`. A single shared predicate would keep the checker and the desugarer from disagreeing about which types are recursive.

---

## Code quality

- **`assert false` on `Not_found`** — `src/lustre/lustreNodeGen.ml:1186,1200` wraps the whole `Ident` fallback chain in `try ... with Not_found -> assert false`, converting a `Not_found` into a location-less `Assert_failure`. This reads like a leftover debugging aid; either drop it or raise a proper internal error.
- **`simplify.ml` — missing `EQ` case.** `ADT` was added to every `assert false` list, but `EQ` has no `[ADT a; ADT b]` branch analogous to the existing `[Abstr a; Abstr b]` at `src/simplify.ml:2097`. ADT equality therefore falls through to `relation_eq` → `relation` → `assert false` at `src/simplify.ml:1037`. The current tests don't reach it, but mirroring the `Abstr` case would close the hole.
- **Weakened invariants.** `src/lustre/lustreNode.ml:1281` and three sites in `src/terms/term.ml:1808-1876` replace `function [] -> ... | _ -> assert false` with permissive `List.fold_left`. `T.Var` and `T.Const` always have `[]` args, so this removes a real invariant check for no functional gain. Worth reverting unless something concretely needed it.
- **Placeholder-vs-full datatype identity is a footgun.** Self-referential fields use `Type.mk_datatype name []`, a *distinct* hashconsed type from `Datatype (name, ctors)`. `Type.equal_types` distinguishes them; only `check_type` is nominal. Concretely: `TermLib.default_of_type` builds `mk_uf_symbol ctor` with *placeholder* argument types while `lustreNodeGen`'s `ADTTerm` case builds the same-named symbol with *full* argument types — and `UfSymbol.mk_uf_symbol` raises `Invalid_argument` on a signature mismatch. This is only safe today because `default_of_type` always selects a base-case constructor. Consider making equality/hashing nominal for `Datatype`, or representing the self-reference as a distinct constructor so the invariant is visible in the type.
- **`adt_mono_key` produces identifiers containing `<`, `>`, `;`** (`src/lustre/lustreInstantiatePolyNodes.ml:625`). Fine today because those names become records that get flattened away, but fragile if one ever reaches the SMT layer or user-facing output.
- **Pretty-printer inconsistency.** In `src/lustre/lustreExpr.ml:767`, `` `IsConstructor `` prints SMT-LIB syntax `((_ is C) a)` from a *Lustre* pretty-printer; the surface syntax is `C?(a)` (`src/lustre/lustreParser.mly:1159`). The `` `UF `` case being un-`assert false`d is a genuine improvement, though.
- **`assert false` vs `failwith` for unsupported backends.** `src/moxi/moxiInput.ml:379` and `src/smt/yicesDriver.ml:139,157` use `assert false` for `Datatype`, while `yicesNative.ml` uses a descriptive `failwith`. The `failwith` form is preferable — these are reachable user-facing paths, not internal invariants.

### Likely-unintended behavior change

`src/lustre/lustreAstNormalizer.ml:1254-1264` refactors three `Ctx.add_ty` folds into `add_exports_to (add_ivars_to (add_ovars_to info))`. Since `add_ty_to_info` is exactly `{ info with context = Ctx.add_ty info.context id ty }`, this is functionally identical **except that the application order is reversed**: previously exports → ivars → ovars (ovars shadow), now ovars → ivars → exports (exports shadow). Nothing in the diff explains this, and it isn't ADT-related. If it's intentional, please note why; otherwise restore the original order.

### Diff hygiene

Several files carry pure formatting churn that makes the diff harder to review: `src/lustre/lustreInput.ml:209` (adds trailing whitespace only), `src/lustre/lustreParser.mly:1420-1423`, `src/lustre/lustreSyntaxChecks.ml:959-964` (reindent, no semantic change), plus trailing whitespace introduced in `lustreAstNormalizer.ml` and `lustreFlattenRefinementTypes.ml`. Worth dropping. The unexplained reordering of existing test cases in `testLustreFrontend.ml` (moving four blocks to sit after the new one) adds diff noise too.

---

## Test coverage

The 7 new tests are well chosen for the happy path — constructor injectivity, constructor disjointness, projection, a nested pattern (`Enc(Atom(n), k)`), a `Option<Msg>` round-trip mixing recursive and monomorphized ADTs, and two falsifiable cases. All pass locally.

Gaps, in rough priority order (the first four are the repros above):

- Field names reused across constructors **and projected** (issue 1b) — note the existing tests declare this shape but never exercise it.
- A record field name colliding with an ADT field name (issue 1a).
- Two recursive datatypes where one references the other (issue 2).
- `-json` / `-xml` output for a datatype-typed stream (issues 4, 5) — a schema-validation test would have caught both.
- Deeper nesting (`Enc(Enc(a,b), c)`) and `k`-inductive properties over recursive ADTs; everything currently passing is `k=0`/`k=1`.
- Recursive ADTs in contracts (assume/guarantee/ghost vars) — `collect_poly_adt_uses_ci` handles contract items but nothing tests it.
- The `DT`-rejection paths added to six solver drivers are untested; at minimum one `--smtsolver` rejection test would confirm the message fires rather than crashing.

## Performance / security

No performance concerns for existing (non-ADT) inputs — the new work is gated on `Datatype` types existing. The one thing to watch: `find_recursive_selector` runs a nested `StringMap.fold` for *every* `FieldProject` in the program whenever any recursive ADT is declared; fixing issue 1 by keying on the scrutinee's type removes that cost as a side effect. `logic_of_sort` forcing `ALL` whenever `DT` is present (`src/terms/termLib.ml:285`) is conservative but correct. No security-relevant surface.

## Summary

The design is right and the solver-driver gating is thorough. Before merge I'd want issues **1–5** addressed (all are confirmed crashes, hangs, or malformed output reproducible from small edits to the PR's own tests), and **6–7** converted from silent wrong answers into explicit unsupported-feature errors, given they're already acknowledged TODOs.

lorchrob:
I have addressed every component of the PR review comment. The only thing I did not address is the following:

> * **`assert false` on `Not_found`** — `src/lustre/lustreNodeGen.ml:1186,1200` wraps the whole `Ident` fallback chain in `try ... with Not_found -> assert false`, converting a `Not_found` into a location-less `Assert_failure`. This reads like a leftover debugging aid; either drop it or raise a proper internal error.

I did not address this because `assert false` is the typical way we handle paths we expect to be unreachable in the code. If such a path is taken, we can debug from the location of the assert failure and the call stack.

daniel-larraz (changes requested):
SMT-LIB 2.6 requires all constructor declarations to be enclosed in parentheses, including nullary constructors. While Z3 permissively allows omitting parentheses for nullary constructors, cvc5 strictly enforces the standard syntax. Consequently, running Kind 2 with cvc5 on `tests/regression/success/adt_recursive_field_name_reuse.lus` triggers a parse error.

lorchrob:
> SMT-LIB 2.6 requires all constructor declarations to be enclosed in parentheses, including nullary constructors. While Z3 permissively allows omitting parentheses for nullary constructors, cvc5 strictly enforces the standard syntax. Consequently, running Kind 2 with cvc5 on `tests/regression/success/adt_recursive_field_name_reuse.lus` triggers a parse error.

Addressed

daniel-larraz (approved):

PR #1402, opened 2026-07-14
Display named refinement types by name in property expressions

daniel-larraz:
## Problem

A quantifier over a **named refinement type** was displayed using the expanded, flattened refinement instead of the type name. For example, with `type HostId = subtype { i : Nat | i < N };`, a property `forall (h: HostId) ...` printed as:

```
forall (h: subtype { i: int | ((0 <= i) and (i < N)) }) (0 <= h)
```

instead of:

```
forall (h: HostId) (0 <= h)
```

This is **not ADT-specific** — it affects any quantifier whose bound-variable annotation is a named refinement synonym.

## Why the name is lost

The synonym name disappears very early: the type checker's `check_type_well_formed` expands type synonyms in type annotations for well-formedness, and the flattened refinement is **semantically required** downstream (it carries the quantifier's guard — dropping the flattening makes valid properties invalid). So the name cannot be preserved through the pipeline; it can only be restored at the display layer.

## Change

Recognize a refinement type as an instance of a named refinement synonym and render it by name:

- `ref_type_canonical_key` computes a canonical string for a refinement type (bound variable normalized to a fixed placeholder, positions ignored by the pretty-printer).
- `compile_type_decl` records each named refinement synonym's already-flattened definition → its name in a new `cstate.ref_type_names`. The flattened *type declaration* matches the flattened form quantifier annotations take, and normalization leaves `TypeDecl`s untouched, so the forms line up.
- `string_of_expr_as_source` collapses a refinement type back to its synonym name when the canonical keys match; anonymous refinements (no matching synonym) are left expanded.

This is applied to all display paths — user properties, generated `SubType` constraints, and contract assumptions/guarantees. `compile_contract_item` now also routes through `string_of_expr_as_source`, which additionally gives contracts consistent ADT rendering.

These are display-only changes (`prop_expr` / contract `src_expr`); the verification terms are untouched.

## Result

```
forall (h: HostId) (0 <= h)
```

Contract guarantees/assumptions collapse too, e.g.:

```
forall (hostid: HostId) (Preference[hostid] = Yes) => ...
```

## Testing

New regression test `tests/regression/success/quantifier_named_refinement.lus` exercises both the property and contract paths. The full regression suite passes (421 tests).

### DIFF
--- a/src/lustre/lustreDesugarADTs.ml
+++ b/src/lustre/lustreDesugarADTs.ml
@@ -531,12 +531,30 @@ let desugar_adts ctx type_and_const_decls node_contract_decls =
     ) ctx' type_and_const_decls' in
     (type_and_const_decls', node_contract_decls', ctx', adt_map)
 
+(* Canonical string key for a refinement type, used to recognize a refinement
+   type as an instance of a named refinement synonym for display.  The bound
+   variable is renamed to a fixed placeholder so that alpha-equivalent
+   refinements share a key; positions are ignored by the pretty-printer.
+   Returns None for non-refinement types. *)
+let ref_type_canonical_key ty =
+  match ty with
+  | LA.RefinementType (_, (_, id, base), pred) ->
+    let placeholder = LA.Ident (Lib.dummy_pos, HString.mk_hstring "$refvar") in
+    let pred = LH.substitute_naive id placeholder pred in
+    Some (Format.asprintf "%a | %a"
+            LA.pp_print_lustre_type base LA.pp_print_expr pred)
+  | _ -> None
+
 (* Rewrite a desugared expression back toward source-level ADT syntax.
    RecordExprs whose type name is in adt_map are reconstructed as ADTTerm nodes
    so that pp_print_expr renders them as "Ctor(arg1, arg2)" or just "Ctor".
-   This is the partial inverse of desugar_expr for the ADTTerm case. *)
-let rec rewrite_as_adt_terms adt_map expr =
-  let r e = rewrite_as_adt_terms adt_map e in
+   This is the partial inverse of desugar_expr for the ADTTerm case.
+   [ref_type_names] maps the canonical key (see [ref_type_canonical_key]) of a
+   named refinement synonym's flattened definition to its type name, so that a
+   refinement type printed in a quantifier renders as the synonym name (e.g.
+   "HostId") rather than its expansion. *)
+let rec rewrite_as_adt_terms ref_type_names adt_map expr =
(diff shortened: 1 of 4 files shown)


src/lustre/lustreDesugarADTs.ml
```diff
@@ -632,7 +648,11 @@ let rewrite_as_adt_terms adt_map expr =
     | LA.Map (pos, kt, vt) -> LA.Map (pos, rewrite_type kt, rewrite_type vt)
     | LA.Set (pos, t) -> LA.Set (pos, rewrite_type t)
     | LA.RefinementType (pos, (p2, id, t), e) ->
-      LA.RefinementType (pos, (p2, id, rewrite_type t), r e)
+      (* Collapse to the named refinement synonym when the type matches one. *)
+      (match ref_type_canonical_key ty with
+       | Some key when List.mem_assoc key ref_type_names ->
+         LA.UserType (pos, [], List.assoc key ref_type_names)
+       | _ -> LA.RefinementType (pos, (p2, id, rewrite_type t), r e))
```
lorchrob: Two structurally identical but differently-named refinement synonyms collide here: `ref_type_names` is keyed only by shape (`ref_type_canonical_key`), so if two synonyms share a shape, `List.assoc` silently returns whichever one comes first, regardless of which type the quantifier actually used.

Concrete example:
```lustre
type A = subtype { i : int | i > 0 };
type B = subtype { i : int | i > 0 };

node main(a: bool) returns (ok: bool);
let
  ok = a;
  check "b_named" forall (x: B) (x > 0);
tel
```

Suggested fix: only collapse when exactly one distinct synonym name maps to the key; otherwise fall back to the fully expanded name. 
```suggestion
    | LA.RefinementType (pos, (p2, id, t), e) ->
      (* Collapse to the named refinement synonym when the type matches one
         unambiguously. Distinct synonyms can share a canonical key (e.g. two
         separately-declared "subtype { i: int | i > 0 }" aliases): in that
         case we cannot tell which name the quantifier's type came from, so we
         fall back to the expansion rather than guess and risk showing the
         wrong name. *)
      (match ref_type_canonical_key ty with
       | Some key ->
         (match List.filter (fun (k, _) -> String.equal k key) ref_type_names
                |> List.map snd
                |> List.sort_uniq HString.compare
          with
          | [name] -> LA.UserType (pos, [], name)
          | [] | _ :: _ :: _ ->
            LA.RefinementType (pos, (p2, id, rewrite_type t), r e))
       | None -> LA.RefinementType (pos, (p2, id, rewrite_type t), r e))
```

lorchrob (approved):

PR #1403, opened 2026-07-14
Incremental JSON output

Jabe03:
Add the `-ijson` option which prints Kind 2 output in the standard JSON format. This format has a different structure to the usual `-json` format, as it prints a sequence of JSON objects instead (as opposed to a single array of JSON objects which is not parsable until the Kind 2 output is complete).


daniel-larraz (changes requested):
Can you please update the user documentation?
https://kind.cs.uiowa.edu/kind2_user_doc/3_output/2_machine_readable.html

daniel-larraz (approved):

PR #1405, opened 2026-07-16
Add option to limit printed container elements

Jabe03:
Add option `--print_container_limit`, which is set to 5 by default, that controls how many elements of a container is printed. This applies to sets, maps, and arrays. It does not apply to tuples nor records since those are flattened in the pretty print output.

This option has no effect on JSON output.

### DIFF
--- a/src/flags.ml
+++ b/src/flags.ml
@@ -3107,8 +3107,21 @@ module Global = struct
     )
 
   let real_precision () = !real_precision
+  let default_arr_elements_printed = 5
+  let arr_elements_printed = ref default_arr_elements_printed
+  
+  (* Array printing *)
+  let _ = add_spec
+    "--max_array_elements_printed"
+    (Arg.Int (fun n -> arr_elements_printed := n))
+    (fun fmt ->
+      Format.fprintf fmt
+        "Specifices number of elements to print when printing array-like types in pretty-print output.@ \
+        Default: %d"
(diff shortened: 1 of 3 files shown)


src/flags.ml
```diff
@@ -3107,8 +3107,21 @@ module Global = struct
     )
 
   let real_precision () = !real_precision
+  let default_arr_elements_printed = 5
+  let arr_elements_printed = ref default_arr_elements_printed
+  
+  (* Array printing *)
+  let _ = add_spec
+    "--max_array_elements_printed"
```
daniel-larraz:
```suggestion
    "--print_container_limit"
```

src/flags.ml
```diff
@@ -3107,8 +3107,21 @@ module Global = struct
...
+    (Arg.Int (fun n -> arr_elements_printed := n))
+    (fun fmt ->
+      Format.fprintf fmt
+        "Specifices number of elements to print when printing array-like types in pretty-print output. Set to -1 to always print all elements.@ \
```
daniel-larraz:
```suggestion
        "Maximum number of elements to print for each array, map, or set in plain-text output.@ \
        Set to -1 to always print all elements.@ \
```

daniel-larraz (changes requested):
Can you update the wording to refer to containers instead of array-style data structures?

PR #1410, opened 2026-07-22
Fix exception on struct update of aggregate-element arrays

daniel-larraz:
## Problem

A struct update `a[i := v]` on an array whose **elements are aggregates** (records, sets, maps, ...) raises an internal exception during compilation. Minimal reproducer:

```
type R = struct { a: int; b: int };
node N(i: int; v: R) returns ();
var arr: R^3;
let
  arr = (R {a = 0; b = 0}^3)[i := v];
  check true;
tel
```

```
<Error> Error opening input file '...':
        Invalid_argument("numeral_of_term")
```

The same failure occurs for `set<T>^N` and `map<K, V>^N` (e.g. `responses[id := s]` for `responses : set<Request>^N`), and for numeral indices too.

## Cause

There are two issues in `compile_struct_update` ([`src/lustre/lustreNodeGen.ml`](../blob/main/src/lustre/lustreNodeGen.ml)):

1. `Term.numeral_of_term index_term` was evaluated **before** the `Term.is_numeral` guard, so a variable index raised `Invalid_argument("numeral_of_term")`.
2. The real gap: in the trie encoding produced by `compile_ast_type`, the array dimension of `T^N` is the **trailing** index of every leaf's path, with the aggregate structure (record fields, set/map `SetMapIndex`, map-tagged `TupleIndex`, ...) appearing **first**. The update logic assumed the array index was at the **head** of the path, so it hit `assert false` for every aggregate-element array — it only ever worked for scalar-element arrays and top-level tuples/records.

## Fix

The existing (correct) update logic is factored out into `struct_update_core`. For an aggregate-element array we peel off the aggregate prefix, run the core update on the trailing array dimension for each element value, then restore the prefix. `cexpr2` (the replacement element) has the same aggregate structure but without the trailing array index, so its leaf at a given prefix is exactly the new element value to store — letting the proven store/ite encoding be reused unchanged.

Genuine Lustre tuples (`TupleIndex (_, None)`) stay on the core path (direct element replacement); only map/set-tagged `TupleIndex (_, Some _)` entries are treated as aggregate arrays, since a tuple-of-arrays would otherwise be ambiguous.

## Tests

- `tests/regression/success/struct_update_aggregate_array.lus` — array-of-record, array-of-set, and array-of-map updates with both numeral and variable indices (all properties valid).
- Full regression suite passes (424 tests).

### DIFF
--- a/src/lustre/lustreNodeGen.ml
+++ b/src/lustre/lustreNodeGen.ml
@@ -1360,33 +1360,45 @@ and compile_ast_expr
   and compile_struct_update expr1 index expr2 =
     let cexpr1 = compile_ast_expr cstate ctx bounds map expr1 in
     let cexpr2 = compile_ast_expr cstate ctx bounds map expr2 in
-    let rec aux accum = function
-      | [] -> List.rev accum
-      | A.MapIndex _ :: _ -> assert false 
-      | A.SetIndex _ :: _ -> assert false
-      | A.Label (_, index) :: tl ->
-        let index = HString.string_of_hstring index in
-        let accum' = X.RecordIndex index :: accum in
-        if X.mem_prefix (List.rev accum') cexpr1 then
-          aux accum' tl
-        else assert false (* guaranteed by type checker *)
-      | A.Index (_, index_expr) :: tl ->
-        let index_cexpr = compile_ast_expr cstate ctx bounds map index_expr in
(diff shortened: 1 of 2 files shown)


daniel-larraz:
Closing in favor of PR #1412.

PR #1412, opened 2026-07-23
Struct update compilation

lorchrob:
`arr[i := v]`-style struct updates only worked for arrays of scalars, plus a partial fix (PR #1410) for arrays of records/sets/maps. That fix determined whether an update was a tuple-slot replace or an array-row replace by inspecting the compiled trie's shape after the fact, which is provably ambiguous — a tuple of two same-sized arrays and an array of same-shaped tuples compile to identical trie keys, so no amount of shape-based cleverness can tell them apart correctly in general. Stress-testing the fix also turned up two more severe, pre-existing bugs unrelated to nesting depth: any variable-indexed array update silently overwrote the entire array instead of one position (previously masked by an unconditional crash that fired first for every non-numeral index), and every numeral-indexed array update failed at the solver with an "unknown constant" error due to a type mismatch between the update's bound index variable and the one the whole-array-equation compiler builds independently elsewhere.

The fix adds an `access_kind` tag to `StructUpdate`'s `Index` constructor, mirroring the tag `IndexAccess` already carries for reads, so the type checker records the tuple-vs-array decision once (where it already knows the answer) instead of discarding it. `compile_struct_update` in `lustreNodeGen.ml` is rewritten around that tag: record-field and tuple-slot updates are a simple whole-subtree swap; array updates use a new recursive function that peels compile-time-flattened aggregate structure (record fields, tuple/ADT positions, nested arrays, sets, maps) before updating at the array's own dimension, symmetric to the existing read-side compile_array_index'. 

The pre-existing `Flags.Arrays.smt()` (genuine SMT array theory) branch is preserved for parity, though it remains non-functional — it was already broken before this change because it doesn't compose with the scalar, bound-variable-parameterized shape the whole-array-equation compiler expects.

A second limitation: chaining multiple updates to the same array within a single bracket (arr[e1 := e2; e3 := e4]) still fails, because an intermediate update's result is a scalar term parameterized by a bound index variable, which can't be selected into again as if it were a genuine array value — exactly the operation a second chained update needs. Splitting the chain into separate named equations at the Lustre level (arr1 = base[e1 := e2]; arr2 = arr1[e3 := e4];) sidesteps the issue entirely and was verified to work correctly, including with aliased variable indices, but that rewriting is not implemented as an automatic compiler behavior — it's a manual workaround, not a fix. There are also some other related bugs, but they are all pre-existing in `main` and I will prepare a follow-up PR.


daniel-larraz (commented):
Built this branch and `origin/main`, ran the full regression suite on the branch (**454 passed, 0 failed**), and differentially tested ~20 hand-written cases against `main`.

## Summary

The design is right: recording the tuple-vs-array decision in the type checker, where it is actually known, instead of re-deriving it from the compiled trie's shape. And it demonstrably fixes real bugs. Verified independently:

| case | main | this PR |
|---|---|---|
| variable-index update (any array) | `Invalid_argument("numeral_of_term")` crash | ✅ correct |
| `int^3^3` var index, `int^3^3^3` | crash | ✅ correct |
| array of tuple / record / ADT / set / map | crash | ✅ correct |
| numeral index into ctor-built array | crash | ✅ correct |
| `M[1 := v]` on a literal-built matrix | ❌ **wrong answer** | ✅ correct |
| tuple-of-arrays slot update, record field update | ✅ | ✅ (no regression) |

The index-variable numbering in `update_array_element` matches `expand_tuple'` (`lustreNodeGen.ml:500-513`) exactly — var `0` is the first `select`, typed from the *last* key element — and mirrors the read-side `compile_array_index'` (`lustreNodeGen.ml:1647-1690`). That symmetry is what makes it work; worth keeping the two in sync deliberately.

I also checked that all the `assert false` arms are genuinely unreachable: `A.Index` is only ever constructed with `Tuple` or `Array` (5 sites, all in the type checker), and `MapIndex`/`SetIndex` are intercepted at `lustreNodeGen.ml:1837-1838` before the call.

## 1. Regression: the documented nested-update idiom no longer compiles

`doc/usr/source/2_input/2_arrays.rst:150-152` documents `M[1 := M[1][2 := 0]]` as *the* way to update an element of a multidimensional array. On this branch it aborts:

```
Error opening input file 't.lus': LustreExpr.Type_mismatch
```

On `main` it compiles (and computes a wrong answer), so neither is right — but this goes from *wrong* to *won't build*.

More importantly, **the PR description understates the limitation**. It attributes it to chaining within one bracket (`arr[e1 := e2; e3 := e4]`), but the real trigger is broader: *any* array update whose new-value expression is itself an array update. `M[1 := M[1][2 := 0]]` is a single update per bracket and still fails, because `cexpr2` is a scalar term parameterized by index var 0 and `E.mk_select_and_push` is then applied to it at `lustreNodeGen.ml:1563`. Both forms are documented syntax.

```lustre
node CtorBase() returns ();
var M, R: int^3^3;
let
  M = 5^3^3;
  R = M[1 := M[1][2 := 0]];        -- LustreExpr.Type_mismatch on this branch
  check "x" R[1][2] = 0 and R[1][0] = 5 and R[0][0] = 5;
tel
```

Not a blocker given `main` is also broken here. But please either fix it, or (cheaper) reject it in the type checker with a real `Unsupported` message and a position — a bare `LustreExpr.Type_mismatch` with no location is rough for something the manual tells users to write. Same for `A[0 := 1; 2 := 3]` at `2_arrays.rst:148`. And update the docs to match whatever you decide.

## 2. Unsoundness on the rewritten line (pre-existing, but please add a TODO)

`lustreNodeGen.ml:1589` keeps only `expr_init`, so an index whose init and step differ silently uses the init value in both:

```lustre
node UnsoundUpdateIndex(v: int) returns ();
var arr: int^3; n: int;
let
  n = 0 -> 1;
  arr = (7^3)[(0 -> 2) := v];
  -- FALSE in Lustre semantics: at step >= 1 the update targets slot 2,
  -- so arr[0] must still be 7. Kind 2 reports it VALID.
  check "unsound_false_prop_proved_valid" (n = 1) => (arr[0] = v);
tel
```

```
unsound_false_prop_proved_valid: valid (k=1)
```

The counterexample for the dual property makes the mechanism explicit — `arr` is `[3,7,7]` then `[4,7,7]`, i.e. slot 0 is updated at every step. Binding the index to a variable (`i = 0 -> 2; arr = (7^3)[i := v]`) hides it, because the index then compiles to a state variable whose init and step terms coincide; the inline `(0 -> 2)` is load-bearing.

Same on `main` (`compile_array_index` at `lustreNodeGen.ml:1636` has it too), so it is **not** introduced here. But this PR rewrote that exact line, and a *false property proved valid* is the worst failure mode there is. Worth at least a `TODO` at the site now, and ideally a type-checker rejection of non-init-step-equal update indices until it can be handled properly.

## 3. `X.choose` drops bindings where the read side folds

At `lustreNodeGen.ml:1539-1540` the array branch takes `X.choose old_sub` / `X.choose new_elem` and returns `X.singleton key updated_v`. The symmetric read path (`lustreNodeGen.ml:1652-1658`) instead folds over *every* key at that level.

I could not construct a reachable multi-key case — array literals are abstracted into `array_literal_vars` before this point, so `[1,2,3][i := v]` works fine — but the asymmetry is undocumented, and it would silently drop array elements rather than fail loudly if one ever arises. Either add a comment justifying the singleton assumption, or fold like the read side does.

`X.choose new_elem` also raises a bare `Not_found` if `new_elem` is empty, which is reachable through `X.find_prefix [top] new_elem` at `lustreNodeGen.ml:1530` if the two shapes ever diverge.

## 4. Minor

- `lustreNodeGen.ml:1542` — "known to be buggy (from before this git blame) and should be investigated" reads oddly and has trailing whitespace. Prefer `(* TODO(#NNNN): ... *)` with an issue link; "before this git blame" won't mean much to a future reader.
- `lustreAst.mli:208` — reusing the 4-constructor `access_kind` for `Index` makes two of the four states unrepresentable-but-typed, which forces the `A.Map` / `A.Unknown` assert arms. A dedicated 2-constructor type would make `lustreNodeGen.ml:1591-1592` disappear entirely. The doc comment is good mitigation if you keep it as-is; consider mirroring it into `lustreAst.ml`.
- Indentation at `lustreNodeGen.ml:1545-1547`: the `else` branch body sits at the same column as `else`, unlike the `then` branch. Legal, but reads as a dangling block.
- `has_prefix` (`lustreNodeGen.ml:1506-1512`) hand-rolls prefix matching (allocating `[p]` / `[k]` per element) that the trie already offers; `X.subsume` may do this directly.

## Test coverage

`struct_update_nested_array.lus` is well-constructed — each node checks both "updated" and "unchanged", so it does catch the whole-array-overwrite bug. But it only covers the *array* side of the ambiguity this PR is about. Suggested additions, all of which I verified pass on this branch:

- **tuple-of-same-sized-arrays slot update** (`t: [int^3, int^3]`, `t[0 := v]`) — the exact counterpart the PR description names as provably ambiguous under the old scheme. Testing only the array side leaves half the fix unlocked.
- 3-D array (`int^3^3^3`) at a variable index.
- array of ADT, array of `set<int>`, array of `map<int,int>` — these exercise the `AdtTagIndex` / `AdtPayloadIndex` / `SetMapIndex` arms of `update_array_element`, which currently have no coverage at all.
- record-field update whose field is an array (the `Label` → `replace_prefix` path).
- an `error`-class test pinning the nested/chained-update behavior so it cannot silently regress further.

## Verdict

I'd merge after adding the tuple-of-arrays test and turning the nested-update failure into a proper diagnostic (or fixing it). Items 2 and 3 are fine as follow-ups if they go into the follow-up PR you mentioned — but please add the `TODO` for #2 now, since a false property proving valid is easy to lose track of.

lorchrob:
I addressed all PR comments. I will say I would not call the first issue a regression, as crashing is better than being unsound :) but I added a proper error in the type checker nonetheless

daniel-larraz:
> I addressed all PR comments. I will say I would not call the first issue a regression, as crashing is better than being unsound :) but I added a proper error in the type checker nonetheless

I completely agree (Claude's words!). I don't know why Claude discloses the PRs it creates but not the reviews it writes. Thank you!

It looks like one of Claude's suggestions introduced a regression. I'm posting it now.

daniel-larraz (commented):
Re-reviewed `ceb2b50c..3e45e3f3`. Built and tested the new head, plus a build of `origin/main` for baseline.

All five points are addressed, and your pushback on the first one is fair — a crash *is* better than silent unsoundness, "regression" was the wrong word on my part. The added `Unsupported` error is a good outcome either way: I confirmed all four of my repros now give a positioned diagnostic instead of a bare `LustreExpr.Type_mismatch`. `X.find inner_dims new_elem` in the fold is also better than what I suggested — more precise than `X.choose`.

## One of my suggestions was wrong and introduced a regression

`X.subsume` in `replace_prefix` (c1528a47) does not mean what I assumed. `trie.mli:175-182`:

> Return a new trie containing only entries with keys that are **not subsets** of the given key
> `subsume t k` **assumes that all keys in the trie `t`, and the key `k` are sorted and do not contain duplicates.**

Subset, not prefix — and neither precondition holds for struct-update tries (record keys are in field order, not sorted; `int^3^3` gives duplicate `ArrayVarIndex` elements). Concretely, `subsume'` splits on the head and then recurses into the **left** sibling subtries with the *full* key, so it strips the index from anywhere deeper inside siblings that sort before it. Sorry for sending you down this path.

```lustre
type Inner = { f: int; g: int };
type Outer = { a: Inner; f: int };          -- 'a' sorts before the updated 'f'

node SubsumeOverReach(v: int) returns ();
var r: Outer;
let
  r = (Outer { a = Inner { f = 1; g = 2 }; f = 3 })[f := v];
  check "outer_f" r.f = v;
  check "inner_f" r.a.f = 1;
  check "inner_g" r.a.g = 2;
tel
```

```
Internal error at s1.lus:9:3: Type mismatch in equation: record indexes do not match
File "src/lustre/lustreNodeGen.ml", line 612, characters 10-16: Assertion failed
```

`r.a.f` is dropped from the trie. Tuples too — `'('(1, 2), 3)[1 := v]` loses `t[0][1]` (assertion at `lustreNodeGen.ml:551`).

Two things that confirm the diagnosis:
- It is **asymmetric**. Rename the sibling `a` → `z` (sorts *after* `f`) and it passes — exactly the `M.split` left-subtrie recursion.
- It fails **loudly**, never with a wrong answer.

This works on `main` and at `ceb2b50c`, so it is a genuine regression against `main`, not something pre-existing. The suite doesn't cover the shape, so CI stays green — I get **454 passed** on the as-pushed head.

I verified a fix, using only primitives with the semantics we actually want:

```ocaml
let replace_prefix prefix old_trie new_sub =
  let kept =
    X.fold
      (fun k _ acc -> X.remove (prefix @ k) acc)
      (X.find_prefix prefix old_trie)
      old_trie
  in
  X.fold (fun k v acc -> X.add (prefix @ k) v acc) new_sub kept
in
```

With that: the three repros above pass, my 15-case battery passes, and the suite is **454 passed**. Reverting to your original hand-rolled `has_prefix` + `X.filter` works equally well — that version was correct, and I should not have suggested replacing it.

Also worth noting: the new `RecordFieldArrayUpdate` test can't catch this, because `RecArr` is a single-field record and it only has an "updated" check. A two-field record with a colliding nested field name (and an "unchanged" check on the sibling) would pin it.

## Two smaller things

**The new guard is narrower than its message.** `nested_array_update_pos` inspects only the immediate `ue` and `e`, so a nested array update wrapped in anything still reaches the raw exception:

```lustre
arr = (R { f = 0^2 }^3)[i := R { f = a[j := v] }];      -- LustreExpr.Type_mismatch
arr = (7^3^3)[i := (if c then a[j := v] else a)];       -- LustreExpr.Type_mismatch
```

Both behave identically at `ceb2b50c`, so this is pre-existing rather than something the commit broke — just flagging that the guard doesn't cover as much as "or new value is itself an array-element update" suggests. (I checked that the legitimate cases still work: an array update under a *tuple slot* or a *record field* update compiles and verifies fine, so the guard is correctly not firing there.)

**No error-class test** for the new diagnostic — `tests/regression/error/` already has `struct_update_non_updatable_ref_type.lus` as precedent.

## Verdict

Everything else looks good to merge. The `replace_prefix` fix is the one thing I'd block on, since it silently breaks a shape that works on `main` today.

lorchrob:
I addressed the first regression (I should have double-checked `subsume`, my bad). 

For `nested_array_update_pos`, fully generalizing the error message is pretty hairy, so I think it's probably better to just work on the follow-up PR that addresses the bug rather than nail down a nice but temporary "unsupported" message. And further, since this is temporary, I don't think it makes much sense to include a test case (it would just be removed in the next PR).

PR #1426, opened 2026-08-04
Decreases clauses for ADTs

lorchrob:
For a description of the check, see `src/lustre/lustreCheckADTDecreases.mli`

This PR smuggles in three minor bug fixes 
1. It makes the parse rule for `match` arms context-sensitive to allow quantifiers in the context of specifications (previously we were spuriously rejecting models with quantifiers in match arms in contracts)
2. The diff on `check_map_set_type` addresses a possible infinite recursion with the same style as the rest of the recursive functions that handle recursive ADTs (essentially, only expand them once, and remember which ones you've expanded by tracking them with a set `seen`)
3. Andrew requested I substitute "Marmy" for "Marmaduke", but only in one file


src/lustre/lustreCheckADTDecreases.ml
```diff
@@ -0,0 +1,209 @@
...
+  | Some (_, items) ->
+    List.fold_left (fun acc item ->
+      match item with
+      | LA.Decreases (_, e) -> Some e
+      | _ -> acc
+    ) None items
+
+(* Check whether the type of e is a recursive ADT according to adt_map. *)
+let is_adt_decreases ctx adt_map e =
+  match Chk.infer_type_expr ctx None e with
```
lorchrob: TODO: Should not pass `None` here, and in general scrutinize usage of `ctx` as we do not have full type checking context

daniel-larraz:
Reviewed the `adt-decreases` diff against `main` (new `LustreCheckADTDecreases` static termination check, type-checker support for ADT decreases measures, `rec`-without-recursion error, parser change for quantifiers in match arms). Each finding below was verified by running the branch-built `./bin/kind2` on a minimal reproducer.

### 1. Soundness: unguarded field selector accepted as a strictly smaller measure

`src/lustre/lustreCheckADTDecreases.ml:58` — `is_strict_subterm` accepts `f(n.pred)` as strictly decreasing. A selector applied to the wrong constructor is unconstrained in SMT, and an ADT measure disables all SMT termination checks (`termination_checks` becomes `[]`, so `guard_rec_call` leaves the recursive call's contract abstraction unguarded). With that shape I got Kind2 to prove both `f(x) = 42` and `f(x) = 43` valid for the same program.

The `match`-desugared form is safe because of the tester guard; it's the bare selector that needs to be rejected, or checked against a guard.

### 2. Crash on a call argument supplying multiple outputs

`src/lustre/lustreCheckADTDecreases.ml:213` — `substitute_formals` folds formals against syntactic arguments with `List.fold_left2`. A legal call like `f(split(tl))`, where one argument expression supplies two outputs, aborts the frontend with an uncaught `Invalid_argument("List.fold_left2")`.

### 3. Non-recursive ADT measure falls through both checks

`src/lustre/lustreTypeChecker.ml:2512` — the type checker admits *any* ADT measure, but the static check only covers *recursive* ADTs, and `src/lustre/lustreNodeGen.ml:2366` only suppresses the SMT ranking function when all bindings are datatypes. A non-recursive ADT wrapping a datatype (e.g. `datatype Box = B (l: List)`) escapes both: every engine dies with `Sort mismatch ... supplied sort is List`.

### 4. `StructUpdate` key position is not traversed

`src/lustre/lustreCheckADTDecreases.ml:119` — `collect_rec_calls` ignores the `label_or_index` sub-expressions, so a recursive call in key position (`m[f(n,m) := 1]`) is silently unchecked, while `m[1 := f(n,m)]` is correctly rejected.

---

Checked and cleared: the `check_type_well_formed` re-parenthesization is a no-op; the `check_map_set_type` `seen` guard cannot skip a reachable unsupported type; type-synonym and refinement-type aliases of ADT measures resolve correctly; `Result.get_ok` on `syn_expr_equal None` is safe; the `assert false` paths in `check_func_decl`/`check_call` are protected by the `MissingDecreasesClause` check and the new `RecursiveAnnotationWithoutRecursion` error. All seven new `tests/regression/success/adt_decreases_*.lus` pass.

Out of scope, noting it only so it isn't mistaken for a regression here: a recursive call inside a function's own contract crashes at `src/lustre/lustreTransSys.ml:1187`, but that reproduces identically with integer measures on `main`.

daniel-larraz:
Re-reviewed at b6ddc9de. All four points are fixed, and the new tests are genuine: raw selectors are rejected outright, multi-output call arguments align via `arity_of_arg`/`align_formals_to_args` (with irrelevant positions still permitted), non-recursive ADT measures — including the `Box = B (l: List)` wrapper — are now a type error, and `StructUpdate` key positions are traversed via `fold_label_or_index`. The doc rewrite matches the new implementation. All 17 negative tests are rejected and all 10 `tests/regression/success/adt_decreases_*.lus` pass.

Two problems came in with the new design, though.

### 1. Soundness: `safe_env` is keyed by name, so shadowing re-admits a bad call

`src/lustre/lustreCheckADTDecreases.ml:126-131` — a match on an *unsafe* scrutinee that binds a name already in `safe_env` leaves the stale entry in place, so the inner binding inherits "is a strict subterm of the measure":

```lustre
datatype List = Cons (head: int, tail: List) | Nil;

function rec F (n: List; m: List) returns (out: int)
con decreases n; noc
let
  out = match n with
  | Nil : 0
  | Cons (h, tl) :
      (match m with
       | Nil : 0
       | Cons (h2, tl) : 1 + F(tl, m)   -- inner tl comes from m, not from n
       end)
  end;
tel
```

Renaming the inner binder to `tl2` is correctly rejected; keeping the name `tl` is accepted. The recursion never terminates, and Kind2 then proves

```lustre
node main (m: List) returns ()
let
  check (Cons?(m) and Cons?(m.tail)) => F(m.tail, m) = 42;
  check (Cons?(m) and Cons?(m.tail)) => F(m.tail, m) = 43;
  check (Cons?(m) and Cons?(m.tail)) => 0 = 1;
tel
```

all three **valid**. The same program with the inner binder renamed and the recursion made well-founded reports the `0 = 1` property invalid, as expected.

When the scrutinee isn't safe, the arm should *remove* the names its pattern binds from `safe_env` rather than pass `safe_env` through unchanged. Quantifier binders can't reach this (the parser rejects quantifiers in that position), so nested matches look like the only vector.

### 2. Crash: `assert false` for arguments denoting several values that aren't direct calls

`src/lustre/lustreCheckADTDecreases.ml:275` — `arity_of_arg` only recognizes `Call`/`Condact`/`Activate`/`RestartEvery`, so any other group-typed argument counts as 1 and `align_formals_to_args` runs out of args with formals still to assign. Both of these abort the frontend with `Assertion failed`:

```lustre
function rec F (n: Nat; m: Nat; k: Nat) returns (out: int)
con decreases n; noc
let
  out = match n with
  | Zero : 0
  | Succ (p) : 1 + F(p, (p, p))    -- group expression
  end;
tel
```

```lustre
  | Succ (p) : 1 + F(p, (if p = p then pair(p) else pair(p)))   -- ternary over two 2-output calls
```

Both type-check, so they reach this pass. This is the same shape as the `f(split(tl))` crash that was just fixed: either compute the arity from the argument's inferred type, or treat a leftover formal as imprecise (`precise = false`) instead of asserting.

daniel-larraz:
Re-reviewed at 944a3632. Both findings are fixed, and the fixes hold up under probing.

**Shadowing.** `collect_rec_calls` now threads a `shadowed` set and removes rebound names from `safe_env` (`src/lustre/lustreCheckADTDecreases.ml:110-160`), covering match patterns, quantifier binders, `any`/`choose`, and binders inside types. My repro is now rejected, as are the deeper variants I tried: an unsafe rebind followed by a match on the rebound name, and the measure's own name being shadowed. The legitimate cases still work — a sibling arm keeps the outer safe variable, a `VarPat` alias of a safe variable stays safe, and a recursive call under a non-shadowing quantifier is accepted.

**Crash.** `flatten_args` plus the shared `TypeCheckerContext.arity_of_expr` replace the local arity guess, and the `assert false` is now an imprecise alignment that rejects instead of crashing. `F(p, (p, p))` and `F(p, if … then pair(p) else pair(p))` both go through cleanly.

Also noting the work beyond what I reported: recursive calls in a function's own contract are now a clean `RecursiveCallInContract` error (that is the `lustreTransSys.ml:1187` crash I had flagged as pre-existing and out of scope), decreases clauses are rejected wherever they would be silently ignored (on nodes, on non-`rec` functions, in standalone contracts, and when duplicated), ADT measures are restricted to input parameters, and `substitute_naive` gained capture-avoiding alpha-renaming for quantifiers.

Regression checks I ran: the full ounit suite passes; frontend acceptance across every file in `tests/regression` is identical to the base build except the one test intentionally moved to ounit; all 22 negative tests are rejected and all ADT-decreases success tests are clean.

### One new item — pre-existing, not introduced by this PR

`substitute_naive`'s `Match` case has exactly the capture bug that was just fixed one constructor away for `Quantifier`: it skips substitution when the arm binds `var`, but never alpha-renames when the *replacement* expression's free variables are bound by the arm (`src/lustre/lustreAstHelpers.ml:304-310`). Match desugaring substitutes `v := b.v` into arm bodies, so a nested arm binding `b` captures it:

```lustre
datatype Box = B (v: int);
datatype BoxBox = BB (inner: Box);

node N (b: Box; c: BoxBox) returns (y: int)
let
  y = match b with
      | B (v) : (match c with
                 | BB (b) : v      -- v should be the outer b.v
                 end)
      end;
tel

node main (b: Box; c: BoxBox) returns ()
let
  check N(b, c) = b.v;
tel
```

The property is reported invalid, with `b = B(1)`, `c = BB(B(0))` giving `y = 0` — `v` resolved against the inner `b`. This reproduces identically on 53a8d9d1, and `main`'s `substitute_naive` is byte-identical there, so it is not a regression from this PR. Mentioning it because the fix is the same alpha-renaming, in the function being edited here anyway.

lorchrob:
> Re-reviewed at [944a363](https://github.com/kind2-mc/kind2/commit/944a3632a25cbbd6d2bf2e470c8a65e24291c278). Both findings are fixed, and the fixes hold up under probing.
> 
> **Shadowing.** `collect_rec_calls` now threads a `shadowed` set and removes rebound names from `safe_env` (`src/lustre/lustreCheckADTDecreases.ml:110-160`), covering match patterns, quantifier binders, `any`/`choose`, and binders inside types. My repro is now rejected, as are the deeper variants I tried: an unsafe rebind followed by a match on the rebound name, and the measure's own name being shadowed. The legitimate cases still work — a sibling arm keeps the outer safe variable, a `VarPat` alias of a safe variable stays safe, and a recursive call under a non-shadowing quantifier is accepted.
> 
> **Crash.** `flatten_args` plus the shared `TypeCheckerContext.arity_of_expr` replace the local arity guess, and the `assert false` is now an imprecise alignment that rejects instead of crashing. `F(p, (p, p))` and `F(p, if … then pair(p) else pair(p))` both go through cleanly.
> 
> Also noting the work beyond what I reported: recursive calls in a function's own contract are now a clean `RecursiveCallInContract` error (that is the `lustreTransSys.ml:1187` crash I had flagged as pre-existing and out of scope), decreases clauses are rejected wherever they would be silently ignored (on nodes, on non-`rec` functions, in standalone contracts, and when duplicated), ADT measures are restricted to input parameters, and `substitute_naive` gained capture-avoiding alpha-renaming for quantifiers.
> 
> Regression checks I ran: the full ounit suite passes; frontend acceptance across every file in `tests/regression` is identical to the base build except the one test intentionally moved to ounit; all 22 negative tests are rejected and all ADT-decreases success tests are clean.
> 
> ### One new item — pre-existing, not introduced by this PR
> `substitute_naive`'s `Match` case has exactly the capture bug that was just fixed one constructor away for `Quantifier`: it skips substitution when the arm binds `var`, but never alpha-renames when the _replacement_ expression's free variables are bound by the arm (`src/lustre/lustreAstHelpers.ml:304-310`). Match desugaring substitutes `v := b.v` into arm bodies, so a nested arm binding `b` captures it:
> 
> ```
> datatype Box = B (v: int);
> datatype BoxBox = BB (inner: Box);
> 
> node N (b: Box; c: BoxBox) returns (y: int)
> let
>   y = match b with
>       | B (v) : (match c with
>                  | BB (b) : v      -- v should be the outer b.v
>                  end)
>       end;
> tel
> 
> node main (b: Box; c: BoxBox) returns ()
> let
>   check N(b, c) = b.v;
> tel
> ```
> 
> The property is reported invalid, with `b = B(1)`, `c = BB(B(0))` giving `y = 0` — `v` resolved against the inner `b`. This reproduces identically on [53a8d9d](https://github.com/kind2-mc/kind2/commit/53a8d9d1fd1b77f879f609c98f21b839f1f544ba), and `main`'s `substitute_naive` is byte-identical there, so it is not a regression from this PR. Mentioning it because the fix is the same alpha-renaming, in the function being edited here anyway.

The issue raised in this comment is addressed by #1439. #1439 will not be merged directly, but it will be split into separate PRs which will be merged. When a new PR opens that is specific to this issue, I will reference it here.

EDIT: the substitute_naive fix is implemented in PR #1440

PR #1437, opened 2026-08-14
Proof obligations for selectors

lorchrob:


daniel-larraz:
## Review

I built this branch and its merge base (38534885), ran the full `tests/regression` suite against both (no existing test regresses), and probed the new feature with hand-written models. Five findings, all reproduced against the two binaries.

### 1. Selector desugaring is no longer idempotent — crash/regression

[`src/lustre/lustreDesugarADTs.ml:420`](https://github.com/kind2-mc/kind2/blob/708516d88b1c3a9e068e81c9298c2f0ede14cfa3/src/lustre/lustreDesugarADTs.ml#L420)

Keeping the `Selector` marker after renaming the field to the internal payload name makes the `Match` case (which re-desugars arm bodies that already contain the desugared scrutinee) rename twice. On a non-recursive nested ADT, `match o.inner with | A(v) : v | B(w) : w end` produces `o.P_P_inner.A_x` and kind2 aborts with `Not_found`. The same file is accepted on the base commit.

### 2. Dedup never fires for recursive ADTs

[`src/lustre/lustreAstNormalizer.ml:124`](https://github.com/kind2-mc/kind2/blob/708516d88b1c3a9e068e81c9298c2f0ede14cfa3/src/lustre/lustreAstNormalizer.ml#L124)

`SelectorHash.equal` delegates to `AH.syn_expr_equal`, which has no `ADTTester` case and returns `Ok false`. A selector normalized three times (inlinable call under a lazy guard inside a quantifier) yields three identical failing properties `Selector[...]_1/_2/_3`; the non-recursive equivalent yields one.

### 3. Index variables paired with dimensions positionally

[`src/lustre/lustreAstNormalizer.ml:384`](https://github.com/kind2-mc/kind2/blob/708516d88b1c3a9e068e81c9298c2f0ede14cfa3/src/lustre/lustreAstNormalizer.ml#L384)

`index_sizes_of_inductive_vars` matches index variables to dimensions by `HString` creation order rather than using the per-variable size already present in `info.inductive_variables`. The generated boolean local can get swapped dimensions (`bool^2` where `bool^3` is required), and the length-mismatch guard silently degrades the type to plain `bool` for a local that is still array-generalized.

### 4. `Arrow` pushes no guard

[`src/lustre/lustreAstNormalizer.ml:2373`](https://github.com/kind2-mc/kind2/blob/708516d88b1c3a9e068e81c9298c2f0ede14cfa3/src/lustre/lustreAstNormalizer.ml#L2373)

A selector read only at the initial instant (`o = x.val -> 0`) gets its obligation asserted at every step, so a correct model reports a failing `Selector` property.

### 5. A mode's `require` does not discharge the obligation

[`src/lustre/lustreAstNormalizer.ml:2029`](https://github.com/kind2-mc/kind2/blob/708516d88b1c3a9e068e81c9298c2f0ede14cfa3/src/lustre/lustreAstNormalizer.ml#L2029)

The idiomatic `mode ( require Some?(p); ensure o = p.val; )` reports a spurious failure. This case also isn't covered by the docs. Conceptually, the solution could be to use lazy implication when desugaring the contract. The same happens with the provided clause. It may require some thought, since this could impose limitations on what can occur within the mode.

### Verified good

Obligations propagate correctly through contracts, imported nodes, contract imports, condacts, and `--modular`; they coexist with `--ivc`, `--enable MCS`, and `--enable CONTRACTCK`; `assert` discharges as documented; all 30 new regression tests land in the directory their name implies. The `Unresolved` gap for refinement types on node inputs is pre-existing — the base commit crashes identically.

Finding 1 is the one to fix before merge; 4 and 5 are soundness-of-reporting issues that will produce spurious failures on idiomatic code.

lorchrob:
For the above comment.

4. The semantics of arrow are eagerm so the `Selector` property failing is expected behavior. Perhaps we should discuss this as a usability issue with Cesare, but imo this is not something that should be addressed in this PR
5. Whether modes should be interpreted as lazy or eager implications is something we will discuss with Cesare and not address in the PR 

I will address issues 1-3.

daniel-larraz:
> 4. The semantics of arrow are eagerm so the `Selector` property failing is expected behavior. Perhaps we should discuss this as a usability issue with Cesare, but imo this is not something that should be addressed in this PR

I agree.

CHANGES.md
```diff
@@ -32,6 +32,7 @@ Improvements:
 - Updated parser code for compatibility with Menhir >= 20260112.
 
 Breaking changes:
+- A user-written [ADT selector](https://kind.cs.uiowa.edu/kind2_user_doc/2_input/14_algebraic_datatypes.html#selectors) `e.f` now generates a proof obligation that the constructor owning `f` is active, reported as a property named `Selector[L<line>C<column>]`. Models that read a field without establishing its constructor will report a new failing property. The obligation is discharged by the short-circuiting operators (`and then`, `or else`, `==>`), `when ... then ... else` expressions, `when` blocks, and `match` arms; their eager counterparts (`and`, `or`, `=>`, `if` expressions and `if` blocks) do not discharge it.
```
daniel-larraz: This should go at the top of the file under a `Unreleased` section.

daniel-larraz (changes requested):

CHANGES.md
lorchrob: Done in 2f88d5a5 — the entry now lives in a new `# Unreleased` section at the top of the file, and the `# Kind 2 v3.0.0` section is back to being identical to `main`.

daniel-larraz (approved):

Issue #10, opened 2013-09-24
Configure errror


lememta:
In ./configure I get this error:

configure: error: cannot find install-sh, install.sh, or shtool in "." "./.." "./../.."

PR #1438, opened 2026-08-16
Documentation Migration

pburer8:
Migrating the user documentation from Sphinx to Hugo with Hextra


daniel-larraz (commented):
Thanks for taking this on — it's a big migration and the content conversion itself is solid. I checked out the branch, built the site with Hugo 0.165 extended, regenerated the PDF, and diffed against the old Sphinx sources: no content was lost (word counts match page-for-page), all 13 `csv-table`s and the admonitions converted correctly, and the site builds clean with zero warnings and no broken internal page links.

There are a few output-breaking bugs, though, mostly in the asset/PDF plumbing rather than the content.

## Blocking

**1. KaTeX CSS 404s on the deployed site** — [`doc/hugo.yaml:66-68`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/hugo.yaml#L66-L68)

`math.katex.base` is set but `css` is not. In that configuration Hextra takes its "local path only" branch ([`scripts/katex.html:70-73`](https://github.com/imfing/hextra/blob/v0.12.3/layouts/_partials/scripts/katex.html#L70-L73)) and emits a raw, unprocessed link:

```html
<link rel=stylesheet href=/vendor/katex/dist/katex.min.css>
```

Two problems. Hugo never publishes `assets/vendor/katex/` because nothing routes it through the resources pipeline — there is no `public/vendor/katex/` after a build, I confirmed. And the absolute `/vendor/...` path drops the `/kind2_user_doc/` baseURL prefix. FlexSearch works precisely because it *does* go through the pipeline (`href=/kind2_user_doc/vendor/flexsearch/dist/flexsearch.bundle.min.<hash>.js`).

Net effect: math renders unstyled on all 4 math pages. Setting `css: vendor/katex/dist/katex.min.css` fixes both publishing and the prefix, but note the fonts then need to land under `css/fonts/` for the stylesheet's relative `url(fonts/…)` references to resolve.

**2. Every diagram is missing from the PDF** — [`doc/pdf-tools/build_print_html.py`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/pdf-tools/build_print_html.py)

`pdfimages -list print/kind2-user-documentation.pdf` returns nothing at all. The merge keeps each page's own relative `src`s (`src="../3_stopwatch_1.png"`), which were relative to `/docs/advanced-features/<page>/`; with `base_url="print"` they resolve to `doc/3_stopwatch_1.png`, which doesn't exist. All four figures — both stopwatch DAGs, the certification process diagram, and the proof sketch — are silently dropped. Image `src`s need rewriting to absolute paths under `public/` during the merge.

**3. Math is triplicated and garbled in the PDF**

KaTeX output is server-side rendered, so `<main>` contains `.katex-mathml` (MathML *plus* a raw-TeX `<annotation>`) and `.katex-html`. The print CSS doesn't include KaTeX's stylesheet, which is what normally hides the MathML copy — so WeasyPrint prints all three. Actual text from the current PDF:

```
It is equivalent to the formula ∀ i ∈ 0;n0; n0; n. ( i = 0 ⇒ Aiii = 2 ) ⋀ ( i ≠ 0 ⇒ Aiii = Bi−1i-1i − 1 ).
```

Minimum fix: `decompose()` everything matching `.katex-mathml` during the merge. Better, also inline `katex.min.css` so the `.katex-html` spans lay out correctly.

**4. Unconverted cross-references pointing at deleted Sphinx paths**

Four survived pandoc as literal code spans rather than links:

- [`doc/content/home.md:9`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/content/home.md#L9), [`:13`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/content/home.md#L13), [`:24`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/content/home.md#L24) — `` `Kind 2 Input <2_input/1_lustre>` ``, `` `JSON / XML Output <3_output/2_machine_readable>` ``, `` `Techniques <1_techniques/1_techniques>` ``. These flow straight into the generated root `README.md`.
- [`doc/content/docs/inputs-and-outputs/tuples.md:25`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/content/docs/inputs-and-outputs/tuples.md#L25) — `` `set data types <10_sets>` ``, which ships on the live site.

**5. The `PYTHON=` escape hatch doesn't work** — [`doc/Makefile:82-92`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/Makefile#L82-L92)

A second, unconditional `pdf: html venv` rule follows the `ifneq`/`else`/`endif` block and overrides it:

```
Makefile:89: warning: overriding recipe for target 'pdf'
Makefile:82: warning: ignoring old recipe for target 'pdf'
```

So `PYTHON=python3 make pdf` still builds a venv — exactly the situation the error message at lines 65-77 tells users to escape from. Deleting lines 89-90 fixes it.

## Should fix

**6. "Edit this page" 404s everywhere** — [`doc/hugo.yaml:61`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/hugo.yaml#L61) still points at `doc/usr/source`; should be `doc/content`.

**7. The PDF build can't fail** — `build_print_html.py` prints `MISSING` / `NO MAIN` and then `continue`s, exiting 0. Under `set -euo pipefail` in `build_pdf.sh` nothing catches that, so a truncated PDF passes CI silently. Worth tracking a failure count and `sys.exit(1)`.

**8. Mangled license badge** — [`doc/content/header.md:8`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/content/header.md#L8) has the RST `:target:` / `:align:` option lines URL-encoded into the shields.io query string (`?color=green%0A%C2%A0%20:target:%20…LICENSE.rst%0A…:align:%20middle`), and it references `LICENSE.rst`, which this PR deletes.

**9. Root `LICENSE.md` carries Hugo front matter** — `cp content/docs/license.md ../LICENSE.md` copies the `title:` / `weight:` block verbatim into the canonical license file. Worth stripping it in the `update` recipe (and it may affect GitHub's license detection, which the README badge depends on).

**10. `home.md` and `docs/_index.md` are ~350-line near-duplicates** that now have to be hand-synced — and `home.md` is the worse-converted copy, holding all the broken refs from #4. The old setup avoided this: a single `home.rst`, with the `doc_root` sed rewriting `:ref:` targets into absolute URLs for GitHub. Restoring that shape — one source, with `{{< relref >}}` rewritten to `https://kind.cs.uiowa.edu/kind2_user_doc/…` — seems better than maintaining two copies.

**11. Dead and fragile `update` recipe** — [`doc/Makefile:26-49`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/Makefile#L26-L49) still runs the `:ref:` and `.. include::` seds against Markdown (no-ops now), and `${doc_root}` was dropped from this Makefile so it expands to empty. Also `echo '…\n'` only produces a real newline because `/bin/sh` is dash on Linux; on macOS it emits a literal `\n`. `printf` would be safer. Relatedly, the requirements list dropped the old "GNU version of `sed` (`gsed` on OSX)" note even though `sed -E -i` is still GNU-only.

**12. Stale instructions in several places**

- [`doc/content/README.md:12`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/content/README.md#L12) says the theme is "vendored directly (not a Hugo Module)" — it's a submodule as of 0adc4d3.
- [`doc/content/README.md:37-43`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/content/README.md#L37-L43) and the comment at [`doc/hugo.yaml:47-49`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/hugo.yaml#L47-L49) both say KaTeX/FlexSearch are fetched from `cdn.jsdelivr.net` at build time, which `fetch-vendor-assets.sh` exists specifically to avoid.
- [`doc/content/docs/_index.md:319`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/doc/content/docs/_index.md#L319) points readers at `doc/usr/README.rst`.
- `doc/content/home.md` leaves the requirements list in RST syntax (`` `Hugo <https://gohugo.io/installation/>`_ ``), so it renders literally in the root README, and it omits the PDF dependencies entirely.

**13. CI** — [`.github/workflows/kind2-ci.yml:112`](https://github.com/kind2-mc/kind2/blob/d38181f662ec7387fea17754f5e7a00c145ccda1/.github/workflows/kind2-ci.yml#L112)

- `Setup Hugo` lacks the `if: runner.os == 'Linux'` guard that the doc step has, so it runs on macOS and Windows for no reason.
- `hugo-version: 'latest'` is unpinned against a project that has shipped breaking changes; a floor or an exact version would be safer.
- `pip install beautifulsoup4 lxml weasyprint` is dead weight: `make doc` always builds its own venv (see #5), so those packages are never used — while the system libraries WeasyPrint actually needs (documented in `doc/pdf-tools/README.md:26`) are never installed.

## Minor

- `doc/.hugo_build.lock` is committed (renamed from the old `.gitkeep`) *and* listed in `doc/.gitignore` — it's a Hugo runtime lock file, so it should just be dropped.
- Root `.gitignore:43` still ignores `doc/usr/build/**`.
- `doc/pdf-tools/destmap.py` keys the PDF page list on old RST stems (`9_other/5_proofs`), so adding a page means inventing a fake Sphinx path — and forgetting means silent omission, per #7. Deriving the order from Hugo's own section weights would remove the dual source of truth.
- `.gitmodules` uses `branch = v0.12.3`, which is a tag rather than a branch; harmless for `--init`, but `--remote` won't work.
- `make doc` rewrites `README.md` / `LICENSE.md` on every CI run with no drift check. A `git diff --exit-code` step would catch un-regenerated READMEs.
- `svg { display: none }` in the print CSS suppresses every SVG, not only the oversized arrows it was added for.

daniel-larraz (commented):
Re-reviewed at 02ee443 — I rebuilt the site and the PDF and checked each item.

Nice work on the PDF side in particular; the rewrite of `build_print_html.py` is a real improvement over what it replaced.

## Verified fixed

- **#2 (PDF images)** — `pdfimages -list` now reports exactly 4 embedded images: both stopwatch DAGs, the certification process diagram, and the proof sketch. Emitting paths relative to `print/` rather than `file://` URIs was the right call.
- **#3 (PDF math)** — now reads `∀ i ∈ 0; n. ( i = 0 ⇒ Ai = 2 ) ⋀ ( i ≠ 0 ⇒ Ai = Bi − 1 )`, and `pdffonts` confirms `KaTeX_Main`, `KaTeX_Math-Italic`, `KaTeX_AMS` and `KaTeX_Size2` are all embedded. Inlining the CSS with rewritten font URLs handled this cleanly.
- **#10** — dropping `home.md` and generating the README from `docs/_index.md` with `relref` → absolute URL rewriting is exactly the right shape. I confirmed both `README.md` and `LICENSE.md` regenerate byte-for-byte identically to what's committed.
- **#4, #5, #7, #9, #11, #14, #16, #19** all confirmed fixed.

## New problems introduced

**1. The license badge is now wrong** — [`doc/content/header.md:8`](https://github.com/kind2-mc/kind2/blob/02ee443abbc55914fbe4677115184732a4e84885/doc/content/header.md#L8)

The new badge reads `License-BSD_3--Clause`, but Kind 2 is Apache-2.0 — per `kind2.opam:13` (`license: "Apache-2.0"`), `dune-project:5`, the source file headers, and `LICENSE.md` itself. This ends up on the project's front page, so it's worth fixing before anything else.

**2. CI pins a Hugo version the theme cannot run on** — [`.github/workflows/kind2-ci.yml:115`](https://github.com/kind2-mc/kind2/blob/02ee443abbc55914fbe4677115184732a4e84885/.github/workflows/kind2-ci.yml#L115)

`hugo-version: '0.123.0'` is below Hextra v0.12.3's declared minimum: `min_version = "0.146.0"` in `theme.toml`, and `[module.hugoVersion] min = '0.146.0'` in the theme's `hugo.toml`. The theme also uses the Hugo 0.146+ layout structure (`_partials/`, `_markup/`, `_shortcodes/`), and `doc/content/README.md` states "Requires Hugo Extended v0.146.0+". The doc build will fail on this pin. Pinning something ≥ 0.146.0 (I tested against 0.165.0 locally) addresses the original concern without breaking the build.

**3. KaTeX fonts are still not published** — #1 is only half fixed

The stylesheet resolves correctly now (`/kind2_user_doc/css/katex.min.<hash>.css`), but a clean `hugo --minify` produces **zero** `.woff2` files anywhere under `public/`, while the published CSS still contains `url(fonts/…)` — which resolves to `/kind2_user_doc/css/fonts/…` and 404s. KaTeX depends entirely on its own fonts for symbols, sized delimiters and so on, so math still won't render correctly on the deployed site.

The `module.mounts` block added at [`doc/hugo.yaml:70-76`](https://github.com/kind2-mc/kind2/blob/02ee443abbc55914fbe4677115184732a4e84885/doc/hugo.yaml#L70-L76) can't take effect for two reasons: it's nested under `params:`, but `module` is a top-level Hugo config key (so as written it's just an inert params value), and it mounts `static/`, which doesn't exist in `doc/` — `fetch-vendor-assets.sh` writes to `assets/vendor/`.

One approach that works within Hextra's contract: have `fetch-vendor-assets.sh` drop the fonts into `assets/css/fonts/` and list them under `params.math.katex.assets`. The theme calls `.Publish` on those, which publishes them at their asset-relative path — i.e. exactly the `/css/fonts/` location the inlined CSS expects.

**4. Techniques now sorts last, in both the sidebar and the PDF** — [`doc/content/docs/techniques/_index.md`](https://github.com/kind2-mc/kind2/blob/02ee443abbc55914fbe4677115184732a4e84885/doc/content/docs/techniques/_index.md)

This section index has no `weight`, while `inputs-and-outputs` and `advanced-features` have 3 and 4. Hugo sorts the unweighted one last, and the new auto-discovery in `build_print_html.py` mirrors that (falling back to `10**9`). The PDF order is now Inputs and Outputs → Advanced Features → Techniques → License; the old `destmap.py` hardcoded Techniques first, matching the original toctree, so switching to auto-discovery inherited a latent site bug into the PDF. Adding `weight: 2` fixes both at once.

**5. PDF filename mismatch between .gitignore and the Makefiles**

[`.gitignore:43`](https://github.com/kind2-mc/kind2/blob/02ee443abbc55914fbe4677115184732a4e84885/.gitignore#L43) was renamed to `doc/user-documentation.pdf` (hyphen), but the root [`Makefile:33`](https://github.com/kind2-mc/kind2/blob/02ee443abbc55914fbe4677115184732a4e84885/Makefile#L33) still copies to `doc/user_documentation.pdf` (underscore) — so the file `make doc` actually produces is no longer ignored and shows up as untracked. `doc/Makefile` also adds its own `cp print/kind2-user-documentation.pdf user-documentation.pdf`, so `make doc` now writes two copies of the same PDF under different names. Worth settling on one name across `.gitignore`, both Makefiles, and the release workflow's `user_documentation.pdf` reference.

## Smaller leftovers

- `LICENSE.md` now opens with a blank line and `Version 2.0, January 2004`. Stripping the front matter removed the only "Apache License" title — the phrase now survives only inside the boilerplate appendix at line 178. Emitting the front-matter title as an `# ` heading during `update` would restore it.
- The generated `README.md` lost its `# Kind 2` H1 for the same reason (the title now lives only in `docs/_index.md`'s front matter).
- `doc/pdf-tools/destmap.py` is still tracked but no longer imported by anything — safe to delete now.
- `build_pdf.sh` hard-requires `pdfinfo` and `pdfimages`, but poppler-utils isn't listed in `doc/pdf-tools/README.md` alongside the WeasyPrint system libraries, and neither it nor those libraries are installed in CI.
- **#6**: the editURL path is fixed, but the branch is now `master`; this repo's default branch is `main`. GitHub's legacy rename redirect will probably keep the links working, but it should say `main`.
- **#13**: `Setup Hugo` still lacks the `if: runner.os == 'Linux'` guard, so it runs on the macOS and Windows matrix legs.
- `doc/content/README.md:38-43` still advises setting `markup.goldmark.extensions.passthrough.enable: false` for network-restricted builds — that would now break math rendering, and with the vendored assets the advice is moot.
- Minor: `build_print_html.py` still strips `a[href*="edit/main"]`, which no longer matches now that editURL points at `master`. Harmless in practice — I confirmed no edit links appear inside `<main>` — but it's dead as written.

daniel-larraz (commented):
Re-reviewed at 0ff5f2c — rebuilt vendor assets, the site, and the PDF from scratch.

Most of the previous round is confirmed fixed. The remaining problem is that the KaTeX font fix went one step too far: the URL rewrite added to `fetch-vendor-assets.sh` reintroduces the baseURL bug it was meant to solve, and breaks the PDF fonts as a side effect.

## Verified fixed

- **License badge** now reads Apache 2.0, matching `kind2.opam`, `dune-project` and `LICENSE.md`.
- **Section ordering** — with `weight: 2` on the Techniques index, Techniques is now first in both the sidebar and the PDF (I checked the merge order directly).
- **KaTeX fonts are published** — a clean build now emits all 20 `.woff2` files to `public/css/fonts/`. The `params.math.katex.assets` approach works.
- **PDF filename** is consistent across the root `Makefile` and `.gitignore`, and the duplicate `cp` in `doc/Makefile` is gone.
- `destmap.py` deleted, editURL back to `main`, `# Apache License` and `# Kind 2` headings restored (I confirmed both files still regenerate byte-for-byte), poppler-utils documented, the stale passthrough advice removed.

## Still broken

**1. KaTeX font URLs still miss the `/kind2_user_doc/` prefix** — [`doc/fetch-vendor-assets.sh:33-37`](https://github.com/kind2-mc/kind2/blob/0ff5f2c6/doc/fetch-vendor-assets.sh#L33-L37)

The new `sed 's|url(fonts/|url(/css/fonts/|g'` makes the font references absolute from the *server* root. After a clean build:

```
stylesheet link:  /kind2_user_doc/css/katex.min.<hash>.css     <- correct
font URL inside:  /css/fonts/KaTeX_Main-Regular.woff2          <- missing the prefix
actually served:  /kind2_user_doc/css/fonts/KaTeX_Main-Regular.woff2
```

So the browser requests `https://kind.cs.uiowa.edu/css/fonts/…` and gets a 404. This is invisible under `hugo server`, which rewrites baseURL to `/` — it only shows up at the deployed subpath.

The fix is to **drop the sed entirely** and keep KaTeX's original relative `url(fonts/…)`. Hugo publishes the stylesheet at `/kind2_user_doc/css/katex.min.<hash>.css`, so a relative `fonts/…` resolves to `/kind2_user_doc/css/fonts/…` — exactly where the `assets:` list already publishes them. The relative form is prefix-agnostic, so it also keeps working if `baseURL` ever changes.

**2. That same sed removed the KaTeX fonts from the PDF** — [`doc/pdf-tools/build_print_html.py:114-118`](https://github.com/kind2-mc/kind2/blob/0ff5f2c6/doc/pdf-tools/build_print_html.py#L114-L118)

The script inlines the stylesheet and rewrites `url(fonts/` to a path relative to `print/`. That string no longer occurs in the CSS, so the replacement is now a no-op and the inlined CSS keeps pointing at `/css/fonts/…`, which WeasyPrint resolves from the filesystem root and doesn't find. `font_dir` on line 114 also still points at `assets/vendor/katex/dist/fonts`, which the new fetcher no longer creates.

Measured on the PDF I just built: `pdffonts` reports **zero** KaTeX faces, where the previous round embedded `KaTeX_Main`, `KaTeX_Math-Italic`, `KaTeX_AMS` and `KaTeX_Size2`. The text still extracts correctly, but the math is laid out in fallback faces, and KaTeX's metrics assume its own fonts — sized delimiters and radicals are what degrade most visibly.

Fixing #1 restores this automatically. It would also be worth giving `font_dir` the same existence check that `KATEX_CSS` gets a few lines above, so a future path change fails loudly instead of silently dropping the fonts.

**3. CI `apt-get install` lost its `-y`** — [`.github/workflows/kind2-ci.yml:121`](https://github.com/kind2-mc/kind2/blob/0ff5f2c6/.github/workflows/kind2-ci.yml#L121)

```yaml
sudo apt-get install libpango-1.0-0 libpangocairo-1.0-0 libcairo2 libgdk-pixbuf2.0-0 poppler-utils
```

Without `-y`, apt prompts and aborts when stdin isn't a terminal. If every package happens to be preinstalled on the runner image apt exits cleanly and this passes by luck, but it will fail the first time one of them actually needs installing. The step this replaced used `-y`.

## Minor

- `hugo-version: 'latest'` is correct again (the theme needs ≥ 0.146.0), though pinning a concrete version ≥ 0.146.0 would still give reproducible builds; that was the intent behind the original note.
- `Setup Hugo` still lacks `if: runner.os == 'Linux'`, so it runs on the macOS and Windows matrix legs.
- `doc/.gitignore` was broadened from `assets/vendor/` to `assets/`. Nothing under `doc/assets/` is tracked today, so this is harmless now, but it will silently swallow any hand-written site asset added later — `assets/vendor/` plus `assets/css/fonts/` would be tighter. (The file also still has no trailing newline.)
- `distclean` still removes only `assets/vendor`, so `assets/css/fonts/` survives it.
- `LICENSE.md` has a doubled blank line after the new `# Apache License` heading.
- In `doc/content/README.md`, the note says `hugo build` fetches the assets via `fetch-vendor-assets.sh` — it's `make html` / `make vendor-assets` that runs it, not Hugo. The trimmed sentence also lost its closing period.
- `fetch-vendor-assets.sh` fetches only `.woff2`, while `katex.min.css` also references `.woff` and `.ttf` for each face (40 URLs that will always 404). Harmless for any browser or WeasyPrint build in use today, but a one-line comment in the script would save someone a confusing debugging session later.

daniel-larraz (commented):
Re-reviewed at edfa46d — clean rebuild of the vendor assets, the site and the PDF, plus an end-to-end `make doc`.

Dropping the `sed` fixed the site: the stylesheet is published at `/kind2_user_doc/css/katex.min.<hash>.css`, its font references are back to the relative `url(fonts/KaTeX_Main-Regular.woff2)`, and all 20 `.woff2` files land in `public/css/fonts/` — so they resolve correctly at the deployed subpath. I also downloaded Hugo 0.146.0 extended and built with it rather than relying on the theme's declared minimum: it builds cleanly and emits the identical stylesheet link and font URL as 0.165. `make doc` exits 0 and leaves the tree clean.

One item remains — the other half of the previous round's #2.

**KaTeX fonts are still missing from the PDF** — [`doc/pdf-tools/build_print_html.py:114`](https://github.com/kind2-mc/kind2/blob/edfa46d1ab124e03cd5c3ada21f06db22c0f1520/doc/pdf-tools/build_print_html.py#L114)

```python
font_dir = KATEX_CSS.parent / "fonts"     # -> assets/vendor/katex/dist/fonts
```

Now that the CSS is pristine again the `url(fonts/` replacement on line 118 matches, but it rewrites to `url(../assets/vendor/katex/dist/fonts/…)` — and the fetcher no longer creates that directory, since the fonts moved to `assets/css/fonts/`. WeasyPrint finds nothing there, so `pdffonts` on the PDF from a full `make doc` reports zero KaTeX faces.

I verified the fix rather than just guessing at it:

```python
font_dir = ROOT / "assets" / "css" / "fonts"
```

That produces `url(../assets/css/fonts/…)` and embeds five KaTeX faces (`KaTeX_Main`, `KaTeX_Math-Italic`, `KaTeX_AMS`, `KaTeX_Size2`, `KaTeX_SansSerif`).

Worth pairing it with the existence check suggested last round — `KATEX_CSS` gets one a few lines above, and giving `font_dir` the same treatment would have turned this into a build failure instead of a silent font fallback.

daniel-larraz:
That fixed all the issues above. Thank you!

However, since you translated the documentation, other PRs have updated it, which has resulted in merge conflicts. Could you please incorporate the latest changes into the new documentation files?

PR #1440, opened 2026-08-20
Fix match binder capture

lorchrob:
Fix variable capture in `substitute_naive`'s `Match` and `Quantifiers` cases, reported in review of #1426.


daniel-larraz:
## Review

I checked out the branch, built it, ran the new tests both with and without the fix, and ran the full regression suite. The fix looks right; details below, plus one adjacent bug in the same family that this PR doesn't cover.

### Verification

- Builds clean; full regression suite **468/468 passed**.
- The four new tests are genuine regressions. On base (`ab864671`) they fail hard:
  - `adt_match_binder_capture.lus` → `Assertion failed` at `src/lustre/lustreAstNormalizer.ml:109`
  - `adt_match_binder_capture_recursive.lus` → malformed SMT (`unknown constant Atom_0 (Int)`)
  - `quant_binder_capture.lus` → **false "invalid"** on `InvProp[L12C3]` and on the subtype obligation — so the bug was producing wrong verification verdicts, not just crashes. Good catch.
- Two claims in the new comments check out:
  - `rename_pat_var`'s "a `VarPat` is always a genuine binder, never a 0-arg constructor" — confirmed. `bind_pattern_ty` (`src/lustre/lustreTypeChecker.ml:1605-1618`) rewrites a 0-arg-constructor `VarPat` into `Pat (pos, id, [])` and the rewritten arms are what's stored in the AST, so the `Absent` arms in the poly test can never be alpha-renamed.
  - Not renaming quantifier *binder types* is safe: a binder type can't mention another binder of the same quantifier — the syntax check rejects it (`Illegal variable 'x' in type of constant`).
- The `expr_is_id` → `expr_contains_id id` change in `lustreAstHelpers.ml:222-225` is a real latent bug fix (the old code answered "does this type mention *any* identifier"), and it now matters because `expr_contains_id` is load-bearing for the capture decision.

`expr_contains_id` over-approximates in both places it's used here (it ignores shadowing inside `Match`/`Quantifier`), which is the safe direction for capture avoidance.

### 1. Same bug family, still present in `desugar_arm` (pre-existing)

`src/lustre/lustreDesugarADTs.ml:207` folds the pattern substitutions **sequentially**:

```ocaml
List.fold_left (fun b (var, expr) -> LH.substitute_naive var expr b) body subs
```

Each `expr` is `FieldProject (pos, scrut, field)` and so embeds the scrutinee. A binder that shadows a free variable of the scrutinee therefore rewrites inside projections inserted by an *earlier* binder. It's order-dependent — `P (w, a)` is fine, `P (a, w)` is not:

```lustre
datatype Pair = P (x: int, y: int);
node N (w: Pair) returns (r: int)
let r = match w with | P (a, w) : a end; tel
```

`subs = [(a, w.x); (w, w.y)]` → substituting `a` yields `w.x`, then substituting `w` turns that into `w.y.x`. Kind 2 dies with:

```
<Error> Error opening input file 'seq.lus': Not_found
```

I confirmed this reproduces identically on `ab864671`, so the PR neither causes nor worsens it. But it is a match-binder-capture bug in the very function this PR is fixing, so it seems worth either folding in here or filing separately.

The natural fix also resolves point 2 below: teach `apply_subst_in_expr` the same alpha-renaming and use it here, since simultaneous substitution is what this code actually wants.

### 2. `apply_subst_in_expr` still has the un-fixed version of this bug

`lustreAstHelpers.ml:449-456` filters `sigma` by `pat_bound_vars` (so shadowing is handled) but never renames binders that occur in `sigma`'s range — exactly the capture just fixed in `substitute_naive`.

It's unreachable today: `Match` is eliminated at step 10 (`lustreInput.ml:194`) and every `apply_subst_in_expr` caller is in the normalizer at step 22. So no live bug, but the two functions now disagree on a subtle invariant with nothing marking it. A comment on that case would be the minimum.

### Nits

- `.bound_*` names reach user-facing output: `forall (.bound_y_1: int) ((.bound_y_1 >= 0) => (y >= 0))`. `rewrite_as_adt_terms` (`lustreDesugarADTs.ml:806-818`) exists to rename generated bound vars to `$1`, `$2`, …, but `is_generated_bound_var` only matches digit-prefixed names, so `.bound_` names bypass it. If the leading-dot form is the intended display (which the commit message implies), that's fine — otherwise extending `is_generated_bound_var` would make the two generated-name schemes consistent.
- `fresh_bound_ident` guards against duplicating the `.bound_` prefix but not the numeric suffix, so a doubly-renamed binder would become `.bound_x_1_2`. I couldn't construct a case that reaches it; just noting the guard is half of the idea.

daniel-larraz:
## Re-review of `9c1e3d5b`

Rebuilt and re-ran everything. Collapsing `substitute_naive` into `apply_subst_in_expr` is the right call — it removes the duplicated 120-line traversal and makes it impossible for the two to drift apart again, which is how the `StructUpdate` gap below arose in the first place.

### All three points addressed

1. **`desugar_arm` sibling capture** — `lustreDesugarADTs.ml:206` now uses simultaneous substitution. My repro is fixed and both orderings agree:

   ```lustre
   datatype Pair = P (x: int, y: int);
   node N (w: Pair) returns (r: int)
   let r = match w with | P (a, w) : a end; tel   -- was: Not_found crash
   ```
   `check N(w) = w.x` → valid; the `P (w, a) : a` / `w.y` control still valid.

2. **`apply_subst_in_expr`** — now the single implementation, with capture avoidance for `Match`, `Quantifier` and `any`/`choose`. `subst_captures` is sound: I checked every constructor where `expr_contains_id` could disagree with the traversal `apply_subst_in_expr` actually performs. It over-approximates for `Match` (ignores pattern binders) and `Last`, which only causes redundant renames; the `Quantifier` and `AnyOp`/`ChooseOp` cases now line up exactly with what gets rewritten.

3. **`.bound_*` display** — confirmed, `quant_binder_capture.lus` now prints `forall ($1: int) (($1 >= 0) => (y >= 0))`.

### Also verified

- Full suite still **468/468**.
- Nested shadowing is handled correctly. `type T = subtype { v: int | forall (y: int) ((forall (y: int) (y >= 0)) => v >= 0) }` used at `y: T` yields `forall ($1: int) (forall (y: int) (y >= 0) => (y >= 0))` — outer binder renamed, inner shadowing binder correctly left alone.
- Record-typed refinements: `type R = struct { f: Nat }` with `Nat` quantifying over `y`, used at `y: R`, yields `forall ($1: int) (($1 >= 0) => (y.f >= 0))`.
- `substitute_naive var t e = apply_subst_in_expr [(var, t)] e` is behaviour-preserving. The only two differences are both improvements: `apply_subst_in_expr` no longer skips `StructUpdate` index expressions (the old one did — a latent bug for the existing normalizer callers), and `any`/`choose` no longer `assert false`.
- The `!=` → `<>` fix in `expr_contains_id`'s `AnyOp`/`ChooseOp` case is a genuine latent bug fix (physical equality on `HString.t`), independent of the capture work.

### Remaining, all minor

**a. The `any`/`choose` cases are unreachable.** `LGN.gen_nodes` at `lustreInput.ml:178` (step 6) unconditionally rewrites every `AnyOp`/`ChooseOp` into a call to a generated node, and every `apply_subst_in_expr` / `substitute_naive` caller runs at step 10 or later — `lustreDesugarADTs.ml:465` still asserts exactly this. So `subst_under_binder` and its two call sites can't execute, and the `.mli` now advertises support that no input can exercise. Defensible as future-proofing, but the old `assert false` did encode a real invariant that's now undocumented. A short comment saying the cases are defensive would preserve it.

**b. Two comments disagree about binder types.** `lustreAstHelpers.ml:347` says a binder type "can only be a refinement type over constants", so quantifier binder types are left alone; `lustreAstHelpers.ml:438` says an any/choose binder's type "may depend on outer variables" and substitutes into it. They describe the same construct. The first one is right — the syntax check rejects a variable in a type outright (`Illegal variable 'x' in type of constant`) — which makes the `map_lustre_ty` in `subst_under_binder` a no-op. Worth aligning so the next reader doesn't have to work out which is accurate.

**c. No test covers the newly fixed paths.** The suite is unchanged at 468. Nothing exercises the `desugar_arm` sibling-binder fix, the `StructUpdate` index traversal, or the `<>` fix. The first is the one I'd add — a two-line node that crashed with `Not_found` before this commit is cheap insurance, and it's a different bug from the four cases already covered by the new `.lus` files.

**d.** `fresh_bound_ident` still accumulates numeric suffixes on a doubly-renamed binder (`.bound_x_1_2`); with the `$n` display fix this is now invisible in output, so it's cosmetic only.

lorchrob:
I addressed point (c) above (test coverage). 

For various reasons, I don't believe the other comments should be addressed.

I don't think point (a) should be addressed. The `AnyOp`/`ChooseOp` cases may be useful in the future, and I don't think we need to comment every bit of code that is currently unused (as long as it is not too complicated, this case is analogous with other cases that are indeed used)

For point (b), the divergence is intentional and I believe both comments are correct. We have some restrictions on the types of quantified variables that don't apply to any/choose ops so the cases aren't exactly the same.

I don't think point (d) is relevant; this name is never presented to the user so IMO this cannot even be classified as "cosmetic"

PR #1443, opened 2026-08-23
Eliminate quantifiers over finite domains by instantiation

daniel-larraz:
## What

Quantifiers whose bound variables range over a finite domain — the Booleans and enumerated types — are expanded into a conjunction (`forall`) or disjunction (`exists`) over every value of that domain, in `LustreExpr.eval_forall` / `eval_exists`, eliminating the quantifier instead of handing it to the solver.

Enums are encoded as integer ranges and `LustreAstNormalizer` already emits the range guard, so such a quantifier currently reaches the solver as:

```smt2
(forall ((X1 Int)) (=> (and (<= 0 X1) (<= X1 2)) (_select_2 %f22@0 X1)))
```

which becomes `(and (_select_2 %f22@0 0) (_select_2 %f22@0 1) (_select_2 %f22@0 2))`.

This covers three families of quantifier, all of which turn out to be finite-domain:

1. User quantifiers over enums and over `bool`.
2. Map/set equality over an enum domain (the `enum_range_cond` terms from `compile_equality`).
3. **Quantifiers over algebraic datatypes** — an ADT compiles to a tag plus payload fields, and when the tag and all payloads have finite domains the whole quantifier goes away. On the model below, `exists (m: Message)` is four enum-ranged variables and expands to 36 instances.

## Results

Two-phase-commit reachability model, `type HostId = enum {H1, H2, H3}`:

| | off | on |
|---|---|---|
| runtime | 93.9s | **6.9s** |

Scaling the host enum — same k-depths and same verdicts throughout:

| hosts | off | on |
|---|---|---|
| 1 | 1.0s | 0.9s |
| 2 | 18.6s | 2.3s |
| 3 | 102.4s | 7.3s |
| 4 | 759.5s | 77.1s |
| 5 | >2400s (reached 5 of 7 properties) | 840s |

The same model with an *uninterpreted* host type (`type HostId;`) is unaffected — 34.0s vs 34.5s — as expected, since nothing there has a finite domain.

`Summary of Properties` is byte-identical with the flag on and off on every model tested. Side effect worth noting: `<Warning> Cannot mine invariants in quantified terms` disappears on these models, so invariant generation runs where it previously bailed out. That should matter more for valid properties than for the reachability queries measured here.

## Design note

Expansion is **all-or-nothing** — a quantifier is expanded only when that eliminates it outright. My first version expanded whatever variables it could and re-quantified the rest, which made the *uninterpreted* model above go from 36s to 50s: `exists (m: Message)` there has an `Abstr`-typed field, so it produced 12 copies of a body that stayed quantified. Two guards implement this:

- decline if any bound variable lacks a finite domain;
- decline if the body is itself quantified. Multi-variable quantifiers are normalized into nested single-variable ones, so `forall (c: Color; i: int)` arrives as an enum quantifier wrapped around an integer one — the same partial-expansion case. (Without this guard the enum-then-int nesting also drove `Term.destruct` onto a de Bruijn variable.)

Substituting constants leaves the range guards and ADT constructor tests as ground terms (`(=> (and (<= 0 0) (<= 0 2)) …)`, `(ite (= 10 11) X16 3)`), so there is a small local constant folder. `Simplify` cannot be reused: it sits above `LustreExpr` in the dependency graph (`Simplify → Model → LustreExpr`).

## Flags

- `--instantiate_finite_quant` (default `true`)
- `--instantiate_finite_quant_budget` (default `100000` term nodes) — a quantifier whose expansion would produce a larger term is left untouched.

The budget is on the size of the term produced, not the number of instances, so that nested expansion is self-limiting: a multi-variable quantifier is normalized into nested single-variable ones, and each level is charged for the body the level below it already grew. The product of the domain sizes saturates rather than overflowing. See the review thread for the two cases that motivated this.

The default is measured rather than guessed: the largest expansion any regression model asks for is 660 nodes, and the largest the two-phase-commit benchmark asks for is 2592, so 100000 leaves ~40x of headroom.

## Tests

`make test`: **473 passed** (466 existing + 7 new). Each new case verified to give the same exit code with the flag on and off, except where noted below:

- `success/quant_finite_instantiation.lus` — valid `forall`/`exists` over enums, over `bool`, over an ADT with enum payloads, over an ADT mixing an enum and a `bool` payload, and mixed enum/int and bool/int domains that must *not* be expanded.
- `falsifiable/quant_finite_instantiation_forall.lus` — a `forall` false at exactly one enum value, so a dropped instance is caught.
- `falsifiable/quant_finite_instantiation_exists.lus` — an `exists` over an enum with no witness, so an invented one is caught.
- `falsifiable/quant_finite_instantiation_bool_forall.lus` — the same for `bool`.
- `falsifiable/quant_finite_instantiation_bool_exists.lus` — the same for `bool`.
- `success/quant_finite_nested_binders.lus` — 16 stacked enum binders, 3^16 instances if the copies are counted per quantifier instead of budgeted.
- `success/quant_finite_binder_count_overflow.lus` — an ADT with 61 Boolean payload fields, whose 62 binders overflow the product of their domain sizes.

## Pre-existing bug this sidesteps

`check forall (b: bool) b` crashes on `main` today, with the quantifier left intact:

```
<Error> Runtime error in invariant manager: File "src/terms/ltree.ml", line 1061, characters 17-23: Assertion failed
```

Reproduced on an unmodified `v3.0.0-92-g7ff7e75` build, and on this branch with `--instantiate_finite_quant false`. It is unrelated to this PR — expanding the quantifier just means `Ltree.fold` never sees it. `falsifiable/quant_finite_instantiation_bool_forall.lus` therefore passes with the flag on and errors with it off. Worth a separate issue; I have not investigated it.

## Not done

- `CHANGES.md` is untouched — the repo's convention appears to be batching changelog entries at release time (no commits to it since the v3.0.0 version bump), so I left it for whoever cuts the next release.
- The same mechanism would apply to bounded `subrange [i,j] of int`, guarded by the same instance limit. Not implemented. Machine integers are deliberately excluded: `uint8` has 256 values, which sits under the default limit and would expand a body 256 times.
- The `100000` default has headroom measured against the models here, but the adversarial direction is only bounded, not cheap: 16 stacked enum binders still cost 0.8s against 0.2s for leaving the quantifiers alone. That is a 4x regression on an input designed to provoke it, which seemed the right trade for keeping the budget generous enough not to silently stop helping large domains.


daniel-larraz:
Reviewed the branch and independently reproduced both findings below. No soundness bug — the all-or-nothing expansion guards hold up, and the constant folding is semantically fine. Both issues are in the instance-count guard at [`lustreExpr.ml:2271-2276`](https://github.com/kind2-mc/kind2/blob/e849df051bc48dab021c474e0d6be7f233411416/src/lustre/lustreExpr.ml#L2271-L2276), and both let a query blow up that `--instantiate_finite_quant false` answers in 0.2s.

## 1. The limit is per-call, but expansions nest and multiply

[`lustreExpr.ml:2276`](https://github.com/kind2-mc/kind2/blob/e849df051bc48dab021c474e0d6be7f233411416/src/lustre/lustreExpr.ml#L2276) — A source-level multi-binder quantifier is normalized into nested single-variable ones, so each level sees `instances = 3` and passes the 1024 limit trivially, while the body it replicates was already expanded by the level below. The product compounds.

The doc comment says nested binders are declined via `Term.has_quantifier` — but that only holds while the inner quantifier *survives*. Once the innermost one is eliminated, the body is quantifier-free and every enclosing level expands.

Confirmed with 11 `Color` binders (3^11 = 177147 instances):

```
type Color = enum { Red, Green, Blue };
node main (x: Color; y: int) returns ();
let
  check "nested"
    forall (c0: Color; c1: Color; c2: Color; c3: Color; c4: Color; c5: Color;
            c6: Color; c7: Color; c8: Color; c9: Color; c10: Color)
      (y > 0 => (c0 = x or c0 <> x));
tel
```

| | runtime |
|---|---|
| flag on | 15.6s |
| flag off | 0.2s |

The two-phase-commit numbers in the PR description aren't affected by this — those quantifiers are single-binder or ADT-shaped. It's specifically stacked binders over the same finite type.

## 2. `instances` overflows the OCaml `int`

[`lustreExpr.ml:2271`](https://github.com/kind2-mc/kind2/blob/e849df051bc48dab021c474e0d6be7f233411416/src/lustre/lustreExpr.ml#L2271) — `acc * List.length values` is an unchecked 63-bit product. A quantifier over an ADT gives one binder per tag-plus-payload field in a *single* `vars` list, so field counts in the sixties are reachable, and at 62 binders `2^62` wraps to `min_int`; `> limit` is then false and the bail-out is skipped.

```
datatype D = Nil | C (b0: bool, b1: bool, ..., b60: bool);
node main (y: int) returns ();
let
  check "big" forall (d: D) (y > 0 => (C?(d) => (d.b0 or not d.b0)));
tel
```

The boundary is exact:

| bool payload fields | binders | product | result |
|---|---|---|---|
| 60 | 61 | 2^61, fits | declined, exit 0 in 0.2s |
| 61 | 62 | 2^62 → `min_int` | `Fatal error: allocation failure during minor GC` after ~6s under a 4 GB cap |
| 62 | 63 | overflows | same |

## Suggested fix

One change covers both. Make the product saturating rather than wrapping:

```ocaml
let instances =
  List.fold_left
    (fun acc (_, values) ->
       let n = List.length values in
       if acc > limit / n then limit + 1 else acc * n)
    1 finite
```

and budget against the size of the term actually produced — compare `instances * Term.size t` (or node count) to the limit instead of `instances` alone. That makes the nesting case self-limiting: after the inner expansion the body is already large, so the enclosing level declines on its own rather than needing to know it is enclosing anything.

If you take the size-based budget, `1024` stops meaning "instances", so the flag's name and help text would want updating to match. That also gives a more defensible default than the guess noted in the PR description, since a term-size budget is the thing actually worth bounding.

daniel-larraz:
Both confirmed and fixed in a0dd1c93. I reproduced each before changing anything — 15.6s vs 0.2s on the 11-binder case, and the ADT boundary exactly where you put it (60 payload fields fine, 61 and 62 dying under a 4 GB cap).

Your diagnosis of the nesting was right and my doc comment was wrong. The `has_quantifier` guard only declines while the inner quantifier *survives*; once the innermost is eliminated the body is quantifier-free and every enclosing level expands. Counting instances per call cannot see that, because every level makes only three.

I took the size-based budget, since as you say it makes the nesting self-limiting rather than needing a level to know it is enclosing anything:

```ocaml
if instances > budget / max 1 (term_size t) then None
```

with the product saturating instead of wrapping. Written as `instances > budget / size` rather than `instances * size > budget` so the comparison itself cannot overflow — for positive ints the two are equivalent under floor division.

| | before | after | quantifiers left alone |
|---|---|---|---|
| 11 stacked enum binders | 15.6s | 0.80s | 0.2s |
| ADT, 61 bool fields | allocation failure | 0.20s | — |
| ADT, 62 bool fields | allocation failure | 0.21s | — |

Renamed to `--instantiate_finite_quant_budget` with help text saying it bounds term nodes.

**On the default.** You were right that the size budget gives a more defensible number, so I measured rather than guessing again: the largest expansion any regression model asks for is 660 nodes, and the largest the two-phase-commit benchmark asks for is 2592 (the ADT `exists (m: Message)` — 36 instances of a 72-node body). I set 100000, which leaves the benchmark ~40x of headroom.

That choice is worth flagging, because tighter would also have been reasonable. A budget of 20000 holds the nesting case to 0.30s instead of 0.80s, but it would silently stop expanding the two-phase-commit ADT quantifier at roughly 20 hosts, where 100000 carries it to about 115. Since the whole value of the pass is on larger finite domains, I preferred headroom over the adversarial case being cheap. Happy to tighten it if you would rather have the opposite trade.

So the pathological direction is now **bounded, not cheap** — 0.8s against 0.2s for leaving the quantifiers alone is still a 4x regression on an input built to provoke it. I have noted that in the PR description rather than leaving the table above to imply it is free.

Both cases are now regression tests: `success/quant_finite_nested_binders.lus` (16 stacked binders, 3^16 unbudgeted) and `success/quant_finite_binder_count_overflow.lus` (61 Boolean payload fields). `make test`: 473 passed. The two-phase-commit speedup is unaffected — 7.1s vs 96.1s with the flag off, results identical — as are the uninterpreted-type model and the regression suite.

PR #1444, opened 2026-08-23
Open quantifiers when folding a term instead of skipping their variables

daniel-larraz:
## The bug

`Ltree.fold` passes over the variables of a quantifier without folding a value for them, unlike every other leaf it meets. This crashes on:

```lustre
node main () returns ();
let check forall (b: bool) b; tel
```

```
<Error> Runtime error in invariant manager:
        File "src/terms/ltree.ml", line 1061, characters 17-23: Assertion failed
```

Reproduced on an unmodified `v3.0.0-92-g7ff7e75`.

## Why

The value of a term is built from what the folded function returns at each of its leaves, and that function receives a leaf as a `flat` — `Var | Const | App`, which has no representation for a bound variable. So there was nothing to hand it, and the variable was skipped:

```ocaml
| Skip -> fold fail_quant f subst accum tl        (* accumulator untouched *)
```

There is a standing `TODO: implement fold over quantified terms` above the function.

Skipping leaves the result stack one element short, which has two consequences:

- Where something else had been accumulated, the enclosing application is silently rebuilt **without that argument**. `select_terms` on `forall i. a[i]` returned a select of an array with no index.
- Where nothing had, the fold ends with an empty result stack and fails its own assertion. That needs the quantifier body to be a bare bound variable, and bodies are boolean, so it takes exactly `forall (b: bool) b` or `exists (b: bool) b` — which is why this has gone unnoticed.

It is not one crash site. Patching `state_vars_of_term` alone just moves it to `LustreNode.stateful_vars_of_node`; there are 9 callers passing `fail_on_quantifiers:false`.

## The fix

Open the quantifier instead, the way the `Let` case beside it already assigns the terms it binds: give each bound variable a free variable standing for it, so the body folds as if it had no binder over it.

```ocaml
List.map2
  (fun dbi srt -> (dbi, S (db, ht_free_var (T.mk_binder_var srt dbi))))
  (int_seq (db + 1) (List.length n))
  n
```

`BaseTypes` has required `mk_fresh_var` for this since it was written, but nothing in `ltree.ml` ever called it. `Skip` is now unreachable and removed, so the type guarantees every binder installs a substitution.

This is the locally nameless treatment of binders (Aydemir et al., *Engineering Formal Metatheory*, POPL'08; Charguéraud, JAR 2012). The alternative — adding a bound-variable case to `flat` — was tried first and needs a new case at 15 match sites, 6 of which are `destruct`-based and cannot reach it.

Two details the straightforward version gets wrong:

**The variable must not be freshly minted each time.** `vars_of_term` would stop being a function of its argument, and `SMTSolver.get_interpolant` does `Var.VarSet.diff (vars_of_term t2) (vars_of_term t1)`, comparing the results of two folds. It would also grow the variable table without bound on a hot path. So `Var.mk_binder_var` memoizes per sort and binding depth. Two binders at the same depth over the same sort share a variable, which is sound because their bodies never mention both. It comes from `mk_fresh_var` the first time, so it is a name no term already uses.

The memo cannot live in `Ltree`: `BaseTypes` exposes only `hash_of_sort`, which is `Hashcons.hash` returning `hkey` rather than the tag, so it collides and could yield a wrong-sorted variable. Opening a binder is a property of the term representation, so `mk_binder_var : sort -> int -> var` joins `BaseTypes`.

**`vars_of_term` must not report the opened variables.** Collectors selecting state variables are unaffected, since a variable standing for a bound one is free. `vars_of_term` returns variables unfiltered, so it drops them explicitly — a variable a quantifier binds is not a variable of the term it occurs in. This keeps the change invisible to existing callers, notably `LustreExpr.is_const_expr` and the dependency graphs.

## Behaviour on models that do not exercise this

Unchanged, and not incidentally. `mk_binder_var` has a single call site, in the quantifier case of `fold`, and `eval_t` defaults to `fail_on_quantifiers:true`, which raises before reaching it. A model with no quantifiers never creates a binder variable, so the table stays empty and the `vars_of_term` guard cannot hold. Such a model also does not pay for the guard: it short-circuits on the table's O(1) `length` and hashes nothing.

## Verification

**All 467 regression models** run through baseline and patched binaries, with **baseline against itself as a control** — necessary because Kind2 races engines in parallel, so the winning engine, the counterexample, and `k` all vary between any two runs of the same binary:

| | control (baseline vs itself) | patched vs baseline |
|---|---|---|
| identical | 392 | **422** |
| differing verdicts | 75 | **45** |
| differing exit codes | **0** | **2** |

The patched binary differs from baseline *less* than baseline differs from itself. Exit codes are deterministic (0 differences in the control) and the 2 are exactly the new tests below. Stripping `k` from the verdict diffs leaves only those, plus artifacts of truncating each diff hunk.

This method only detects changes larger than that 75-file noise floor; the guarantee rests on the structural argument above, which the experiment corroborates.

**Tests** — `make test`: 468 passed (466 + 2), plus 5 new OUnit cases:

- `tests/ounit/terms/testTermQuant.ml` — body is a bare bound variable; bound variables are not free; **folding is deterministic**; **application arity is kept** (the select regains its index); `fail_on_quantifiers` still raises.
- `regression/falsifiable/quant_bare_bound_var_forall.lus`, `regression/success/quant_bare_bound_var_exists.lus` — the two crashing shapes, now invalid and valid respectively.

**Performance** — a wash. Three runs each on a quantifier-heavy model: medians 18.62s baseline, 18.42s patched. An initial +11% reading on one benchmark did not survive repetition; that file has a ~10% run-to-run spread.

## Relationship to #1443

Independent — this branches from `main` and touches only `src/terms/`. #1443 happens to mask this crash by expanding `forall (b: bool) b` away before the fold sees it, but only for finite domains; any quantifier over an infinite or uninterpreted domain still reaches this code. Either can merge first.

### DIFF
--- a/src/terms/ltree.ml
+++ b/src/terms/ltree.ml
@@ -50,6 +50,8 @@ sig
 
   val mk_fresh_var : sort -> var
 
+  val mk_binder_var : sort -> int -> var
+
   val import_symbol : symbol -> symbol
 
   val import_var : var -> var
@@ -989,9 +991,6 @@ struct
     (* Cached evaluation *)
     | E of 'a
 
-    (* Skip evaluation *)
-    | Skip
-
   (* ********************************************************************* *)
   (* Folding function keeping the term                                     *)
   (* ********************************************************************* *)
@@ -1154,10 +1153,7 @@ struct
 
             )
 
-          | Skip -> fold fail_quant f subst accum tl
-            
           | exception Not_found -> assert false
-                                     
 
       )
 
@@ -1199,11 +1195,27 @@ struct
                  }) :: tl ->
 
       if fail_quant then raise (Invalid_argument "Ltree.fold : quantified term");
-      
-      let vs = int_seq (db + 1) (List.length n) in
 
-      let s = List.map (fun dbi -> (dbi, Skip)) vs in
-      
+      (* Open the quantifier: assign to each of its variables the variable
+         standing for it, exactly as the let binding above assigns the terms it
+         binds. That variable is free, so the body can be folded as if it had
+         no binder over it.
+
+         Folding the body with its variables left bound is not an option. The
+         value of a term is built from what [f] returns at each of its leaves,
+         and [f] receives a leaf as a [flat], which has no representation for a
+         bound variable. Passing over such a variable without folding a value
+         for it leaves the result stack one element short, which silently
+         rebuilds the enclosing application without that argument, and fails
+         the assertion at the top of this function when nothing else is
+         accumulated. *)
+      let s =
+        List.map2
+          (fun dbi srt -> (dbi, S (db, ht_free_var (T.mk_binder_var srt dbi))))
+          (int_seq (db + 1) (List.length n))
+          n
+      in
+
       (* Add to the context stack *)
       let subst' = s @ subst in
 
--- a/src/terms/ltree.mli
+++ b/src/terms/ltree.mli
@@ -98,6 +98,11 @@ sig
 
   val mk_fresh_var : sort -> var
 
+  (** [mk_binder_var s d] is the variable of sort [s] used to open a binder at
+      binding depth [d]. Must be deterministic: the same [s] and [d] always
+      give the same variable, and it must be one no term uses free. *)
+  val mk_binder_var : sort -> int -> var
+
   val import_symbol : symbol -> symbol
 
   val import_var : var -> var
--- a/src/terms/term.ml
+++ b/src/terms/term.ml
@@ -77,6 +77,8 @@ struct
   let sort_of_var = Var.type_of_var
 
   let mk_fresh_var = Var.mk_fresh_var
+
+  let mk_binder_var = Var.mk_binder_var
   
   let import_symbol = Symbol.import 
 
@@ -1833,9 +1835,14 @@ let state_vars_of_term term  =
 (* Return all variables in term *)
 let vars_of_term term = 
 
-  (* Collect all variables in a set *)
+  (* Collect all variables in a set.
+
+     Folding a quantified subterm opens it, so the variables it binds are met
+     as free variables. They are not variables of [term], so drop them. *)
   eval_t ~fail_on_quantifiers:false 
     (function 
+      | T.Var v when Var.is_binder_var v ->
+        (function [] -> Var.VarSet.empty | _ -> assert false)
       | T.Var v -> 
         (function [] -> Var.VarSet.singleton v | _ -> assert false)
       | T.Const _ -> 
--- a/src/terms/var.ml
+++ b/src/terms/var.ml
@@ -456,8 +456,14 @@ let mk_binder_var var_type depth =
 
 (* Return true if the variable was made by [mk_binder_var], and so stands for
    the variable of an opened binder rather than for a variable of the term
-   that binder occurs in. *)
-let is_binder_var v = VarHashtbl.mem (is_binder_vars ()) v
+   that binder occurs in.
+
+   Asked of every variable of every term folded to its variables, so answer
+   the common case -- a term with no binder in it anywhere, which has left
+   this table empty -- without hashing the variable. *)
+let is_binder_var v =
+  VarHashtbl.length (is_binder_vars ()) > 0
+  && VarHashtbl.mem (is_binder_vars ()) v
 
 
 (* ********************************************************************* *)
--- a/src/terms/var.ml
+++ b/src/terms/var.ml
@@ -404,6 +404,62 @@ let mk_fresh_var var_type =
   mk_free_var v var_type 
 
 
+(* Variables standing for the variables of a binder, keyed by the binding
+   depth of the binder.
+
+   Private to each domain, copied from the parent at spawn, for the same
+   reason as the counter above. *)
+let binder_vars_key =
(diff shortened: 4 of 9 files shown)


daniel-larraz:
### Review finding: `binder_vars_key`'s domain split is shallow

The core fix looks sound — I built the branch, ran the new ounit suite (5/5 pass), and ran both new `.lus` regression tests against the built binary (`quant_bare_bound_var_forall.lus` reports invalid, `quant_bare_bound_var_exists.lus` reports valid, as labelled). The de Bruijn level ↔ sort-list pairing in the new `List.map2` matches `mk_lambda`'s `List.rev`-based binding order.

One problem, in the new domain-local state:

https://github.com/kind2-mc/kind2/blob/442861112fc4195ddf3169438726d4604ae17ac5/src/terms/var.ml#L412-L414

`Type.TypeHashtbl.copy` is a *shallow* copy: it copies the outer `type -> (depth -> var)` table, but its values are the inner `Hashtbl`s, which stay shared by reference across domains — while `is_binder_vars` is genuinely per-domain. The pattern was copied from `fresh_var_ids` just above, where it is safe only because those values are immutable ints.

Two consequences:

**1. Binder variables leak out of `vars_of_term` in other domains.** A probe I built and ran against this branch: the supervisor folds a depth-1 quantifier, engine A folds a depth-2 one (adding to the *shared* inner table, but registering the variable only in A's `is_binder_vars`), and engine B then gets

```
vars_of_term = [probe.a@0; __X_Int_5]
```

i.e. the synthetic binder variable escapes as a free variable of the term, which is exactly what `is_binder_var` is meant to prevent.

Worse, the leaked variable carries a tag from A's hash-cons table, so in B the same probe has `Var.compare_vars` return `0` for `__X_Int_5` and an unrelated local variable `bb4` — `Var.VarSet` / `VarMap` / `VarHashtbl` in B then conflate two distinct variables.

**2. Unsynchronized concurrent access.** `Hashtbl.find` / `Hashtbl.add` on that shared inner table:

https://github.com/kind2-mc/kind2/blob/442861112fc4195ddf3169438726d4604ae17ac5/src/terms/var.ml#L442-L455

run concurrently from engine domains, and OCaml 5's stdlib `Hashtbl` is not domain-safe — a lost binding, a torn bucket array on resize, or a hang.

Deep-copying the inner tables at split fixes both:

```ocaml
let binder_vars_key =
  Domain.DLS.new_key
    ~split_from_parent:(fun t ->
      let t' = Type.TypeHashtbl.copy t in
      Type.TypeHashtbl.iter
        (fun ty by_depth -> Type.TypeHashtbl.replace t' ty (Hashtbl.copy by_depth))
        t;
      t')
    (fun () -> Type.TypeHashtbl.create 7)
```

Nothing else in the diff looked wrong to me. The unguarded `List.map2` in the new quantifier case (unlike the `Let` case, which wraps it in `try ... with Invalid_argument _ -> assert false`) cannot raise, since `int_seq s k` returns exactly `k` elements. Sibling binders at the same depth and sort sharing one variable is safe for every current `eval_t ~fail_on_quantifiers:false` consumer I checked (`vars_of_term`, `state_vars_of_term`, `logic_of_flat`, `is_atom`, `stateful_vars_of_expr`), and nesting always strictly increases the depth key, so two simultaneously live scopes cannot collide.

daniel-larraz:
Confirmed and fixed in 96529354. You were right on both counts, and thank you for the probe — I rebuilt it before changing anything and it reproduced immediately:

```
A made:            __X_Int_2
B got:             __X_Int_2
B recognises it:   false   <-- must be true
```

B is handed the variable A made, through the inner table they still share, and does not know it for a binder variable — so it would report it among the variables of the term. Exactly as you described.

**I took a different fix from the one you suggested.** Deep-copying at the split is correct, but it leaves the invariant resting on a `split_from_parent` that a later edit has to remember to keep deep. Holding the depths of a type in a `Map` instead makes the shallow copy correct by construction:

```ocaml
module DepthMap = Map.Make (Int)
```

A map is immutable, so `Type.TypeHashtbl.copy` is a copy in full and each domain gets its own. That closes the unsynchronized-access half too, since nothing is shared to race on. The set of binder variables beside it is already safe this way and I have said so in a comment, since the reason is not obvious: its values are `unit`.

If you would rather have the deep copy for symmetry with how `fresh_var_ids` is written, say so and I will switch — but I think it is better to have the two tables differ visibly and explain why than to have them look alike and be safe for different reasons.

**Regression test.** `binder variables are per domain` in the ounit suite: the parent folds a quantifier at depth 1, one domain folds a nested one at depth 2, and a second domain then folds at that same depth and checks both that the variable is known for a binder variable and that `vars_of_term` still comes back empty. I verified it fails on the unfixed code and passes on the fixed one, rather than only that it passes now:

```
before:  FAILED: Cases: 6 Tried: 6 Errors: 0 Failures: 1
after:   Ran: 6 tests ... OK
```

`make test`: 468 passed, ounit 6/6. The crash probes are unchanged, and three runs each on a quantifier-heavy model show no cost to the map lookup (18.42/18.42/18.62s baseline, 18.32/18.13/18.22s patched).

One thing your review turned up that I should be straight about: I had reasoned about the `is_binder_vars` table being per-domain and about the sort-and-depth key being deterministic, but I never checked what `split_from_parent` actually does to a table of tables. The `fresh_var_ids` pattern directly above it looked like precedent and I took it as one without checking that it holds ints rather than tables.

PR #1445, opened 2026-08-23
Support arrays in IC3IA

daniel-larraz:
IC3IA turned down any system whose logic mentioned arrays:

```
<Warning> IC3IA disabled for property InvProp[L8C3]: arrays are not supported.
```

The check keyed off the array *sort*, so it also caught systems IC3IA reasons
about perfectly well. A map with enum keys is one of them: it is encoded as an
array, but every access is a select on a literal index, and the property below
is proved at k=1 once the engine is allowed to try.

```
type ABC = enum {A, B, C};

node Top() returns (m: map<ABC,int>);
let
  m = map[A := 1; B := 2; C := 3];
  check A in m;
tel
```

Opening the check up exposes three problems it was hiding.

**Counterexample extraction.** IC3IA used `get_model`, and solvers do not print
models for array-typed variables in a form Kind 2 can read back: they range
over the uninterpreted sort that encodes arrays. The values have to be
recovered one index at a time with `get_var_values`, which is what BMC already
does. Until now IC3IA could not report a counterexample over an array at all.

**Global constraints.** IC3IA asserts them directly, without the
`Term.convert_select` that `TransSys.assert_global_constraints` applies. An
unconverted select over a free variable of array type hits an assertion in the
SMT-LIB driver.

**A latent unsoundness.** With a quantifier over array selects -- an
uninterpreted function over an uninterpreted sort -- MathSAT answers
unsoundly, and IC3IA went on to report a *valid* property as falsifiable. This
was pre-existing: the engine already took the counterexample branch, and the
crash in `get_model` masked it. Quantified formulas over arrays are still
turned down for that reason; quantifiers on their own are unaffected.

What remains is decided per interpolating solver rather than for all of them.
Z3qe and cvc5qe are ruled out up front, since their quantifier elimination
tactics reject the encoding partway through computing an interpolant. The
solvers that reject it when the instance is created -- OpenSMT and Bitwuzla
refuse arrays outright, MathSAT refuses arrays over bool, as sets and maps
have -- now report an unsupported feature instead of dying with a runtime
failure. That path is reached only once interpolation is needed, so a property
IC3IA settles without refining is still proved.

## Tests

A plain regression test does not hold IC3IA to anything here: every engine
runs, so 2-induction answers the quantified-array model before IC3IA reports
its spurious counterexample, and a test that regresses from proved back to
"engine disabled" exits 30, which the harness accepts for any test.

Tests under an `ic3ia` directory therefore pin the engine and the interpolating
solver, and exit 30 is a failure for them: these models settle in milliseconds,
so it means IC3IA turned the system down rather than that it ran out of time.
Tests under `ic3ia/declined` invert that, requiring only that the property is
not reported falsifiable. They skip when MathSAT is absent rather than
reporting a run that never finished.

Each test was checked to fail without the fix it covers: reverting only the
select conversion fails `ic3ia_global_const_map`, reverting only the
counterexample extraction fails `ic3ia_array_counterexample`, and dropping the
guard on quantified arrays fails `ic3ia_quantified_array_sound`.

Over the 113 array/map/set/ADT tests in `regression/success`, IC3IA on its own
now proves 63, up from 49, with no unsound answers and no runtime failures.

## CI

The new tests reach arrays only through an interpolating solver that can
represent them, so with Z3 alone they would skip. MathSAT is installed too,
pinned and cached by version the way the benchmark workflows do it.

Two points worth a look:

- This replaces the three per-platform Z3 steps with one setup covering all
  five, which also pins Z3 on macOS, where it came from Homebrew and was
  whatever version that had.
- The cache lives in the workspace rather than the home directory, so the same
  path can go on PATH on Windows, where it has to be a native path.

Adding MathSAT makes it the interpolating solver Kind 2 detects for *every*
test in CI, in place of Z3qe. The suite was run under that configuration
(z3 + mathsat only, 482 passed) and with Z3 alone (475 passed, 7 skipped), so
a failed MathSAT download degrades rather than breaking the job.

### DIFF
--- a/.github/workflows/kind2-ci.yml
+++ b/.github/workflows/kind2-ci.yml
@@ -9,6 +9,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.head_ref || github.ref }}
   cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}
 
+env:
+  Z3_VERSION: "4.15.3"
+  MATHSAT_VERSION: "5.6.17"
+
 jobs:
   kind2-build:
     strategy:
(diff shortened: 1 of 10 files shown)


daniel-larraz:
## Review

Findings below were each checked against the code at 50d4be3.

### 1. The quantified-array soundness guard is bypassable via `--smt_logic`

https://github.com/kind2-mc/kind2/blob/50d4be3d157d1430971845115ea4f48c6b3e96ca/src/ic3ia/IC3IA.ml#L1025

The guard matches only `` `Inferred l when mem A l && mem Q l ``. With `--smt_logic ALL` (or `none`), `TransSys.get_logic` returns `` `SMTLogic s ``/`` `None `` ([transSys.ml#L1900-L1906](https://github.com/kind2-mc/kind2/blob/50d4be3d157d1430971845115ea4f48c6b3e96ca/src/transSys.ml#L1900-L1906)), so the match falls through to `| _ -> ()`. Running

```
kind2 --smt_logic ALL --enable IC3IA --smt_itp_solver MathSAT ic3ia_quantified_array_sound.lus
```

on this PR's own model reaches the unsound MathSAT path.

The `` `Inferred ``-only shape is pre-existing — the old array check had it too — but the consequence is what changes here: before this PR that path ended in a `get_model` crash; now it can report a valid property falsifiable. `check_arrays_are_supported` ([L1078](https://github.com/kind2-mc/kind2/blob/50d4be3d157d1430971845115ea4f48c6b3e96ca/src/ic3ia/IC3IA.ml#L1078)) has the same hole for Z3qe/cvc5qe, where it costs only a wrong diagnosis rather than a wrong answer.

This is the one I would fix before merge — it is the only finding that can produce an incorrect result.

### 2. `mentions "only supports BV logics"` catches the wrong Bitwuzla failure

https://github.com/kind2-mc/kind2/blob/50d4be3d157d1430971845115ea4f48c6b3e96ca/src/ic3ia/IC3IA.ml#L360

[`BitwuzlaDriver.string_of_logic`](https://github.com/kind2-mc/kind2/blob/50d4be3d157d1430971845115ea4f48c6b3e96ca/src/smt/BitwuzlaDriver.ml#L59-L70) raises that string for `DT`, `IA`, or `RA` (and for an `--smt_logic` string containing `I` or `R`) — nothing to do with arrays. So `--smt_itp_solver Bitwuzla` on a plain *integer* model now warns "the interpolating solver does not support the arrays this system uses" at the first refinement, and demotes `ExitCodes.error` to a warning.

Bitwuzla's failure is also raised in `header_logic`, i.e. *after* `Unix.create_process` — unlike OpenSMT, whose check sits in `Driver.cmd_line` — so swallowing it strands the spawned process and its pipes outside `SMTSolver`'s registry. Matching only on the array-specific strings, and letting the BV-logic failure through, addresses both.

### 3. `_is_ic3ia` matches against the absolute path

https://github.com/kind2-mc/kind2/blob/50d4be3d157d1430971845115ea4f48c6b3e96ca/tests/conftest.py#L275-L276

`file_path` from `pytest_collect_file` is absolute, so `self.path.parts` includes the whole checkout prefix. A checkout or worktree under a directory named `ic3ia` pins the *entire* regression suite to `--enable IC3IA --smt_itp_solver MathSAT`, or skips all of it when MathSAT is absent. Matching against `file_path.relative_to(regression_dir)` avoids it — `relative_path` is already computed at [L128](https://github.com/kind2-mc/kind2/blob/50d4be3d157d1430971845115ea4f48c6b3e96ca/tests/conftest.py#L128).

### 4. `declined` tests accept any non-40 exit

https://github.com/kind2-mc/kind2/blob/50d4be3d157d1430971845115ea4f48c6b3e96ca/tests/conftest.py#L285-L290

A crash, a usage error, or an unsupported-solver exit all pass as "IC3IA declined". Restricting to the codes actually expected (proved / unknown / unsupported) would keep the test honest about what it is pinning.

### Checked and cleared

- **Counterexample bound.** Cubes sit at offsets `-1 .. len-2` (cube `i` bumped by `i-1`, `sys_def_unrolling 0` supplying `INIT@0`), so `vars_of_bounds 0 (len-2)` matches exactly what `Model.path_from_model` reads, and `len >= 2` always holds: a `Frame 0` cube can only arrive through a `next_cube` edge, since `depth = 0` makes `get_bad_cube` unsat given `prop = P ∨ %init`. No empty `(get-value ())`.
- **Global-constraint select conversion** mirrors `TransSys.assert_global_constraints`, and the select UFs it can produce are already declared by `declare_selects` inside `define_and_declare_of_bounds`.
- **Functional congruence.** Dropping the blanket array check does not open that soundness gap: `fn_congruence_groups` is aggregated up from subsystems (lustreTransSys.ml), so the retained `has_fn_congruence_groups sys` covers the whole subtree.
- **Solver instances on the new `UnsupportedFeature` paths** are cleaned up by `SMTSolver.destroy_all ()` in `on_exit_of_process`. Separately worth knowing: `| Terminate -> (); SMTSolver.delete_instance solver` ([IC3IA.ml#L996](https://github.com/kind2-mc/kind2/blob/50d4be3d157d1430971845115ea4f48c6b3e96ca/src/ic3ia/IC3IA.ml#L996)) parses as part of the `Terminate` arm, so that call never runs on normal completion — pre-existing, and harmless for the same reason.
- **CI.** The download/cache block mirrors the working `kind2-benchmarks-all-platforms.yml`, the `env` and `GITHUB_PATH` handling is correct, and "Check the solvers run" fails loudly rather than letting a missing solver turn into a silent skip.

daniel-larraz:
All four addressed in f06376d and 2c49545.

### 1. Guard bypassable via `--smt_logic` — fixed

Confirmed first, on the PR's own declined model:

```
--smt_logic detect  rc=30  unknown
--smt_logic ALL     rc=40  invalid after 1 steps     <- valid property, wrong answer
```

`features` now falls back to recomputing the feature set from the terms
whenever the logic on record is not an inferred one, mirroring the `detect`
branch of `mk_trans_sys` and folding over the subsystem tree. Both guards read
from it, so the Z3qe/cvc5qe diagnosis is fixed by the same change. After:

```
--smt_logic detect | ALL | none | QF_AUFLIA   all rc=30, all "quantified formulas over arrays are not supported"
```

### 2. `only supports BV logics` — removed

You are right that it is not an array failure. `--smt_itp_solver Bitwuzla` on
an integer-only model now reports its own error rather than claiming the
system uses arrays:

```
Runtime failure in property directed reachability (IA): Bitwuzla only supports BV logics
```

The two cases the catch exists for still degrade cleanly (OpenSMT on an array
counterexample, MathSAT on a map over bool). Comment updated to say why the
match is array-specific, including the `header_logic` point.

### 3. Absolute path — fixed

Matching against `self.path.relative_to(regression_dir).parts` now. For a
checkout at `/home/dev/ic3ia`, an ordinary test like
`regression/success/array.lus` went from `is_ic3ia = True` to `False`.

### 4. `declined` accepting any non-40 exit — fixed

Restricted to `(success, incomplete_analysis)`. Checked end to end by making
the declined model exit 3: it fails now, and passes again once restored.

Full suite is green both ways: 482 passed with MathSAT present, 475 passed and
7 skipped with Z3 alone.

Thanks for the checked-and-cleared section too — the note that
`| Terminate -> (); SMTSolver.delete_instance solver` parses into the
`Terminate` arm is worth a separate look, but it is pre-existing and I have
left it alone here.

daniel-larraz:
All four addressed at 2c49545. I rebuilt and checked each one rather than reading the diff alone.

### 1. `--smt_logic` bypass

The bug was real: rebuilt at 50d4be3,

```
kind2 --smt_logic ALL --enable IC3IA --smt_itp_solver MathSAT \
  tests/regression/success/ic3ia/declined/ic3ia_quantified_array_sound.lus
```

reported the valid property `invalid after 1 steps`, exit 40. At 2c49545 the guard fires and it exits 30 with "quantified formulas over arrays are not supported" — under `ALL`, `none`, and `detect` alike.

`features_of_terms` is the right shape. `logic_of_flat`'s `Var v -> logic_of_sort` case picks up array and datatype sorts from the terms, so the DT guard survives the change too — checked on `adt_recursive_cross_reference.lus`, where it fires under all three logic settings. Passing `[]` for `fun_symbols` rather than the subsystem init/trans symbols only over-adds `UF`, which none of these guards test.

### 2. Bitwuzla

`--smt_itp_solver Bitwuzla` on a plain integer model now reports

```
<Error> Runtime failure in property directed reachability (IA): Bitwuzla only supports BV logics
```

instead of claiming the solver cannot handle the system's arrays.

### 3 and 4. conftest

`_regression_parts` is right, and its `ValueError` branch covers the `extra_files` that live outside the tree. `ic3ia_declined_codes` closes the accept-anything hole.

### Suite

`dune build` clean, full regression suite 482 passed, the 8 IC3IA tests among them.

### One thing left, minor

https://github.com/kind2-mc/kind2/blob/2c495453922506a78435eb8dab27a2842bd2dcc3/src/ic3ia/IC3IA.ml#L1123-L1127

The MathSAT "both integers/reals and machine integers" check is now the only guard in `main` still matching `` `Inferred `` on `TransSys.get_logic`, so it keeps the bypass this commit set out to close. `l` is already in scope there, and the nested `let open TermLib in let open TermLib.FeatureSet in` is redundant for the same reason.

It is a capability check rather than a soundness one — bypassing it costs a MathSAT error, not a wrong answer — so it is consistency rather than a blocker. I could not get a model past slicing that keeps both features live, so I have not seen it fire either way.

daniel-larraz:
Fixed at 420932b. It reads from `l` now, which also retires the nested
`let open` the match needed.

It did keep the bypass, and there is a model that shows it — one property that
holds both features live past slicing:

```
node Top(a: uint8) returns (n: int);
let
  n = 0 -> (if pre n < 5 then pre n + 1 else pre n);
  check n <= 5 and a >= a;
tel
```

At 2c49545:

```
--smt_logic detect  rc=30  MathSAT does not support programs with both integers/reals and machine integers
--smt_logic ALL     rc=0   valid (k=1)          <- straight past the check
```

At 420932b it turns the system down under `detect`, `ALL`, and `none` alike.

You are right that it costs a MathSAT error rather than a wrong answer, so
this is consistency. I have not added a test: the check fires under `detect`
either way, so a test would have to pin `--smt_logic`, and that is not
something the harness expresses for a single file.

`dune build` clean, full suite 482 passed.

PR #1446, opened 2026-08-23
Let no solver outlive Kind 2

daniel-larraz:
A Windows CI run [failed](https://github.com/kind2-mc/kind2/actions/runs/32657498757/job/97238457246) with Kind 2 gone and the harness still waiting on its output:

```
Killed after 204s: the run did not stop on its own, although it was given 84s
Kind 2 had already exited (status 30), but its output pipes were still open: something it spawned outlived it
```

A solver of that run was still alive. On Windows it held the standard output of Kind 2 open, because a child there inherits every inheritable handle and not only the ones it is given, so the harness reading that output never saw the end of it and waited until it was killed. The same leak is what the job that [hung for 94 minutes](https://github.com/daniel-larraz/kind2/actions/runs/31332087766) ran into, before there was a harness to stop it.

Kind 2 honoured its own `--timeout` in both cases, and exited 30 on schedule. It is the cleanup that was incomplete.

## The leak

The sweep before the process exits was not final. Engines the supervisor gave up on keep running in the background and die with the process, at a point they do not choose; one that starts a solver after the sweep leaves it behind, with nobody left to kill it. The window is wider than it looks, since a solver process is running before it is registered: it is started by the functor application, and the registration only follows.

Latch the registry closed in `destroy_all_of_process`, under the lock that guards it, so that the sweep and a registration cannot interleave: a solver either makes it into the map, and is killed by the sweep, or finds the latch closed, in which case `create_instance` kills the process it has just started and raises `Exiting`. Kill rather than ask to exit, since nothing is left to compute and an engine that was given up on may still be inside a call.

Kill the solvers in the last-resort exit path too. It exists for when the orderly one does not finish, and went straight to `Unix._exit`, which leaves the children of the process running: exactly the case this is about.

## Validation

A new ounit test creates a solver, sweeps, and checks that creating another raises and that no child process is left running. It was checked both ways:

- with the latch check removed, it fails on the first count;
- with the raise removed as well, it fails on the second, reporting `a solver process was left running`.

So the leak reproduces deterministically, and on Linux: the bug is not Windows-specific, only its consequence is. On Linux an orphaned solver is harmless to whoever reads the output of Kind 2, which is why it never showed up there.

The ounit suite and the 475 regression tests pass, the latter in 36.7s, and the strict profile builds clean.

## Not in scope

This stops a solver outliving Kind 2. It does not stop a solver inheriting the standard output of Kind 2 in the first place, which is worth doing as defence in depth: `Unix.pipe ~cloexec:true` for the solver pipes would also stop each solver inheriting the pipes of the previous one, a file descriptor leak on every platform. Left for a separate change, since clearing inheritance on the standard output of Kind 2 itself interacts with the `Sys.command` calls in the certificate paths, which rely on it.

### DIFF
--- a/src/kind2Flow.ml
+++ b/src/kind2Flow.ml
@@ -426,7 +426,13 @@ let post_clean_exit process base_status exn =
              killing it does not kill them. One left behind holds the
              standard output of Kind 2 open on Windows, where a child
              inherits every inheritable handle, so whoever reads that
-             output keeps waiting for an end that never comes. *)
+             output keeps waiting for an end that never comes.
+
+             This is the path that must not hang, and it now waits for
+             a lock to take it. Every critical section that lock
+             guards is a map operation with no I/O in it, held for as
+             long as an insertion or a swap takes, so waiting for it
+             cannot become the reason we never get to [_exit]. *)
           ( try SMTSolver.destroy_all_of_process () with _ -> () ) ;
           Unix._exit status)
         ()
@@ -434,6 +440,12 @@ let post_clean_exit process base_status exn =
     with _ -> () ) ;
   (* Close tags in JSON/XML output. *)
   KEvent.terminate_log () ;
+  (* The engines that are still running are on their way out with the
+     process, whatever they are in the middle of. Saying so before the
+     sweep is what keeps an engine that was given up on from being
+     reported as having crashed when the solver it asks for is refused,
+     the way one is not reported for the solvers killed under it. *)
+  EngineDomains.set_terminating true ;
   (* Kill all live solvers of the whole process. *)
   SMTSolver.destroy_all_of_process () ;
   (* Exit with status. *)
--- a/src/kind2Flow.ml
+++ b/src/kind2Flow.ml
@@ -422,6 +422,12 @@ let post_clean_exit process base_status exn =
           Thread.delay 10.0 ;
           prerr_endline "Kind 2 did not exit in time, terminating." ;
           Stdlib.flush_all () ;
+          (* The solvers first: they are children of this process, and
+             killing it does not kill them. One left behind holds the
+             standard output of Kind 2 open on Windows, where a child
+             inherits every inheritable handle, so whoever reads that
+             output keeps waiting for an end that never comes. *)
+          ( try SMTSolver.destroy_all_of_process () with _ -> () ) ;
           Unix._exit status)
         ()
       |> ignore
--- a/src/smt/SMTSolver.ml
+++ b/src/smt/SMTSolver.ml
@@ -21,6 +21,7 @@ open Lib
 
 exception Unknown
 exception Timeout
+exception Exiting
 
 
 module IntMap = Map.Make(
@@ -76,10 +77,30 @@ See [destroy_all]. *)
 let all_solvers = ref IntMap.empty
 let all_solvers_lock = Mutex.create ()
 
-(** Registers a solver, owned by the calling domain. *)
+(* Latched by [destroy_all_of_process]: no solver may be created from
+   then on.
+
+   The sweep before the process exits is only final if nothing can
+   start a solver behind it. Something can: an engine the supervisor
+   gave up on keeps running in the background, and it dies with the
+   process rather than at a point we choose. A solver it starts after
+   the sweep is left running, with nobody to kill it.
+
+   Such a solver outlives Kind 2, and on Windows it holds the standard
+   output of Kind 2 open, since a child there inherits every
+   inheritable handle rather than only the ones it was given. Whoever
+   reads that output, the regression harness for one, then waits for an
+   end that never comes, long after Kind 2 itself is gone. *)
+let no_new_solvers = Atomic.make false
+
+(** Registers a solver, owned by the calling domain. Raises [Exiting]
+    if the process is on its way out, in which case the caller must
+    dispose of the instance it just created: past that point nothing
+    else will. *)
 let add_solver ( { id } as solver ) =
   let owner = (Domain.self () :> int) in
   Mutex.protect all_solvers_lock (fun () ->
+    if Atomic.get no_new_solvers then raise Exiting ;
     all_solvers := IntMap.add id (owner, solver) !all_solvers)
 
 (** Forgets a solver. *)
@@ -201,7 +222,14 @@ let create_instance
       last_assumptions = [| |]; }
   in
 
-  add_solver solver ;
+  (* The solver process is already running: it is started by the
+     functor application above, before there is anything to register.
+     If the sweep has been and gone, kill it here, since it would
+     otherwise be the one solver that outlives Kind 2. *)
+  ( try add_solver solver with Exiting ->
+      let module S = (val fomodule) in
+      S.kill_instance () ;
+      raise Exiting ) ;
 
   solver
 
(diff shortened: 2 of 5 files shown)


daniel-larraz:
## Review

I read the five changed files and their surrounding call paths, and re-ran the new test locally, including the two mutations the description claims.

**Verified.** `dune build @tests/ounit/utils/runtest --force` passes. Both mutations reproduce as described:

| Mutation | Result |
|---|---|
| drop `raise Exiting` from `add_solver` | `expected exception SMTSolver.Exiting, but no exception was raised` |
| keep the raise, drop `S.kill_instance ()` from `create_instance` | `a solver process was left running` (after the 5s grace) |

So the test is load-bearing on both halves of the fix, and the leak is deterministic on Linux.

The core design is right: latching `no_new_solvers` inside the same critical section that swaps the map is what makes the sweep final, and killing under `create_instance`'s handler is the only place that can catch the solver started by the functor application. `kill_instance` is exception-proof in both drivers (`SMTLIBSolver.ml:1159`, `yicesNative.ml:1381`), so the `IntMap.iter` sweep cannot abort part way and strand the remaining solvers.

### 1. `Exiting` escaping an abandoned engine logs a fatal error on a successful run

The normal supervisor path ends at `kind2Flow.ml:1058` (and 1014, 1097, 1215) → `post_clean_exit`. The last `slaughter_kids` before that is the one at `kind2Flow.ml:807`, with `exiting:false`, and it resets the flag at `kind2Flow.ml:401`:

```ocaml
if not exiting then EngineDomains.set_terminating false ;
```

So `post_clean_exit` runs the sweep with `EngineDomains.is_terminating () = false`. An engine abandoned in an earlier analysis that now calls `create_instance` gets `Exiting`, which unwinds to `on_exit_engine` (`kind2Flow.ml:491`). With `is_terminating ()` false it goes to `status_of_exn`, and falls through to the catch-all at `kind2Flow.ml:292`:

```ocaml
| e -> (
  InvarManager.print_stats !latest_trans_sys ;
  KEvent.log L_fatal "Runtime error in %a: %s" ... )
```

`Log.log` writes straight to the output channel (`log.ml:340`) — there is no relay — and this fires *after* `KEvent.terminate_log ()` has closed the XML/JSON tags. So the payoff in the exact scenario this PR is built for is a `L_fatal "Runtime error in <engine>: SMTSolver.Exiting"` plus a stats dump appended past `</Results>`.

It is a narrow race (between the sweep and `exit status`), and still better than the leaked solver it replaces. But it is the same window the PR argues is real, so I would close it. One line in `post_clean_exit`, before the sweep, matching the intent already stated at `kind2Flow.ml:398-400`:

```ocaml
EngineDomains.set_terminating true ;
SMTSolver.destroy_all_of_process () ;
```

That also makes `shutting_down` true, which is consistent with the sweep now killing unconditionally. Alternatively a `| SMTSolver.Exiting -> status` case in `status_of_exn`, but the flag is the more faithful statement of "the process is going away".

### 2. Residual window: the solver is running before `add_solver` is reached

The comment at `SMTSolver.ml:225-228` is right that the process exists before registration, but the handler only closes the window for an engine that *reaches* `add_solver`. `SMTLIBSolver.create_instance` forks at `SMTLIBSolver.ml:950` and then does the whole `set-logic` / header exchange, waiting on replies, before the functor returns. If the sweep and `exit` land inside that exchange — on Windows, where process startup is slowest and the consequence is the hang — the solver is neither in the map nor killed, and this fix does not catch it.

Not a blocker, and the `cloexec` work scoped out in "Not in scope" is the real answer. Two cheap narrowings if wanted here:

- Check the latch *before* the functor application too, so the common post-sweep case never spawns a process at all (right now every abandoned engine that tries still starts a z3 and kills it — a real spawn that momentarily inherits Kind 2's standard output on Windows).
- Or register the pid in `SMTLIBSolver.create_instance` immediately after `Unix.create_process`, which would make the sweep genuinely total.

### 3. `dune runtest` now needs Z3 on `PATH`

`tests/ounit/utils/dune` pulls a real solver into the unit suite for the first time. CI installs Z3 on all three platforms before `make test`, so CI is fine — but a contributor without Z3 now gets a failing `dune runtest` where the ounit suite used to be self-contained. A `skip_if` when the binary is not resolvable would keep that property.

### Smaller notes

- `destroy_all_of_process` (`SMTSolver.ml:277-282`) and `kill_solvers_of_domain` (`SMTSolver.ml:296-301`) now have byte-identical `IntMap.iter` bodies. A shared `kill_entries entries` would be worth the three lines.
- The watchdog at `kind2Flow.ml:430` now acquires `all_solvers_lock` before `Unix._exit`. Every critical section on that lock is a pure map operation with no I/O, so it cannot realistically block — but the last-resort path's whole value is that it cannot hang, and it now has a dependency it did not have. Worth a sentence in the comment acknowledging why that is safe.
- `no_new_solvers` is only ever read and written under `all_solvers_lock`, so `Atomic` buys nothing over a plain `ref` there. Harmless, just noting it in case the `Atomic` was meant to signal a lock-free read somewhere.
- `SMTSolver.mli` documents `Exiting` on the exception itself, but `create_instance`'s own doc comment does not mention it. A `@raise Exiting` there would help callers, since it is now part of that function's contract.
- The `Exiting` path kills the process but does not close the three pipe fds of the half-built instance. Irrelevant at exit; mentioning only so it is a deliberate omission rather than an oversight.

**Overall:** correct fix, well-targeted, and the test earns its place. Item 1 is the one I would want addressed before merge; 2 and 3 are judgement calls that may reasonably be deferred.

daniel-larraz:
Thanks — item 1 was real and is fixed, and the rest is addressed or answered below.

### 1. `Exiting` reported as a crash on a successful run

Confirmed, and fixed. The chain is as described: `slaughter_kids ~exiting:false` clears the flag at `kind2Flow.ml:401`, so `post_clean_exit` swept with `is_terminating () = false`, and `on_exit_engine` then sent `Exiting` through `status_of_exn` to the catch-all. The message is `Runtime failure in <engine>: SMTSolver.Exiting` rather than `Runtime error in`, but the effect is the one you describe, including the stats dump landing after `terminate_log ()` has closed the tags.

Fixed with your first suggestion, `EngineDomains.set_terminating true` before the sweep, which also makes `shutting_down` true and so agrees with the sweep killing unconditionally.

### 2. Residual window before `add_solver` is reached

Left open, and I want to be explicit that I tried your first narrowing and backed it out.

Refusing before the functor application does remove the spawn in the common case. It also takes the test with it: once `create_instance` raises before spawning, neither the check under the lock nor the kill in the handler is reachable from the only scenario the test can set up, and both of your mutations stop failing it.

| Mutation | with the pre-check | without it |
|---|---|---|
| drop the latch check in `add_solver` | passes | `expected exception SMTSolver.Exiting` |
| drop `S.kill_instance ()` from the handler | passes | `a solver process was left running` |

Trading the two mutations that hold the fix up for a window that stays open either way seemed a bad deal, especially as the window is closed properly by not letting solvers inherit handles they were not given. That is the follow-up, and it also makes the momentary inheritance you point at harmless.

### 3. `dune runtest` needing Z3

Fixed: the test skips when `Lib.find_on_path (Flags.Smt.z3_bin ())` finds nothing, so the suite stays self-contained in a checkout without a solver. Verified by running the test binary with Z3 off `PATH`: `OK: Cases: 1 Skip: 1`.

### Smaller notes

- `kill_entries` factored out, used by both sweeps.
- Watchdog comment now says why waiting for `all_solvers_lock` cannot become the reason it never reaches `_exit`: every critical section on it is a map operation with no I/O.
- `no_new_solvers` is now a `ref`. You were right that the `Atomic` bought nothing; it was briefly justified by the pre-check reading it outside the lock, and that is gone.
- `@raise Exiting` added to `create_instance`.
- The pipes of a refused instance are now commented as deliberate.

### Validation

Both mutations fail the test again, the skip works, and the ounit suite and 475 regression tests pass in 34.6s.

PR #1447, opened 2026-08-23
Stop a solver inheriting pipes it was not given

daniel-larraz:
Follow-up to #1446, and the part scoped out of it. That one stops a solver outliving Kind 2; this one stops a solver being handed pipes that are not its own, which turns out to be why an orphaned one could never stop by itself.

## A solver is handed the pipes of every solver before it, and of itself

The pipes of a solver are inheritable, so each solver started after it gets them too. Dumping the descriptors of the first solver of a run says what that means:

```
FD 0 -> pipe:[8088894]      its standard input
FD 1 -> pipe:[8088895]      its standard output
FD 2 -> pipe:[8088896]      its standard error
FD 8 -> pipe:[8088894]      the write end of its own standard input
FD 9 -> pipe:[8088895]
FD 11 -> pipe:[8088896]
```

A solver holds the write end of its own standard input. Closing the Kind 2 end therefore never ends it, and `z3 -in` waits for a command that cannot arrive. **Before this change a solver could not terminate itself on any platform**, and `kill_instance` sending SIGKILL was the only thing that ever stopped one. That is the other half of a leaked solver hanging whoever reads its output, and it is not specific to Windows: running the new test binary outside dune, which kills what it starts, leaves three solvers running for good before this change and none after it.

Later solvers hold the pipes of the earlier ones as well, so Kind 2 does not see the end of the output of a solver that has died until the last solver holding its pipe goes.

Create them closed on exec. `Unix.create_process` duplicates the three ends it is given onto the standard descriptors of the child whatever the flag says, so the solver still gets its own; what stops is everything else reaching it.

## On Windows a solver holds the descriptors of Kind 2

A child there inherits every inheritable handle, not only the three it is given, so a solver holds the standard input, output and error of Kind 2 as well. That is what turned a leaked solver into a [job that hung for 94 minutes](https://github.com/daniel-larraz/kind2/actions/runs/31332087766) and, later, into the [failure that identified it](https://github.com/kind2-mc/kind2/actions/runs/32657498757/job/97238457246): Kind 2 had exited, and the harness reading its output was waiting on a pipe a solver still held.

Clear the flag on those three descriptors at startup, on Windows only. Elsewhere a child of `Unix.create_process` has the three it is given duplicated over its own and never sees these.

The certificate paths run `Sys.command`, which does expect to inherit them, and nothing stops them running on Windows. What they produce there is unusable in any case — `#!/bin/sh` checker scripts, command lines redirecting to `/dev/null` — so what is lost is the console output of something that cannot work there anyway. Handing those paths the descriptors is the way to fix that if it ever matters, rather than leaving them inheritable for every child.

## Validation

A new ounit test reads the descriptors of every solver of a run and requires each to hold exactly the three it was given:

| | first solver | second | third |
|---|---|---|---|
| before | 6 | 9 | 12 |
| after | 3 | 3 | 3 |

It reverts to `expected: 3 but got: 6` when the flag is taken off, skips where there is no `/proc` or no solver installed, and disposes of the solvers before checking, so that a failing run does not leave behind the very processes it is about.

The ounit suite and the 482 regression tests pass, in 35.1s.

The Windows half cannot be exercised from Linux. What it puts at risk is the output of Kind 2 itself, and the solvers, still working there — which the suite covers on Windows CI, and did on the last run of this branch.

## Not in scope

Pipes only. Every file Kind 2 opens is still inheritable on all platforms — the SMT trace files, the XML and JSON output, the debug log, the generated contracts, twenty-five `open_out` in all. On Windows a leaked solver holding those keeps them locked against being deleted or renamed. Worth an increment of its own, most likely as an `open_out` that clears the flag rather than twenty-five call sites.

### DIFF
--- a/src/kind2.ml
+++ b/src/kind2.ml
@@ -116,10 +116,13 @@ let setup : unit -> any_input = fun () ->
   if Sys.win32 then (
     let keep fd name =
       try Unix.set_close_on_exec fd with Unix.Unix_error (e, _, _) ->
-        (* Worth a line: this is the failure that puts the standard
-           output of Kind 2 back within reach of its solvers, on the
-           one platform where that hangs whoever reads it. *)
-        Debug.kind2 "Could not keep %s from the children of Kind 2: %s"
+        (* Loud rather than logged: failing here quietly puts the
+           output of Kind 2 back within reach of its solvers, on the one
+           platform where that leaves whoever reads it waiting. It
+           should never happen, and there is no sign of it if it does. *)
+        KEvent.log L_warn
+          "Could not keep %s from the children of Kind 2 (%s).@ A solver \
+           outliving Kind 2 may hold it open."
           name (Unix.error_message e)
     in
     keep Unix.stdin "stdin" ;
--- a/src/kind2.ml
+++ b/src/kind2.ml
@@ -88,6 +88,28 @@ let setup : unit -> any_input = fun () ->
     /!\ ================================================================== /!\
   *)
 
+  (* Keep the output of Kind 2 to itself.
+
+     On Windows a child inherits every handle marked inheritable, not
+     only the three it is given, so a solver ends up holding the
+     standard output and error of Kind 2 as well as its own pipes. A
+     solver that outlives Kind 2 then keeps them open, and whoever is
+     reading that output waits for an end that never comes, long after
+     Kind 2 is gone. Clearing the flag leaves the solvers with the
+     pipes they are given: [Unix.create_process] duplicates those as
+     inheritable itself.
+
+     Only on Windows. Elsewhere a child of [Unix.create_process] gets
+     the three descriptors it is given duplicated over its own, and
+     never sees these; whereas [Sys.command], which the certificate
+     paths use, does expect to inherit them. *)
+  if Sys.win32 then (
+    try
+      Unix.set_close_on_exec Unix.stdout ;
+      Unix.set_close_on_exec Unix.stderr
+    with Unix.Unix_error _ -> ()
+  ) ;
+
   (* Raise exception on CTRL+C. *)
   Sys.catch_break true ;
 
--- a/src/kind2.ml
+++ b/src/kind2.ml
@@ -88,26 +88,43 @@ let setup : unit -> any_input = fun () ->
     /!\ ================================================================== /!\
   *)
 
-  (* Keep the output of Kind 2 to itself.
+  (* Keep the descriptors of Kind 2 to itself.
 
      On Windows a child inherits every handle marked inheritable, not
      only the three it is given, so a solver ends up holding the
-     standard output and error of Kind 2 as well as its own pipes. A
-     solver that outlives Kind 2 then keeps them open, and whoever is
-     reading that output waits for an end that never comes, long after
-     Kind 2 is gone. Clearing the flag leaves the solvers with the
-     pipes they are given: [Unix.create_process] duplicates those as
-     inheritable itself.
+     standard input, output and error of Kind 2 as well as its own
+     pipes. A solver that outlives Kind 2 then keeps them open, and
+     whoever is reading that output waits for an end that never comes,
+     long after Kind 2 is gone. Its standard input goes the same way:
+     a harness writing to Kind 2 would not see its reader go away.
+     Clearing the flag leaves the solvers with the pipes they are
+     given, which [Unix.create_process] duplicates as inheritable
+     itself.
 
      Only on Windows. Elsewhere a child of [Unix.create_process] gets
      the three descriptors it is given duplicated over its own, and
-     never sees these; whereas [Sys.command], which the certificate
-     paths use, does expect to inherit them. *)
+     never sees these.
+
+     [Sys.command] does expect to inherit them, and the certificate
+     paths, which use it, are not guarded against running here: the
+     output of what they start goes nowhere from now on. What those
+     paths produce is unusable on Windows in any case, being [#!/bin/sh]
+     scripts and command lines that redirect to [/dev/null], so what is
+     lost is the console output of something that cannot work there
+     anyway. Making them work is a matter of handing them the
+     descriptors, not of leaving these inheritable for every child. *)
   if Sys.win32 then (
-    try
-      Unix.set_close_on_exec Unix.stdout ;
-      Unix.set_close_on_exec Unix.stderr
-    with Unix.Unix_error _ -> ()
+    let keep fd name =
+      try Unix.set_close_on_exec fd with Unix.Unix_error (e, _, _) ->
+        (* Worth a line: this is the failure that puts the standard
+           output of Kind 2 back within reach of its solvers, on the
+           one platform where that hangs whoever reads it. *)
+        Debug.kind2 "Could not keep %s from the children of Kind 2: %s"
+          name (Unix.error_message e)
+    in
+    keep Unix.stdin "stdin" ;
+    keep Unix.stdout "stdout" ;
+    keep Unix.stderr "stderr"
   ) ;
 
   (* Raise exception on CTRL+C. *)
--- a/src/smt/SMTLIBSolver.ml
+++ b/src/smt/SMTLIBSolver.ml
@@ -942,9 +942,16 @@ module Make (Driver : SMTLIBSolverDriver) : SolverSig.S = struct
     let solver_executable = solver_cmd.(0) in
     
     (* Create pipes for input, output and error output *)
-    let solver_stdin_in, solver_stdin_out = Unix.pipe () in
-    let solver_stdout_in, solver_stdout_out = Unix.pipe () in 
-    let solver_stderr_in, solver_stderr_out = Unix.pipe () in 
+    (* Close these on exec, so that the only process to get them is
+       the solver they belong to. [Unix.create_process] passes the
+       three ends it is given to the child whatever this says, by
+       duplicating them onto its standard descriptors; what this stops
+       is every later solver inheriting the pipes of every earlier
+       one. A solver holding the write end of a pipe of a solver that
+       has died keeps Kind 2 from ever reading the end of its output. *)
+    let solver_stdin_in, solver_stdin_out = Unix.pipe ~cloexec:true () in
+    let solver_stdout_in, solver_stdout_out = Unix.pipe ~cloexec:true () in
+    let solver_stderr_in, solver_stderr_out = Unix.pipe ~cloexec:true () in
     
     (* Create solver process *)
     let solver_pid = 
--- a/src/smt/yicesNative.ml
+++ b/src/smt/yicesNative.ml
@@ -1206,10 +1206,13 @@ let create_instance
      TODO: expand ~ *)
   let solver_executable = solver_cmd.(0) in
 
-  (* Create pipes for input, output and error output *)
-  let solver_stdin_in, solver_stdin_out = Unix.pipe () in
-  let solver_stdout_in, solver_stdout_out = Unix.pipe () in 
-  let solver_stderr_in, solver_stderr_out = Unix.pipe () in 
+  (* Create pipes for input, output and error output.
+
+     Closed on exec, so that the only process to get them is the solver
+     they belong to: see the same call in [SMTLIBSolver]. *)
+  let solver_stdin_in, solver_stdin_out = Unix.pipe ~cloexec:true () in
+  let solver_stdout_in, solver_stdout_out = Unix.pipe ~cloexec:true () in
+  let solver_stderr_in, solver_stderr_out = Unix.pipe ~cloexec:true () in
 
   
   (* Create solver process *)
--- a/tests/ounit/utils/dune
+++ b/tests/ounit/utils/dune
@@ -1,3 +1,3 @@
 (tests
-  (names testGraph testSolverExit)
+  (names testGraph testSolverExit testSolverPipes)
   (libraries ounit2 unix kind2dev))
--- a/tests/ounit/utils/testSolverCommon.ml
+++ b/tests/ounit/utils/testSolverCommon.ml
@@ -1,0 +1,30 @@
+(* This file is part of the Kind 2 model checker.
+
+   Copyright (c) 2026 by the Board of Trustees of the University of Iowa
+
+   Licensed under the Apache License, Version 2.0 (the "License"); you
+   may not use this file except in compliance with the License.  You
+   may obtain a copy of the License at
+
+   http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or
+   implied. See the License for the specific language governing
+   permissions and limitations under the License.
+
+*)
+
+(* Shared by the tests that need a solver of their own to look at. *)
+
+let new_solver () = SMTSolver.create_instance `None `Z3_SMTLIB
+
+(* Whether the solver [new_solver] asks for can be found. The rest of
+   the unit suite needs nothing installed, and a checkout without a
+   solver should not start failing it. CI installs Z3 on every platform
+   it tests, so skipping on this costs no coverage there. *)
+let solver_missing () =
+  match Lib.find_on_path (Flags.Smt.z3_bin ()) with
+  | _ -> false
+  | exception Not_found -> true
--- a/tests/ounit/utils/testSolverExit.ml
+++ b/tests/ounit/utils/testSolverExit.ml
@@ -27,16 +27,7 @@ open OUnit2
    standard output of Kind 2 open, so whoever reads that output waits
    for an end that never comes. *)
 
-let new_solver () = SMTSolver.create_instance `None `Z3_SMTLIB
-
-(* Whether the solver [new_solver] asks for can be found. The rest of
-   the unit suite needs nothing installed, and a checkout without a
-   solver should not start failing it. CI installs Z3 on every platform
-   it tests, so this skips nowhere that matters. *)
-let solver_missing () =
-  match Lib.find_on_path (Flags.Smt.z3_bin ()) with
-  | _ -> false
-  | exception Not_found -> true
+open TestSolverCommon
 
 (* Reap whatever has exited and report whether a child process is still
    running. The only children here are the solvers created above. *)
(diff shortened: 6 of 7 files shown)


daniel-larraz:
Rebased onto `main` now that #1446 has landed. The only conflict was `tests/ounit/utils/dune`, where both branches add a test to `names`; resolved by keeping all three.

That left both solver tests carrying a copy of the same two helpers — one to ask for an instance, one to say whether the solver it asks for is installed — so there is a second commit putting them in a module of their own.

Re-verified on the rebase: both tests still fail when the thing they test is removed (`a solver holds 9 pipes where the first holds 6` without `cloexec`, `expected exception SMTSolver.Exiting` without the latch), and the 482 regression tests pass in 35.4s.

daniel-larraz:
## Review

Built and ran the branch (b86222d3, on top of merged #1446), and checked the OCaml/Linux semantics the change rests on with a standalone program.

**Verified.**

- All three ounit executables in `tests/ounit/utils/` pass.
- Mutation — revert the three `~cloexec:true` in `SMTLIBSolver.ml:952-954` — fails with `a solver holds 9 pipes where the first holds 6`. That is the table in the description exactly.
- Independently, a standalone program mimicking the pipe/spawn dance confirms `~cloexec:true` does **not** stop `Unix.create_process` handing the child its three descriptors: the child's fds 0/1/2 still resolve to `pipe:[…]`. Counts across three successive children: 6 → 12 without the flag, 3 → 3 with it.

### The change buys more than the descriptor count says

Dumping a solver's whole fd table shows where the "6" comes from. It is 0/1/2 plus the three *parent-side* ends of its own pipes — including `solver_stdin_out`, **the write end of its own stdin**. So before this change a solver never saw EOF on stdin when Kind 2 closed its end: it was holding the pipe open against itself.

That is directly observable. On the mutated build the three solvers `testSolverPipes` starts outlive the test process indefinitely — I collected a dozen stray `z3 -smt2 -in` while iterating. On the fixed build, zero linger: they get EOF and exit on their own.

Worth putting in the PR description, because it is a stronger statement than the leak: previously an orphaned solver could never terminate itself on **any** platform, and `kill_instance`'s SIGKILL was the only thing that ever stopped one. That is the other half of the 94-minute hang, and it makes "not specific to Windows" concrete.

### 1. `last` is not the last solver

`testSolverPipes.ml:76-81` takes the head of `List.filter (fun pid -> pid <> first) (children ())`. `/proc/<pid>/task/<tid>/children` is in creation order (the kernel appends on fork; confirmed it prints `p1 p2 p3`), so this inspects the **second** solver, which inherited from one predecessor — 9 pipes. The third holds 12. The comment says "Two more, so that the last one could have inherited from three", which is the case the code then does not measure.

It still catches the regression, so this is a strengthening, not a bug: take the last element, or assert on all three.

### 2. False failure where `/proc/<pid>/children` is unavailable

The guard at `testSolverPipes.ml:60` checks `/proc/self/task` exists, but `children ()` needs the per-task `children` file, which is `CONFIG_PROC_CHILDREN`. Where that is off, `open_in` fails, `children ()` returns `[]`, and the test hits `assert_failure "expected one solver, found 0"` instead of skipping — the outcome the skip guards were added to avoid.

Same shape in `pipes_held`: `Sys.readdir "/proc/<pid>/fd"` raises `Sys_error` uncaught if the solver is gone between the two reads.

### 3. The Windows `Sys.command` argument is asserted, not enforced

The comment at `kind2.ml:102-105` rests on the certificate paths being POSIX-only. Two of them re-run Kind 2 itself and rely on inheriting stdout/stderr to show the child's output:

- `certify_observer` — `certifChecker.ml:2930`
- `rerun_kind2_on_fec` — `certifChecker.ml:3088`

Neither is gated on `Sys.win32`. The argument is indirect — that the artifacts these paths produce (`#!/bin/sh` checker scripts written at `certifChecker.ml:2901`, `jkind_command_line`'s `>/dev/null 2>`) are useless on Windows. True, but `--certif` on Windows still *reaches* `certify_observer`, and after this change that re-run's output goes nowhere rather than to the console.

`proof.ml:252` is fine either way — `Unix.open_process_full` supplies its own pipes.

Cheap hardening if it should be enforced rather than argued: run those two through `Unix.create_process … Unix.stdin Unix.stdout Unix.stderr`, which OCaml's Windows `create_process` duplicates as inheritable regardless of the flag. Otherwise the comment could say the paths are unguarded and known-unusable there, rather than implying they cannot be reached.

### 4. Residual: pipes only, and stdout/stderr only

Every `open_out` in the process is still inheritable on all platforms — the SMT trace file (`SMTLIBSolver.ml:778`, one per solver), the XML/JSON output channel (`kEvent.ml:435`), the debug log (`kind2.ml:72`), the contract-generation files. So the descriptor leak is fixed for pipes and untouched for regular files, and on Windows a leaked solver still holds Kind 2's *output file* handles, which keeps them locked against deletion or rename.

Not a blocker, and arguably the next increment. But the title claims "nothing else", and that part is not true yet.

### Smaller notes

- `kind2.ml:108-109` clears stdout and stderr but not stdin. On Windows a solver still inherits Kind 2's stdin; if that is a pipe from a harness, the harness's write end never observes the reader going away. Probably deliberate — worth a sentence saying so.
- The `with Unix.Unix_error _ -> ()` around those two calls swallows silently. A failure there reinstates exactly the bug being fixed, on the platform where it cannot be tested, with no trace. A `Debug.kind2` line would cost nothing.
- `testSolverPipes` never disposes of its three solvers. With the fix they exit on EOF (verified), but a *failing* run leaves them running forever — which is how the strays above accumulated. Taking the counts and then calling `SMTSolver.destroy_all_of_process ()` before the assertion would keep a red test from leaving processes behind.
- `testSolverPipes.ml:90` asserts `held_by_last <= held_by_first`. Both should be exactly 3; the `<=` form also passes if a future change raised the baseline uniformly. `=`, or an exact count, would fail more informatively.
- `TestSolverCommon` is not in `(names …)`, so dune links it into `testGraph` as well. Harmless — noting only in case a `(library)` was intended.
- Checked `Lib.find_on_path`: it appends `.exe` on Windows (`lib.ml:1026-1034`), so `solver_missing ()` will not quietly skip on Windows CI. The claim that the suite exercises the Windows half holds.

**Overall:** the mechanism is right, the measurement backs it, and the new test is load-bearing. Items 1 and 2 are small test fixes; 3 is the one place the Windows half could regress something, and it deserves either a guard or a more honest comment; 4 is scope worth naming.

daniel-larraz:
Thanks — all four addressed, and the finding about stdin is now the headline of the description.

### The stdin observation

Confirmed independently, and it is sharper than I had it. Dumping the descriptors of the first solver of a run:

```
FD 0 -> pipe:[8088894]      its standard input
FD 8 -> pipe:[8088894]      the write end of its own standard input
```

Same pipe. So closing the Kind 2 end never ends it. Running the test binary outside dune — which was hiding this by killing the process group of the action — leaves three solvers running for good before the change and none after it. That is in the description now, along with the point that a solver could not terminate itself on any platform before this.

### 1. `last` is not the last solver

Right, and fixed by reading all three rather than picking one. Each is now required to hold exactly the three it was given, rather than no more than the first: the counts turn out identical for every solver, and an exact one says what went wrong when it is not. The mutation now reports `expected: 3 but got: 6`.

### 2. False failure where the children are not listed

Fixed: the test skips when the kernel does not list them, rather than reporting that it found no solver. `pipes_held` returns an option, so a solver that has gone is named instead of raising `Sys_error`.

### 3. The Windows `Sys.command` argument

Taken as the second option — the comment now says the paths are unguarded and reachable, that their output goes nowhere from this point, and that what they produce on Windows is unusable anyway, so what is lost is the console output of something that cannot work there. It also says that handing them the descriptors is the fix if that ever matters, rather than leaving them inheritable for every child.

I did consider the guard. `certify_observer` and `rerun_kind2_on_fec` both pass a shell string to `Sys.command`, so giving them explicit descriptors means naming a shell to run them through, which is its own Windows problem. Not worth it for a path that cannot work there.

### 4. Pipes only

Named in the description rather than fixed: twenty-five `open_out`, and on Windows a leaked solver holding those keeps them locked against deletion or rename. Most likely an increment of its own, as an `open_out` that clears the flag rather than twenty-five call sites. The PR title is now **"Stop a solver inheriting pipes it was not given"**, which does not claim more than that.

### Smaller notes

- Standard input is cleared too now, with the reason: a harness writing to Kind 2 would not see its reader go away either.
- The failure to clear is logged through `Debug.kind2` rather than swallowed.
- The test disposes of its solvers before checking the counts, so a failing run no longer leaves three behind — verified.
- `TestSolverCommon` being linked into `testGraph`: left as it is, harmless.
- Thanks for checking `Lib.find_on_path` on Windows; that was the assumption the Windows coverage rests on.

### Validation

The mutation still fails the test and leaves nothing behind; the ounit suite and the 482 regression tests pass in 35.1s.

daniel-larraz:
Verified against f357dc97.

**Re-ran everything.** Full `dune build` clean; all three ounit executables pass; no stray `z3` after the run. The mutation (revert the three `~cloexec:true`) now fails with

```
solver 169462 holds pipes it was not given, the ones of the solvers before it
expected: 3 but got: 6
```

and leaves **zero** solvers behind, where before it left three per run. The new `children ()` guard does not skip spuriously either — a normal kernel returns `Some []` before any solver exists, so the run reports `Skip: 0` rather than a silent pass.

**All four findings addressed:**

| | |
|---|---|
| 1. `last` was the second solver | Now reads all three and asserts each holds exactly `given = 3`. Ordering no longer relied on. |
| 2. False failure without `CONFIG_PROC_CHILDREN` | `children ()` returns an option; `skip_if (children () = None)`. `pipes_held` returns `None` on `Sys_error` and names the solver. |
| 3. Windows `Sys.command` asserted, not enforced | `kind2.ml:104-117` now states the paths are unguarded and says what is lost, rather than implying they cannot be reached. |
| 4. Residual file handles | Named in "Not in scope", with a count and the right shape for the fix. |

Smaller notes also taken: stdin cleared alongside stdout/stderr, failures logged instead of swallowed, solvers disposed before the assertions, exact rather than `<=`.

The description is now materially better than what I reviewed — the descriptor dump showing a solver holding the write end of its own standard input is the clearest statement of why an orphaned solver could never stop, and the 6/9/12 table matches what I measured independently.

**Three things left, none blocking:**

- `testSolverPipes.ml:77` — `for _ = 1 to solvers do ignore (new_solver ()) done` sits outside the cleanup. If the second `new_solver ()` throws, the first is never disposed of: the leak the cleanup was added for, reached by a different path. `Fun.protect ~finally:SMTSolver.destroy_all_of_process` around the body closes it.
- `Debug.kind2` is gated on `--debug kind2` (`debug.mli:38`), so the "could not keep stdout from the children" line will not be seen in the field — which is the one failure that silently reinstates the Windows hang. `L_warn` would surface it; given how close to never it fires, `Debug` is a defensible call.
- The exact `= 3` is my own suggestion's cost: a stray inheritable pipe descriptor handed to the test process by whatever runs it would now fail the test, where `<=` was immune. Dune does not do that, so this is a note for the record rather than a change I would make.

daniel-larraz:
All three taken, in `ded758ae`.

### The solvers are started outside the cleanup

Fixed by arming it first — `Fun.protect ~finally:SMTSolver.destroy_all_of_process` now wraps the starting as well as the checking. Verified against the case you describe rather than by reading: making the second `new_solver ()` raise leaves zero solvers behind, where it would have stranded the first.

### `Debug.kind2` will not be seen

Agreed, and changed to `L_warn`. You called `Debug` defensible on the grounds that it fires close to never, but that cuts the other way: something that never happens costs nothing to say loudly, and this is the one failure that silently reinstates the hang, on the platform where it cannot be tested. It reads

```
Could not keep stdout from the children of Kind 2 (<reason>). A solver outliving Kind 2 may hold it open.
```

### The cost of the exact `= 3`

Not weakened, but the failure now explains itself. `pipes_held` names the pipes instead of counting them, so the mutation reports

```
solver 174861 holds 0 -> pipe:[8155267], 1 -> pipe:[8155268], 2 -> pipe:[8155269],
                     8 -> pipe:[8155267], 9 -> pipe:[8155268], 11 -> pipe:[8155269]
expected: 3 but got: 6
```

The duplicate is right there — `0` and `8` are the same pipe — which is the thing the count was standing in for. A stray pipe from whatever runs the test would be just as legible, so the exact assertion keeps its bite without the confusing failure it was risking.

### Validation

Mutation still fails and leaves nothing behind; a solver failing to start part way leaves nothing behind; the ounit suite and the 482 regression tests pass in 35.0s.

daniel-larraz:
Two failures on `ded758ae`, one mine and one not.

## CI: mine, and exactly what was predicted

`kind2-build (linux-arm64)`:

```
solver 11418 holds 0 -> pipe:[29643], 1 -> pipe:[29644], 2 -> pipe:[29645],
                   142 -> pipe:[13040], 145 -> pipe:[13041]
expected: 3 but got: 5
```

This is the cost of the exact `= 3` that was flagged in review, and I was wrong to answer that a stray would be legible enough. It is legible — those two are neither the solver's nor duplicates of its — but the test fails all the same. They are pipes the runner handed the test process, which the solver inherited along with everything else of ours.

Fixed in `04d0eaae` by reading what this process holds before any solver exists and discounting those. What is left is what Kind 2 gave the solver, which is the subject. The pipes of a solver cannot be among them, since they are made after the reading is taken, so a solver handed the pipes of another, or holding its own twice, still counts them all.

The test now also makes a pipe of its own before that reading, so that a run anywhere covers what a run under CI does. Without one the discounting is only exercised on runners that happen to leak a descriptor, which is how this reached CI rather than my machine. With the flag reverted the report separates the two cleanly:

```
solver 211926 holds 0 -> pipe:[8211148], 1 -> pipe:[8211149], 2 -> pipe:[8211150],
                    7 -> pipe:[8211136], 8 -> pipe:[8211136],
                    10 -> pipe:[8211148], 11 -> pipe:[8211149], 13 -> pipe:[8211150],
                    of which 2 were already ours
expected: 3 but got: 6
```

`7` and `8` are the pipe we made; `0/10`, `1/11`, `2/13` are the three the solver holds twice.

## Benchmarks: not mine

One benchmark errored, `FMCAD08/Int/large/microwave28.lus`, `Invalid` on base and `Error` on head. The error is

```
<Error> Runtime error in property directed reachability (IA):
        File "src/terms/ltree.ml", line 1156: Assertion failed
```

IC3IA crashing on an unbound variable while folding a term, after BMC had already falsified the property. Nothing this branch touches can reach that: it changes how pipes are created, and on Linux that is the whole of it.

What settles it is that the same change passed the benchmarks twice and failed once:

| commit | benchmarks |
|---|---|
| `b86222d3` | pass |
| `f357dc97` | pass |
| `ded758ae` | **fail** |

and the only difference between the last two is a log line inside `if Sys.win32` and the test file. I also ran the two binaries CI built, on that exact benchmark, 150 times each on four cores: zero occurrences on either side. It is rare and timing-dependent, and which side of a comparison it lands on is chance.

Worth chasing on its own — the assertion sits in the area #1444 and #1445 changed recently — but not here.

daniel-larraz:
Verified at 04d0eaae.

**The fix is right, and I reproduced the CI failure mode directly.** Beyond the in-test stand-in pipe, I ran the built binary with a genuinely *externally inherited* pipe — `testSolverPipes.exe 9< <(sleep 20)`, which is what linux-arm64 did with fds 142/145. Passes with it and without it. Full build clean, all three ounit executables pass, no stray `z3`.

The mutation still fails, and the report separates the two kinds cleanly:

```
solver 402240 holds 0 -> pipe:[8285018], 1 -> pipe:[8285019], 2 -> pipe:[8285020],
                    7 -> pipe:[8285006], 8 -> pipe:[8285006],
                    10 -> pipe:[8285018], 11 -> pipe:[8285019], 13 -> pipe:[8285020],
                    of which 2 were already ours
expected: 3 but got: 6
```

7/8 are the stand-in pipe (one inode, both ends), discounted; 0/10, 1/11, 2/13 are the solver's three held twice. That reads correctly at a glance.

**On the soundness of discounting by inode** — it holds, but the load-bearing invariant is slightly different from what the comment says. The comment argues the solver's pipes cannot be in `ours` because "they are made after this is taken". Newness alone is not quite enough: what actually rules out an alias is that every pipe in `ours` stays *open* for the whole test (the stand-in is held by the inner `Fun.protect`, the runner's by the runner), so its inode cannot be recycled under a solver pipe. Worth the half-sentence, since a future refactor that closed the stand-in early would quietly break it.

**One suggestion.** `Unix.pipe ()` for the stand-in is deliberately not `~cloexec`, and nothing says so out loud. If someone later tidies that to `~cloexec:true`, the discounting path stops being exercised and the test still passes — the same silent gap that let this reach CI in the first place. Cheap self-defence, since the stand-in is guaranteed to contribute two descriptors:

```ocaml
assert_bool
  "the stand-in pipe was not inherited: the discounting is untested"
  (List.length held - List.length own >= 2)
```

`>=` rather than `=` so a runner that leaks its own descriptors still passes.

**One limitation worth knowing, not fixing.** The test can no longer see a solver inheriting a pipe Kind 2 already held before it started creating solvers. That is inherent to discounting and correct in scope here — Kind 2 holds no such pipe in this test, and the cloexec change genuinely cannot affect descriptors it did not create. It just means this test constrains what Kind 2 *hands* its solvers, not everything a solver ends up holding.

daniel-larraz:
Both taken, in `46415327`.

### The soundness argument

You are right that the comment was resting on the wrong thing. Newness does not rule out an alias — the kernel can give a closed pipe's name to a new one, and a solver's pipe carrying a recycled name would be discounted and that solver let off. What rules it out is that every pipe in the reading stays open for the whole test: the stand-in until the end, the runner's for the runner to close. The comment says that now, and says it is the open-ness rather than the order that matters, so that a change closing the stand-in early has something to run into.

### Checking that the stand-in is inherited

Added, in your form — `>=` so a runner leaking descriptors of its own still passes — and the pipe now asks for `~cloexec:false` explicitly rather than relying on the default, with a line saying it is inheritable on purpose.

Verified it guards: making the stand-in close-on-exec trips it with

```
the pipe made for the solvers to inherit did not reach them, so nothing exercises the discounting
```

rather than passing quietly, which was the point.

### The limitation

Written into the comment rather than left implicit: what this cannot see is a solver inheriting a pipe Kind 2 held before it began making them. There is none here, and a descriptor Kind 2 never created is not what closing its own on exec is about.

### Validation

Your external-pipe check reproduced here too — `testSolverPipes.exe 9< <(sleep 20)` passes, with and without. The mutation still fails and leaves nothing behind, and the ounit suite and the 482 regression tests pass in 35.3s.

The benchmark check should want a re-run; as of the last comment nothing in that failure is attributable to this branch.

Issue #1449, opened 2026-08-24
Windows CI: Kind 2 hangs past its own timeout on test-issue-127 [slice_on] since #1445


daniel-larraz:
Since #1445 landed, the Windows CI job hangs intermittently on `test-issue-127.lus [slice_on]`, in a mode distinct from the two that #1446 and #1447 fixed: **Kind 2 itself is still running when the harness kills it**, two minutes past its own `--timeout 84`.

```
Killed after 204s: the run did not stop on its own, although it was given 84s
Kind 2 was still running when it was killed
```

## Occurrences

All on the same single test, all with the same signature:

| Run | Commit | Notes |
|---|---|---|
| [main @ 737ff590](https://github.com/kind2-mc/kind2/actions/runs/32662255012) | #1445 merge | first occurrence, before #1446 |
| [main @ 3f3016de](https://github.com/kind2-mc/kind2/actions/runs/32666896199) | #1446 merge | before #1447 |
| [PR #1447 @ 4641532](https://github.com/kind2-mc/kind2/actions/runs/32687318771) | | after both fixes |

That is 3 failures in the 4 Windows jobs that ran to completion since #1445, after ~14 consecutive green Windows jobs since the harness landed (#1434). The other Windows jobs in that window were cancelled by fail-fast from an unrelated failure, not green.

## What the signature says

- `proc.poll()` returned `None`: the Kind 2 **process** was alive at 204s. This is not a leaked solver holding the output pipes (#1446/#1447) — those report `Kind 2 had already exited (status 30)`.
- In all three occurrences the captured output stops **immediately after the analysis header** (`Analyzing _MAIN ... subsystems`): no engine output, and no `<Timeout> Wallclock timeout.` banner.
- On Windows the wall-clock timeout is enforced only by a check in the supervisor's 10ms polling loop (`invarManager.ml`, the `Sys.win32` check after `handle_events`); there is no SIGALRM. So the silence is ambiguous three ways:
  1. the supervisor loop stopped turning (the whole process wedged — a lock or a stop-the-world that cannot complete), or
  2. `TimeoutWall` fired and the teardown wedged before its first visible line (everything in `slaughter_kids` logs at `L_info`, invisible at the default level; the timeout banner comes only after it completes), or
  3. the teardown ran to the last-resort watchdog and `Unix._exit` never happened.

[...]

PR #1450, opened 2026-08-24
Arm the last resort before the teardown, not after it

daniel-larraz:
Hardens the exit path against the class of hang tracked in #1449, and validated by fault injection on both sides.

## The gap

The watchdog that terminates Kind 2 the hard way when its exit does not finish was started in `post_clean_exit` — *after* `slaughter_kids` had already run. Most of the teardown happens in `slaughter_kids`, so the last resort covered only the tail of what can wedge: a teardown stuck terminating the engines hung Kind 2 forever, with nothing to notice. The wall clock timeout does not reach there — it is ignored while an analysis is torn down, and on Windows it never existed outside the polling loop of the supervisor in the first place.

## The change

Arm the watchdog on entry to `on_exit`, before the first step of the teardown; the paths that reach `post_clean_exit` directly arm it there, as before. The exit status lives in a cell the watchdog reads when it fires, so arming early does not mean exiting with a stale status: `post_clean_exit` refines it once `status_of_exn` has run.

In the watchdog itself, the solvers are swept **before** the channels are flushed rather than after. The sweep is what frees whoever reads the output of Kind 2, so nothing may stand between the watchdog waking and it running; the tail of standard output is worth a try once the sweep is done, and no more than a try.

## Validation

By wedging `slaughter_kids` with an injected sleep and running `test-issue-127.lus` pinned to two cores, so that the wall clock timeout fires with every engine live:

| | outcome |
|---|---|
| `main`, wedged | still running 25s later, **silent** — the watchdog was never armed |
| this change, wedged | `Kind 2 did not exit in time, terminating.` and gone at ten seconds, solvers swept — single-engine and full-fleet scenarios alike, three runs of each |

The ounit suite and the 482 regression tests pass, in 35.9s.

## What it does not cover

The injection also exposed a gap this change leaves open, now recorded in #1449: the **between-analyses** `slaughter_kids`, where the wall clock timeout is equally suppressed and no exit is in sight, hangs Kind 2 forever with no output at all when wedged — and on Windows, so does the whole stretch from spawning the engines to the first iteration of the polling loop, since the polled timeout check only lives inside the loop. That is the window the CI hang's signature falls into, and it needs a design of its own rather than a bigger watchdog.

This is not the fix for the Windows CI failure of #1449: sampling that test on this branch still gives runs of 169s and 183s against a `--timeout` of 84s, where a teardown wedge would have been capped near 94s. That failure overruns its own timeout before the exit path is entered, and is a slow run rather than a hang. What this change fixes is the gap the injection above demonstrates, which is real either way.

### DIFF
--- a/src/kind2Flow.ml
+++ b/src/kind2Flow.ml
@@ -406,38 +406,67 @@ let slaughter_kids ?(exiting = false) process sys =
   if not exiting then Signals.set_sigalrm_timeout_from_flag ()
 
 
+(* Last resort for the whole exit path: a deadlock, a join that never
+   returns or a solver that refuses to die would leave Kind 2 hanging
+   forever once the teardown has begun, since the wall clock timeout is
+   disabled while an analysis is torn down, and on Windows there is no
+   wall clock timeout at all outside the polling loop of the
+   supervisor. Leave the process a few seconds to exit in an orderly
+   way, and terminate it the hard way if it does not.
+
+   Armed before the first step of the teardown rather than after the
+   engines have been dealt with: most of the teardown happens in
+   [slaughter_kids], and a watchdog that only starts after it covers
+   only the tail of what can wedge.
+
+   The status the watchdog exits with is kept in a cell rather than
+   captured, so that arming early does not mean exiting with a status
+   from before the exception was looked at: [post_clean_exit] refines
+   it once [status_of_exn] has run. *)
+let watchdog_status = Atomic.make ExitCodes.error
+let watchdog_armed = Atomic.make false
+
+let arm_exit_watchdog status =
+  Atomic.set watchdog_status status ;
+  if not (Atomic.exchange watchdog_armed true) then
+    ( try
+        Thread.create
+          (fun () ->
+            Thread.delay 10.0 ;
+            prerr_endline "Kind 2 did not exit in time, terminating." ;
+            (* The solvers first: they are children of this process,
+               and killing it does not kill them. One left behind
+               holds the standard output of Kind 2 open on Windows,
+               where a child inherits every inheritable handle, so
+               whoever reads that output keeps waiting for an end that
+               never comes.
+
+               This is the path that must not hang, and it waits for a
+               lock to take it. Every critical section that lock
+               guards is a map operation with no I/O in it, held for
+               as long as an insertion or a swap takes, so waiting for
+               it cannot become the reason we never get to [_exit]. *)
+            ( try SMTSolver.destroy_all_of_process () with _ -> () ) ;
+            (* Flushing after the sweep, not before: the sweep is what
+               frees whoever reads the output of Kind 2, so nothing may
+               stand between the watchdog waking and it running --
+               [prerr_endline] has already flushed the one channel this
+               path writes to. The tail of standard output is worth a
+               try once the sweep is done, and no more than a try. *)
+            ( try Stdlib.flush_all () with _ -> () ) ;
+            Unix._exit (Atomic.get watchdog_status))
+          ()
+        |> ignore
+      with _ -> () )
+
 (** Called after everything has been cleaned up. *)
 let post_clean_exit process base_status exn =
   (* Exit status of process depends on exception. *)
   let status = status_of_exn process base_status exn in
-  (* Last resort: engines abandoned in the background hold no lock the
-     exit path needs, but a deadlock or a solver that refuses to die
-     would otherwise leave Kind 2 hanging forever, since the wall clock
-     timeout is disabled while an analysis is torn down. Leave the
-     process a few seconds to exit in an orderly way, and terminate it
-     the hard way if it does not. *)
-  ( try
-      Thread.create
-        (fun () ->
-          Thread.delay 10.0 ;
-          prerr_endline "Kind 2 did not exit in time, terminating." ;
-          Stdlib.flush_all () ;
-          (* The solvers first: they are children of this process, and
-             killing it does not kill them. One left behind holds the
-             standard output of Kind 2 open on Windows, where a child
-             inherits every inheritable handle, so whoever reads that
-             output keeps waiting for an end that never comes.
-
-             This is the path that must not hang, and it now waits for
-             a lock to take it. Every critical section that lock
-             guards is a map operation with no I/O in it, held for as
-             long as an insertion or a swap takes, so waiting for it
-             cannot become the reason we never get to [_exit]. *)
-          ( try SMTSolver.destroy_all_of_process () with _ -> () ) ;
-          Unix._exit status)
-        ()
-      |> ignore
-    with _ -> () ) ;
+  (* Arms on the paths that come here directly; on the path through
+     [on_exit] the watchdog is already running and this refines the
+     status it will use. *)
+  arm_exit_watchdog status ;
   (* Close tags in JSON/XML output. *)
   KEvent.terminate_log () ;
   (* The engines that are still running are on their way out with the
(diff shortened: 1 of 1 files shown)


daniel-larraz:
Measurement note, so this PR is not credited with something it does not do.

I sampled the failing Windows test on top of this branch (8 runs, MathSAT, `--timeout 84`): 169s, 53s, 87s, 183s, 93s, 106s, 63s, 37s. Had the overrun been in the teardown this arms the watchdog over, every run would have capped near 84 + 10s. They do not, so **this PR does not fix the CI failure in #1449** — that failure turns out to be a run that overruns its own timeout before the exit path is ever entered, and is not a hang at all (details in #1449).

What this PR does fix stands unchanged, and is still worth having: the fault injection in the description shows that a teardown which wedges is otherwise unbounded — silent, forever, with no last resort — because the watchdog only started after `slaughter_kids` had already run. That is a real gap whether or not it is the one behind #1449.

So: worth merging on its own terms, not as a fix for #1449, and I have removed that framing from the last paragraph of the description.

daniel-larraz:
Reviewed by building the branch, running both suites, and fault-injecting in both directions.

## Verification

The central claim holds. With `Thread.delay 60.0` injected at the top of `slaughter_kids` (guarded on `exiting`), running `test-issue-127.lus --timeout 2` pinned to two cores:

| build | outcome |
|---|---|
| `HEAD~1` (#1447), wedged | still running when `timeout 30` killed it (exit 124), silent |
| this branch, wedged | `Kind 2 did not exit in time, terminating.` at ~10s, process gone at 12.1s, exit 30, no stray solvers |

Suites on the clean branch: 482 regression tests passed in 35.5s, ounit OK, no leftover `z3`/`mathsat` processes afterwards.

The comment at `src/kind2Flow.ml:443-447` about the registry lock checks out: `all_solvers_lock` only ever guards map operations, and `kill_entries` -> `kill_instance` reaps with `WNOHANG`, so the sweep cannot become the reason we never reach `_exit`.

## Findings

**1. `Stdlib.flush_all ()` can still keep the watchdog from reaching `Unix._exit`** — `src/kind2Flow.ml:456-457`

`try ... with _ -> ()` catches exceptions; it does not bound a blocking call. If standard output is a pipe whose reader has stalled, `flush_all` blocks and `Unix._exit` is never reached. Minimal repro mirroring the watchdog body -- main thread writing 10MB to a stalled pipe, watchdog thread doing `prerr_endline`, then `flush_all`, then `_exit`:

```
WATCHDOG FIRED
STILL RUNNING at t=12 -> watchdog thread starved
```

The marker after `flush_all` never prints. That is the signature recorded in #1449 as the unreproduced one-in-ten: watchdog printed, process survived.

The reorder in this PR narrows the window, since the sweep now happens first, but whoever reads the output of Kind 2 is only truly freed when Kind 2 closes its own standard output, which happens at `_exit`. So the guarantee the comment states is not yet complete. Cheap fix: run the flush on a helper thread and give it a bounded wait, then `_exit` unconditionally.

**2. Arming before `status_of_exn` means the watchdog exits with a status that reads as a normal verdict** — `src/kind2Flow.ml:421-425`, `src/kind2Flow.ml:487`

The comment says `post_clean_exit` refines the status once `status_of_exn` has run, but that only happens if the wedge is after `slaughter_kids` -- which is precisely the case the early arming was not added for. Measured, SIGINT during a run:

- un-wedged build: exit 122 (`kid_status + Sys.sigint`)
- wedged build: exit 30 (`incomplete_analysis`)

30 is an ordinary Kind 2 verdict; nothing reading the exit code can tell a hard-killed process from a clean incomplete run. The same path gives worse: `on_exit_success` (`src/kind2Flow.ml:1104-1107`, MCS and MoXI modes) arms with `ExitCodes.success`, and `on_exit_safety_results` arms with the results-derived status, which is 0 when everything proved valid -- so a runtime error plus a wedged teardown exits 0 where an un-wedged run exits 1. Against `main` that is not a regression, since it hung there, but a hang fails loudly in the harness and a 0 passes silently. Worth having the watchdog never report a status that reads as success: keep `ExitCodes.error` as the armed value and let `post_clean_exit` refine it downward, or give the watchdog a code of its own.

## Notes

- The ten seconds now span `slaughter_kids ~exiting:true` (2s deadline plus solver kills) as well, so an orderly `post_clean_exit` has about eight rather than a fresh ten. Fine as measured, just no longer a per-phase budget.
- Single-process modes still never arm it: `on_exit_child` (`src/kind2Flow.ml:520-531`, interpreter and contract monitor) goes straight to `exit` through `on_exit_of_process`. May belong in "What it does not cover".
- Separately confirmed while probing: if standard output blocks *during* the analysis, `on_exit` is never entered at all (instrumented run, stderr completely empty, main thread parked in `anon_pipe_write` for 60s). No watchdog armed on the exit path can cover that; it belongs with the #1449 windows rather than here.

Nothing blocking. Findings 1 and 2 are both small, local changes to the watchdog body.

daniel-larraz:
Both findings fixed in dd60bfef, and both reproduced first so the fix has something to answer.

## 1. `flush_all` blocking the watchdog

Confirmed. A stand-in for the watchdog body, main thread wedged writing into a pipe nobody drains:

| flush | outcome |
|---|---|
| as it was | `watchdog: firing`, then still there when killed at 12s — exit 124 |
| bounded | reaches `_exit`, exit 42 |

Your diagnosis of the one-in-ten in #1449 looks right, and it is the same shape as the third thing you found while probing.

The steps that write now go on a thread of their own with a second to use, and the process exits whether they finish or not. Two consequences worth naming:

- The sweep now precedes the message as well as the flush. It is bounded where a write is not, so it should not sit behind one; and since it is what frees the reader, doing it first gives the message its best chance of landing rather than a worse one.
- What the guarantee now says is that Kind 2 goes, not that the tail of its output arrives. That is the right way round — the reader is freed by Kind 2 going — but it does mean the last second of buffered standard output can be lost on this path, where before the intent was to save it and the effect was to hang.

## 2. The status reading as a normal verdict

Confirmed, with the same wedge you used, at `--timeout 3` through the timeout path:

| build | exit |
|---|---|
| 50db9e37, wedged | **30** — an ordinary incomplete-analysis verdict |
| dd60bfef, wedged | **1** |

Both announce themselves at ten seconds; only the code differs.

I took your first suggestion and one half of the second. `arm_exit_watchdog` no longer takes a status, so the cell keeps `ExitCodes.error` until `post_clean_exit` has been through `status_of_exn`; and the verdict is never lowered to success, since a run that had to be killed did not finish, whatever it had proved by then, and the output saying so is the output that was cut short.

I did not give the watchdog a code of its own: it would be a new number in the documented interface for a case that is already covered by "error", and the message on standard error says which error it is. Say the word if you would rather have the distinct code.

## Your notes

- **Ten seconds now spans `slaughter_kids` too** — agreed, and deliberate: the point of arming early is that the budget covers the teardown as a whole rather than each phase separately. Measured, the orderly path uses a fraction of it.
- **Single-process modes never arm it** — right, `on_exit_child` goes straight to `exit`. Left alone here and named in the commit message: those modes have no engines to abandon and no supervisor to wedge, so the case for a last resort there is different, and I would rather not extend this PR into it on speculation.
- **Standard output blocking during the analysis** — agreed that no watchdog on the exit path can cover it. It is now measured and filed: on Windows the wall clock is only looked at once per iteration of the supervisor loop, and an iteration runs 6-8s ordinarily and 42-55s when a counterexample arrives, which is the whole of #1449. Details there.

Suites after the change: 482 regression tests in 35.3s, ounit OK, no solvers left behind, strict profile clean.

PR #1454, opened 2026-08-25
Support functions with container-typed arguments in IC3IA

daniel-larraz:
### Problem

Since #1424, functions with container-typed (set, map, or sized array) arguments are kept deterministic by ground congruence instances that the engines assert lazily. IC3IA did not assert them in its solvers, so its counterexamples for such systems could be spurious with respect to the intended semantics, and the engine turned these systems down wholesale:

```
IC3IA disabled for property P: system requires functional congruence
instances for functions with container-typed arguments.
```

The exclusion note left in `IC3IA.ml` laid out what support requires — the instances in the transition queries, the instances in the concrete unrolling, and an implicit abstraction that accounts for the difference witness functions they contain. This PR implements exactly that.

### Approach

**Transition queries: instances asserted eagerly.** `sys_def` conjoins the instances for the two concrete offsets of the query (`TransSys.fn_congruence_instances` at bounds 0–1, bumped), under the same activation literal as the system definition. Each instance is a valid formula of the intended semantics, so the over-approximation the frames rest on is preserved: every intended trace lifts to an abstract path satisfying them, UNSAT answers stay sound, and the invariants IC3IA reports remain sound for the engines that receive them (their solvers already declare the witness functions, which travel with the function symbols).

**Counterexample validation: the lazy refinement of the other engines.** In `refine`, a satisfiable unrolling stands only if the model satisfies determinism. The conjunction of the instances over the unrolling's bounds is evaluated first — one term, hence one `get-value` — and when it holds the counterexample is genuine and is reported with nothing asserted at all. When it fails, the instances are evaluated one by one and exactly the violated ones are added before looking again; the loop terminates because each round rules out the model that drove it and the instance set is finite.

A violated instance is conjoined into the interpolizer of the last bound it mentions rather than asserted on its own, so the interpolation partition keeps covering every assertion — which MathSAT's interpolation groups and SMTInterpol's named terms both need.

**Implicit abstraction: single-state predicates only.** An instance spans two bounds, so once instances are in the query an interpolant can span two bounds as well. Interpolant atoms are filtered to those over the cut point's own bound before becoming predicates. When that leaves nothing new — a spurious path ruled out by determinism alone, across bounds no single-state predicate relates — the engine declines with a warning instead of looping on the path forever:

```
IC3IA disabled: the functional congruence instances of a function with
container-typed arguments leave no single-state predicate to refine the
abstraction with.
```

That class is real: determinism at distance ≥ 2 with a different argument in between is beyond single-state predicate abstraction, and k-induction proves it (at k=3 in the new `declined` test, which pins that IC3IA must not call the property falsifiable). Solver traces confirm both the violated-instance rounds and the bail-out fire on that model.

### Results

- `two_phase_commit.lus` with `--enable IC3IA --smt_itp_solver SMTInterpol`: previously refused outright; now IC3IA alone proves SafetyAC1, SafetyAC4, and three of the checks within a 120s budget, the rest unknown, with no spurious answers.
- Determinism properties over containers that IC3IA can now settle: `imported fn deterministic over map` from `success/choose_container_determinism.lus` (valid, k=1), and both properties of the new `ic3ia_fn_container_determinism.lus` — a recurring `int^2` array and two structurally equal but differently built arrays — under MathSAT interpolation, matching the pinned configuration of the `ic3ia/` test directory.
- Genuine counterexamples still surface: the vacuity probe of `falsifiable/choose_container_determinism_sound.lus` is reported invalid, as is a fresh probe over a system with congruence groups whose model satisfies determinism.
- MathSAT still declines set/map models at refinement time — it does not support the Bool-element arrays they compile to; that is the existing arrays limitation (#1445), and the message already points at SMTInterpol.
- Regression suite: 485 passed (483 + the two tests added here); strict-profile build clean.

### DIFF
--- a/src/ic3ia/IC3IA.ml
+++ b/src/ic3ia/IC3IA.ml
@@ -134,7 +134,17 @@ let sys_def solver sys offset =
         Term.mk_not init_term_prev
       ]
    in
-   Term.mk_and (global_constraints sys @
+   (* Ground functional congruence instances over the two concrete offsets of
+      the transition query, enforcing determinism of (abstracted) functions
+      with container-typed arguments within the one-step window. Each is a
+      valid formula of the intended semantics, so conjoining them preserves
+      the over-approximation the frames rest on
+      (see [TransSys.fn_congruence_instances]). *)
+   let congruence =
+     TransSys.fn_congruence_instances sys Numeral.one
+     |> List.map (Term.bump_state (Numeral.pred offset))
+   in
+   Term.mk_and (global_constraints sys @ congruence @
      [Term.mk_or [init; trans]])
 
 let sys_def_unrolling solver sys offset =
@@ -404,27 +414,65 @@ let refine fwd solver sys predicates cubes =
 
   in
 
-  let interpolizers =
+  (* Ground functional congruence instances over the bounds of the concrete
+     unrolling; empty when no function of the system has container-typed
+     arguments. They are not asserted up front: a satisfiable answer stands
+     only if it satisfies all of them, so their conjunction is evaluated in
+     the model first, and exactly the instances the model violates are added
+     to the query before looking again
+     (see [TransSys.fn_congruence_instances]). *)
+  let congruence =
+    TransSys.fn_congruence_instances sys (Numeral.of_int (len - 2))
+  in
+
+  (* The interpolation partition must cover every assertion, so an instance
+     is conjoined into the interpolizer of the last bound it mentions rather
+     than asserted on its own. Sitting there, it lets an interpolant at an
+     earlier cut point mention the instance's earlier bound too; such
+     two-state atoms cannot serve as abstraction predicates and are filtered
+     out below. *)
+  let interpolizer_index inst =
+    match Term.var_offsets_of_term inst with
+    | _, Some mx -> Numeral.to_int mx
+    | _, None -> 0
+  in
+
+  let mk_interpolizers asserted =
+    let extras = Array.make len [] in
+    List.iter
+      (fun i ->
+        let g = interpolizer_index i in
+        extras.(g) <- i :: extras.(g))
+      asserted ;
     cubes
     |> List.mapi (fun i c ->
       let cube = Cube.to_term c |> Term.bump_state (Numeral.of_int (i - 1)) in
-      if i < (len - 1) then
-        Term.mk_and
-          [ cube; sys_def_unrolling intrpo sys (Numeral.of_int i) ]
-      else
-        cube
-      (* If prop is not in the set of initial predicates: *)
-      (*Term.mk_and
-        [ Cube.to_term c |> Term.bump_state (Numeral.of_int (i - 1));
-          if i < (len - 1) then
-            sys_def_unrolling intrpo sys (Numeral.of_int i)
-          else
-            Term.mk_not (prop |> Term.bump_state (Numeral.of_int (i - 1)))
-        ]*)
+      let base =
+        if i < (len - 1) then
+          Term.mk_and
+            [ cube; sys_def_unrolling intrpo sys (Numeral.of_int i) ]
+        else
+          cube
+        (* If prop is not in the set of initial predicates: *)
+        (*Term.mk_and
+          [ Cube.to_term c |> Term.bump_state (Numeral.of_int (i - 1));
+            if i < (len - 1) then
+              sys_def_unrolling intrpo sys (Numeral.of_int i)
+            else
+              Term.mk_not (prop |> Term.bump_state (Numeral.of_int (i - 1)))
+          ]*)
+      in
+      match extras.(i) with
+      | [] -> base
+      | l -> Term.mk_and (base :: l)
     )
   in
 
-  SMTSolver.push intrpo;  
+  let rec solve asserted =
+
+  let interpolizers = mk_interpolizers asserted in
+
+  SMTSolver.push intrpo;
 
   (* Compute the interpolants *)
   let names =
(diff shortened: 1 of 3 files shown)


tests/regression/success/ic3ia/declined/ic3ia_fn_congruence_recurrence.lus
```diff
@@ -0,0 +1,19 @@
+-- A valid property that rests on determinism of a function over a container
+-- across two steps, with a different argument in between. The congruence
+-- instance that rules the spurious counterexample out relates two distant
+-- bounds, and no single-state predicate can stand in for it, so IC3IA's
+-- implicit abstraction cannot be refined for it: the engine turns the
+-- property down (and k-induction proves it at k=3). IC3IA must not claim it
+-- is falsifiable: reporting `unknown` is the expected outcome here.
```
daniel-larraz, line 7: **This test is vacuous under the harness config — it never reaches the branch it is meant to cover.**

`conftest.py:101` pins `--smt_itp_solver MathSAT` for everything under an `ic3ia` path, and `set<Color>` compiles to a Bool-element array that MathSAT rejects. The run therefore exits `0` with

> the interpolating solver does not support the arrays this system uses

and never reaches the new `UnsupportedFeature` branch in `refine`. Since `ic3ia_declined_codes` accepts `success`, the test passes for entirely the wrong reason, and would keep passing if the new branch were deleted.

I confirmed this by running it both ways: with the pinned MathSAT it exits `0` on the array message, and only with `--smt_itp_solver SMTInterpol` does it exercise the intended path (exit `30`, "…leave no single-state predicate to refine the abstraction with").

Either give this test a per-test solver override in `conftest.py`, or write the model with an element type that does not compile to a Bool-element array, so the assertion is actually about the new code.

src/ic3ia/IC3IA.ml
```diff
@@ -515,9 +590,41 @@ let refine fwd solver sys predicates cubes =
...
+    let atoms =
+      if asserted = [] then atoms
+      else
+        TS.filter
+          (fun t ->
+            match Term.var_offsets_of_term t with
+            | Some lo, Some hi ->
+              Numeral.(equal lo zero) && Numeral.(equal hi zero)
+            | _ -> true)
+          atoms
```
daniel-larraz, line 611: **The filter discards atoms that would still be usable predicates, which can turn a refinable property into a declined one.**

Keeping only offset-`(0,0)` atoms also drops atoms that are single-state at an *earlier* bound — offsets `(-1,-1)`. Those are perfectly good single-state predicates; they just need `Term.bump_state Numeral.one` to be expressed over the cut point's bound, exactly the normalization already applied elsewhere. As written they are thrown away.

The cost is not merely a missed predicate: if the earlier-bound atoms are the *only* atoms an interpolant yields, `predicates'` comes back unchanged and the no-progress check below (line 615) fires, so IC3IA raises `UnsupportedFeature` and declines a property it could still have refined.

By the partitioning argument this is the only leak that can occur — `I_i` can never mention a bound *later* than `i` — so handling `(-1,-1)` by bumping rather than dropping covers it:

```ocaml
| Some lo, Some hi when Numeral.(equal lo hi) ->
  (* single-state: keep, bumped to the cut point's bound *)
```

daniel-larraz (commented):
Reviewed the whole diff (`src/ic3ia/IC3IA.ml` +168/-37 and the two new `.lus` tests), read `transSys.ml{,i}`, `lustreTransSys.ml`, `SMTSolver.ml` and the existing congruence pattern in `base.ml`/`step.ml`/`certifChecker.ml`, and ran the new tests against a local build. Two findings, both inline.

What I checked and found sound:

- **`sys_def` bounds.** `fn_congruence_instances sys 1` bumped by `pred offset` lands exactly on the transition query's `(1,2)`, inside the declared `0..3`. I suspected the instances sitting *outside* the `init`/`trans` disjunction could over-constrain the phantom pre-state in the `init` branch — it cannot: every blocked cube carries the `¬%init` literal, so all frame clauses are satisfied when `%init` is true and the phantom stays free (`x1 = x2` is always an available witness).
- **`refine` bounds.** `len - 2` matches the concrete unrolling (cube `i` at `i-1`, `sys_def_unrolling i`, cex path over `0..len-2`), and `len < 2` degrades to an empty instance list, so there is no negative-index or `Array.make` hazard. `interpolizer_index` is provably in `0..len-2`, so `extras` is never out of bounds.
- **Interpolation partitioning.** Placing an instance in the partition of its *max* bound means partitions `0..i` still mention only bounds `≤ i`, so `I_i` can never leak a *later* bound and the `-i` bump stays correct. Only *earlier* bounds leak in — which is what the second finding is about.
- **push/pop balance** across the `solve` recursion; declarations happen before each `push` so they survive the `pop`. `get_term_values` returns the original term rather than the solver's echo, so `asserted @ violated` re-asserts exactly the right instances, and the loop terminates because an instance asserted in a partition cannot be violated again.
- **QE interpolators.** `cvc5_QE`/`Z3_QE` are already excluded for these systems by `check_system_is_supported`/`check_arrays_are_supported`, so the witness UFs never reach quantifier elimination. They are registered in the system's `ufs` and declared by `define_and_declare_of_bounds`.
- **`UnsupportedFeature`** is caught in `kind2Flow.ml:274` and downgraded to a warning; the solver leak on that path is pre-existing (lines 386/401 already did it).
- **Regression.** `regression/success/ic3ia` passes (9 tests), as do the container/determinism-adjacent tests, and a hand-written falsifiable container-function model produces a determinism-consistent counterexample.

src/ic3ia/IC3IA.ml
daniel-larraz: Done in 00615415. The filter now keeps every atom whose variables sit at a single bound, bumped by `-lo` to the cut point's bound (only two-bound atoms are dropped, and bound-free atoms are kept as they are). Since an interpolant can never mention a later bound than its cut, that covers the leak you describe.

Measured the cost on `two_phase_commit.lus` under SMTInterpol, 120s single-engine budget, alternating builds twice: both sides prove the same five properties at the same times (SafetyAC4 ~22–28s, SafetyAC1 ~70–90s), so the extra predicates do not show up in the running time.

tests/regression/success/ic3ia/declined/ic3ia_fn_congruence_recurrence.lus
daniel-larraz: Good catch — done in 00615415. The model now uses an `int^2` array, which MathSAT accepts, so under the pinned configuration the run goes through the refinement rounds and exits 30 through the "leave no single-state predicate to refine the abstraction with" warning (checked with both the pinned MathSAT and SMTInterpol). The comment in the test says why the element type is an integer. Deleting the new branch would now turn this run into a non-terminating refinement loop rather than a pass.

PR #1458, opened 2026-08-26
Run the engines the machine can carry, and no more

daniel-larraz:
Fixes the Windows CI failures of #1449, and the reason behind them.

## What goes wrong

An analysis starts an engine per enabled module, each a domain of its own, and nothing asks how many the machine has. The default is ten. On a four-core Windows runner that does not run slowly, it collapses:

- 95% of the wall clock spent making no progress
- 227s against a `--timeout` of 20s, and runs killed by the harness at 204s
- **the run fails to reach a counterexample that BMC alone finds in one second** — it loses answers, not just time

## Why, measured

It is not the machine running short of CPU. During a stall Kind 2 holds all four cores across ~29 threads while the solvers it is waiting on accrue nothing:

```
stall of 44.2s: Kind 2 165.33s CPU (3.7 cores) across 29 threads,
                its 32 solvers 2.64s (0.1 cores)
```

Threads that are all runnable and producing no solver work are spinning, not working. Domains that cannot all be scheduled cannot all reach a stop-the-world rendezvous, and the ones that are scheduled spin holding the cores the missing ones need.

Sweeping the limit on that runner puts the threshold exactly where that account predicts — where the engines plus the supervisor stop fitting on the cores:

| engines | 3 | 4 | 5 | 6 | unbounded |
|---|---|---|---|---|---|
| stalled | **0%, 0%** | 0%, 58% | 33%, 26% | 75%, 59% | 96%, 78% |

That is why the count is the runtime's own, less one for the supervisor. It is a ceiling and not a target: where the engines already fit, which is every machine of any size, nothing changes.

## Why the limit applies by default on Windows only

Limiting is not free: forced to three where a machine did not need it, the regression suite takes 211s instead of 35s and one test falls from proved to timed out. The runners CI uses are small enough that an unconditional limit would engage on all three platforms, so this matters.

It is not about size. The same unlimited run on the runners themselves, three times each:

| platform | CPUs | share of each run stalled | exit | wall |
|---|---|---|---|---|
| macOS arm64 | **3** | 0%, 0%, 0% | 40 | 5s, 5s, 6s |
| Linux | 4 | 0%, 0%, 0% | 40 | 5s, 4s, 4s |
| Windows | 4 | **92%, 51%, 36%** | **30** | 61s, 31s, 25s |

macOS has *fewer* cores than Windows and does not stall at all, and both it and Linux reach the counterexample uncapped in seconds — which is exactly what a limit would take from them. What differs is the scheduler: where the machine is shared proportionally, the domain that is late to the rendezvous arrives and the others merely wait.

So the default limits on Windows and nowhere else. `--max_engines` is available on every platform for anyone who wants the limit anyway, on a shared machine or a small one.

## Which engines go, when they must

The order of `--enable` is not the order to give them up in. Two rules, both about what an engine is worth when there is no room for it.

**An engine that decides properties by itself outranks one that helps another engine decide them.** So the shortest prefix is BMC, k-induction and IC3QE, and IND2 — which earns its keep from invariants another engine generates — goes after the first generator rather than third.

**An invariant generator works over one kind of value, and there is one of them per kind.** A system with no reals in it gets nothing whatever from the generator over reals. So ask the inferred logic which kinds the system uses — the way `process_invgen_mach_modules` already asks it about machine integers — and let the ones with nothing to generate go first. Otherwise which generator survives is settled by the order of `--enable`, which knows nothing about the system.

`--max_engines` overrides the count for anyone whose machine is not what the runtime thinks.

## Validation

| | without | with |
|---|---|---|
| runs hung and killed | 3 of 6 | **0 of 6** |
| stall share | 76–98% | **0% every run** |
| wall clock | 28s–285s | **1–2s** |
| exit code | 30, or killed | **40 — the counterexample** |

(Windows runner, six runs each way.)

- The limit chosen on that runner was `Running 3 of 9 engines at once`.
- Given a system over integers the generator over reals goes first, and given one over reals the generator over integers does. Before this change both dropped the same one.
- On a twenty-core machine the limit does not engage: 482 regression tests unchanged, 36.2s against 35.8s. Forced to three there they still all pass, in 211s, with one test falling from proved to timed out — the cost of the limit where the machine did not need it, and the reason the default suits the machine rather than being a number.
- The strict profile builds clean.

## Worth knowing

- This is a mitigation for a livelock in the runtime's rendezvous under oversubscription, not a cure for it. It keeps Kind 2 below the point where its own domains fight each other.
- The workaround in #1451 can come out once this lands, and #1449 closed with it.
- Not addressed here, and arguably a bug of its own: a system with no integers still *runs* an integer invariant generator when there is room for it. This change orders those last under a limit but does not stop them running.

### DIFF
--- a/src/flags.ml
+++ b/src/flags.ml
@@ -3354,7 +3354,10 @@ module Global = struct
 
      Zero asks the runtime what the machine can carry, which is what
      nearly every run should use: it is a ceiling rather than a target,
-     so it changes nothing where the engines already fit. *)
+     so it changes nothing where the engines already fit. It only asks
+     on Windows, since that is the only platform measured to collapse
+     rather than merely slow down, and limiting costs proofs. Give a
+     number to have the limit anywhere. *)
   let max_engines_default = 0
   let max_engines = ref max_engines_default
   let _ = add_specs ([
@@ -3363,7 +3366,8 @@ module Global = struct
       fun fmt ->
         Format.fprintf fmt
           "\
-            Maximum number of engines running at once (0 to suit the machine)@ \
+            Maximum number of engines running at once@ \
+            0 suits the machine, which limits on Windows and nowhere else@ \
             Engines beyond the limit are not run, dropping the last of@ \
             \"--enable\" first, so the ones that decide properties are kept@ \
             Default: %d\
--- a/src/flags.ml
+++ b/src/flags.ml
@@ -3342,6 +3342,37 @@ module Global = struct
   ])
   let timeout_analysis () = !timeout_analysis
 
+
+  (* How many engines may run at once.
+
+     Every engine is a domain of its own and holds solver processes of
+     its own, so an analysis of the default ten engines puts some forty
+     runnable things on the machine. Where that is well beyond what the
+     machine has, they do not share it, they thrash: on a four core
+     runner the process was descheduled for 95% of the wall clock and
+     failed to reach a counterexample that two engines find in a second.
+
+     Zero asks the runtime what the machine can carry, which is what
+     nearly every run should use: it is a ceiling rather than a target,
+     so it changes nothing where the engines already fit. *)
+  let max_engines_default = 0
+  let max_engines = ref max_engines_default
+  let _ = add_specs ([
+    ( "--max_engines",
+      Arg.Set_int max_engines,
+      fun fmt ->
+        Format.fprintf fmt
+          "\
+            Maximum number of engines running at once (0 to suit the machine)@ \
+            Engines beyond the limit are not run, dropping the last of@ \
+            \"--enable\" first, so the ones that decide properties are kept@ \
+            Default: %d\
+          "
+          max_engines_default
+    )
+  ])
+  let max_engines () = !max_engines
+
   (* Only parse mode. *)
   let only_parse_default = false
   let only_parse = ref only_parse_default
@@ -3877,6 +3908,7 @@ let input_format = Global.input_format
 let real_precision = Global.real_precision
 let timeout_wall = Global.timeout_wall
 let timeout_analysis = Global.timeout_analysis
+let max_engines = Global.max_engines
 let input_file = Global.input_file
 let all_input_files = Global.get_all_input_files
 let clear_input_files = Global.clear_input_files
--- a/src/flags.mli
+++ b/src/flags.mli
@@ -215,6 +215,9 @@ val timeout_wall : unit -> float
 (** Per-run wallclock timeout. *)
 val timeout_analysis : unit -> float
 
+(** Maximum number of engines to run at once, 0 to suit the machine. *)
+val max_engines : unit -> int
+
 (** The Kind modules enabled is a list of [kind_module]s. *)
 type enable = Lib.kind_module list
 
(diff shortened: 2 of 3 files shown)


daniel-larraz:
**Hold this — the default is probably too blunt, and the reason is a good one.**

@daniel-larraz points out that the Windows benchmarks were healthy before MathSAT arrived, and that without a limit Windows performs well and only *occasionally* runs past its timeout. That fits the mechanism here rather than contradicting it, and it undercuts the shape of this fix.

The livelock needs enough domains busy in OCaml **at once** to outnumber the cores. Before #1445, IC3IA on Windows ran under Z3qe, which disables itself on array systems (`IC3IA (Z3qe) disabled: arrays are not supported`) and does less work elsewhere, so fewer domains were ever busy together. MathSAT made IC3IA genuinely work and pushed the count past the threshold.

So the trigger is not that ten engines exist; it is how many are busy together. A fixed limit pays on every run to prevent an occasional crossing, and on the Windows benchmarks it would cut nine engines to three permanently. Measured cost of that where a machine did not need it: the regression suite takes 211s instead of 35s and one test falls from proved to timed out.

Two better shapes, both of which keep every engine:

1. **Fewer rendezvous rather than fewer domains.** The cost is roughly how often the domains must meet times what a meeting costs; the limit attacks the second factor, a larger engine minor heap attacks the first. Each engine also calls `Gc.set` at spawn, and each of those is itself a stop-the-world, so there are ten of them during startup while other engines are already running. Being measured now.
2. **Shed only on detection.** Keep every engine and have the supervisor watch its own loop latency, dropping engines only once it is actually stalling. Full parallelism in the good case, degradation only when the pathology appears.

What is worth keeping from this PR regardless: `--max_engines` as a flag, the priority order for giving engines up, and the domain-aware ranking of the invariant generators. What is in question is only whether anything should be limited *by default*.

The CI workaround in #1451 already keeps the job green, so there is no pressure to land a default in a hurry.

daniel-larraz:
Closing in favour of #1459, which fixes the failure this was aimed at without the price this one charges.

This works — on the four core runner it took stalls from 76-98% to 0% and returned the counterexample six runs out of six, where the unlimited runs never found it. But it buys that by running three engines of nine, on every Windows run, to prevent an occasional overrun. @daniel-larraz's point stands: the Windows benchmarks were healthy, and a limit that always applies taxes every model to protect the few that stall.

What the investigation turned up in the meantime is that the limit was aimed at the wrong thing. The failure in CI is not that Windows is slow, it is that a run can go minutes past the `--timeout` it was given, because on Windows the wall clock is only looked at in the supervisor's polling loop and the supervisor stops running. The process does not stop — it holds every core, spinning in the runtime — so a thread the operating system made can still end it on time. #1459 does that in four files and touches no engines: measured, a run given 20s ends at 85s instead of 250s, and a healthy run never meets the backstop at all.

Not lost, if anyone wants them later:

- The priority order for giving engines up, and the domain-aware ranking of invariant generators — an invariant generator works over one kind of value, so a system with no reals gets nothing from the one over reals, and today `--enable` order decides which survives.
- The separate observation that a system with no integers still *runs* an integer invariant generator when there is room. That is arguably a bug of its own and belongs in its own issue.
- `--max_engines` itself is not needed: `--enable` already lets anyone pick a subset.

The slowness remains real and is still #1449. #1459 only stops it being unbounded.

PR #1459, opened 2026-08-26
Let the wall clock have the last word on Windows

daniel-larraz:
Makes `--timeout` mean something on Windows. Fixes the CI failures of #1449 without touching which engines run.

## The problem

Kind 2 looks at its wall clock in the polling loop of the supervisor, and on Windows that is the only place it looks — there is no SIGALRM, and `Unix.setitimer` raises there. That is enough while the supervisor runs.

It stops running. Where more engines are busy than the machine has cores, the domains cannot all reach a stop-the-world rendezvous, so every one of them parks in it and **no OCaml code executes anywhere in the process** for seconds at a stretch — a thread of the supervisor's domain and a domain of its own were both measured stopped throughout. Runs given `--timeout 20` have been killed by the harness still running at 243s, 251s, 269s and 270s.

The last resort added in #1450 does not save these runs. It is an OCaml thread, so it needs the runtime lock that nothing can release, and across twelve runs here it never once fired: of the four that hung, none wrote its line and none exited. That is not an argument against it — it does its job on the paths where OCaml still runs — it is the reason this one cannot be written in OCaml.

## Why a thread can still do it

The process is not stopped. It holds all four cores throughout, spinning inside the runtime:

```
stall of 44.2s: Kind 2 165.33s CPU (3.7 cores) across 29 threads,
                its 32 solvers 2.64s (0.1 cores)
```

What is stopped is *OCaml*. A thread the operating system made, which never takes the master lock and calls nothing in the runtime, is scheduled as usual. So the last word on the wall clock goes to one of those: it sleeps, says what it is doing on standard error, takes the solvers with it through a job object, and ends the process with `ExitCodes.incomplete_analysis` — the status a wall clock timeout produces on every other platform.

## How long to wait: sixty

It is a backstop and not the timeout. Windows has the ordinary path too — the polling loop notices the clock, raises `TimeoutWall`, and the teardown reports what was proved — and that path works on every run whose supervisor is running, which is every healthy one. The grace is how long to wait for it before concluding it will not come, and firing early forfeits a verdict the run already had.

It was thirty, on the reasoning that a healthy run finishes about ten seconds past its timeout and three times that is safe. **CI has since produced 29.** The same test, `quant_finite_nested_binders.lus [slice_on]`, on the Windows runner across three commits, every one of them a healthy run reporting what it proved:

| job | elapsed | past its 84s budget |
|---|---|---|
| `main` 6ac334cb | 95.5s | +11.5s |
| `5fd366cb` | 88.2s | +4.2s |
| `77cd7e3a` | 113.0s | **+29.0s** |

A grace of thirty would have killed that third run had it been one second slower, and reported an incomplete analysis in place of a verdict it already had. The multiple was the wrong shape of argument: the ordinary path waits out the same stalls this exists because of, so how late it can be has no bound to take a multiple of, and the run that most needs the grace is the stalled one that is most likely to exhaust it. Sixty, from the measurements.

Nothing is at risk from the wait itself: at the 84s the harness asks for, this bounds a wedged run at 144s inside a budget of 204s.

## Validation

Four core runner, `--timeout 20`, with the grace at thirty as it then was, so the backstop falls due at 50s. Twelve runs each way, over four jobs:

| | wall clock | exit |
|---|---|---|
| without | 26s, 28s, 31s, 31s, 38s, 54s, 72s, 88s, 144s, **killed still running at 243s, 251s, 269s, 270s** | 30 where it finished, — where it did not |
| with | 23s, 31s, 33s, 33s, 37s, 39s, 40s, 53s, 55s, 56s, 58s, **63s** | **30 throughout** |

The runs that come in under the due time never met the backstop, which is the point: it does nothing to a healthy run. Raising the grace to sixty moves the due time and nothing else.

**In this pull request's own CI, not only in a throwaway harness.** `kind2-build (windows-x86_64)` is green on every commit here, with #1451's workaround gone. In the job on `5fd366cb`, `test-issue-127.lus` took **117.9s** (22:51:55.255 → 22:53:53.179) against a backstop then due at 114s, and was one of the two entries under "Kind 2 ran out of time" — the grace plus a few seconds, not the ordinary path. Both that run and an ordinary one exit 30, and until now nothing in the report told them apart, so this branch adds the elapsed seconds to that summary:

```
1 test(s) exited 30 after the 84s budget (not counted as failures):
   113.0s  regression/success/quant_finite_nested_binders.lus::quant_finite_nested_binders [slice_on]
```

## Punctuality, and the thread that has to wake up

The backstop is due at a known second, and it cannot speak before it has been scheduled. Twenty runs over two jobs at a grace of five, so that it fires on runs that would otherwise finish first, with the raise on the sleeping thread switched off in one arm:

| | fired | slack |
|---|---|---|
| raised | 8 of 10 | **+0.0s, eight of eight** |
| ordinary priority | 8 of 10 | +0.0s, +1.3s, +2.4s, +2.7s, **+4.6s, +4.8s, +5.7s, +6.0s** |

Every raised run that met the backstop was ended at the second it was due; seven of the eight ordinary ones were late, by up to six seconds. So the raise is what makes it punctual, and the 63s in the table above — thirteen seconds past a nominal fifty — was measured before the thread that sleeps was raised. Ordinary priority is not enough on a machine whose every core is held by domains spinning at a rendezvous: `Sleep` returning is not the same as the thread running. It is raised at its creation site, since it cannot raise itself after waking — making that call is already the part that needs a slice — and it costs nothing, being asleep for the whole run.

The runs that did not fire came in under the due time on their own, except for two in the second job which ended at about thirty seconds without the line appearing. That is consistent with the backstop firing on time and the write being starved through its five second bound, and equally with the run finishing by itself just after — both exit 30, and the harness cannot tell them apart. What it does bound either way is the process: the write never becomes the reason it stays alive.

## The job holds the solvers

Counting the solvers left behind afterwards proves nothing: an idle solver reads the end of its standard input and goes whether or not anything killed it. Asking the job what it holds, at the moment the backstop fires, does:

```
Kind 2 ran past its timeout without stopping, terminating it and the
31 solver processes its job holds.
```

Counts of 7 to 36 across the runs that carried one — the low ones from late-firing runs whose solvers had already begun to go, which is the point of asking at the moment it fires rather than counting what is left afterwards. The query is `JobObjectBasicAccountingInformation`, a fixed size structure, so a buffer too small for the answer is not among the ways it can fail, and a job that holds nothing else says so:

```
Kind 2 ran past its timeout without stopping, terminating it. Its job
holds no other process.
```

which one run produced, so the plain line now means only that the query failed.

`dune build --profile strict` clean; 486 regression tests pass in 35.6s here and in 647.6s on the Windows runner. These are the first C stubs in the tree; outside Windows the file compiles to a stub that does nothing, since SIGALRM already does this properly.

## What it does not do

- **It does not make Windows faster.** A heavy model still takes minutes there; why is #1449. This only stops those minutes being unbounded.
- **It is a hard stop**, so whatever the run had computed is lost. That is the same trade the last resort in `post_clean_exit` already makes, and the alternative is a run that never stops. The grace is the width of the window in which that trade can go wrong, which is why it is set from measurements rather than from a multiple.
- **It is not a substitute for the OCaml watchdog.** That one runs everywhere and can still exit cleanly; this one only knows how to stop the process.

## Bearing on the other work

- **Supersedes #1458** (now closed), which limited how many engines run at once. That works — 0% stalls, right answers, 6/6 — but it costs proofs on every Windows run to prevent an occasional overrun, which is a bad trade for the benchmarks. Closing it in favour of this.
- **Reverts #1451 here**, the workaround the harness carried for this. It accepted that one test being killed after the budget on Windows rather than failing. **The Windows job of this pull request is therefore the evidence**, and it is green.
- #1449 stays open. The slowness is real and undiagnosed beyond "the domains livelock at the rendezvous when they outnumber the cores".

### DIFF
--- a/scripts/gen-linking-flags.sh
+++ b/scripts/gen-linking-flags.sh
@@ -18,7 +18,12 @@ case "$LINKING_MODE" in
                 FLAGS="-noautolink"
                 NAT=$(echo $OCAML_VERSION | awk -F. '{print ($1 <= 4 || ($1 == 5 && $2 == 0)) ? "" : "nat"}')
                 PTHREAD=$(echo $OCAML_VERSION | awk -F. '{print ($1 <= 4) ? "-lpthread" : ""}')
-                CCLIB="-lthreadsnat -lunix$NAT -lcamlstr$NAT -lnums $PTHREAD";;
+                # -noautolink means every C stub archive has to be named
+                # here. kind2dev_stubs is ours; the rest belong to the
+                # libraries it uses. A stub added to `foreign_stubs`
+                # without a line here fails only in a static build, which
+                # no pull request runs.
+                CCLIB="-lkind2dev_stubs -lthreadsnat -lunix$NAT -lcamlstr$NAT -lnums $PTHREAD";;
             *)
                 echo "No known static compilation flags for '$OS'" >&2
                 exit 1
--- a/src/dune
+++ b/src/dune
@@ -21,6 +21,7 @@
 
 (library
  (name kind2dev)
+ (foreign_stubs (language c) (names kind2_stubs))
  (libraries dune-build-info num str threads yojson menhirLib)
  (modules
   (:standard \ C2Icnf kind2 moxi))
--- a/src/kind2.ml
+++ b/src/kind2.ml
@@ -136,6 +136,25 @@ let setup : unit -> any_input = fun () ->
   (* Set sigalrm handler. *)
   Signals.set_sigalrm_timeout_from_flag () ;
 
+  (* On Windows there is no SIGALRM, and the wall clock is looked at
+     only in the polling loop of the supervisor. That is enough while
+     the supervisor runs, and on a machine with fewer cores than there
+     are busy engines it stops running: every domain parks in a
+     stop-the-world rendezvous none of them can complete, no OCaml code
+     executes anywhere, and the run goes minutes past the time it was
+     given. Measured on a four core runner, a run given 20s took 227s.
+
+     So the last word on the wall clock there belongs to a thread the
+     operating system schedules, which the runtime cannot stop. It is a
+     backstop and not the timeout: a generous grace on top, so that it
+     only ever fires on a run that has already failed to stop itself. *)
+  ( if Sys.win32 then
+      match Flags.timeout_wall () with
+      | timeout when timeout > 0. ->
+        NativeTimeout.arm
+          (int_of_float timeout + 60) ExitCodes.incomplete_analysis
+      | _ -> () ) ;
+
   (* Install generic signal handlers for other signals. *)
   Signals.set_sigint () ;
   (* SIGPIPE is ignored, not turned into an exception: signal handlers
--- a/src/kind2.ml
+++ b/src/kind2.ml
@@ -167,8 +167,19 @@ let setup : unit -> any_input = fun () ->
   ( if Sys.win32 then
       match Flags.timeout_wall () with
       | timeout when timeout > 0. ->
-        NativeTimeout.arm
-          (int_of_float timeout + 30) ExitCodes.incomplete_analysis
+        (* A count of seconds the other side can hold. It reaches a
+           thirty-two bit count of milliseconds, and it comes from a
+           float the user chose: [--timeout inf] converts to zero, which
+           would arm the backstop at the grace alone and kill a run that
+           asked for no limit, and a large enough value wraps the
+           multiplication instead. Anything past the ceiling is a run
+           that means to go on indefinitely, and the backstop simply
+           does not arm for it. *)
+        let ceiling = 1_000_000. in
+        if Float.is_nan timeout || timeout > ceiling then ()
+        else
+          NativeTimeout.arm
+            (int_of_float timeout + 30) ExitCodes.incomplete_analysis
       | _ -> () ) ;
 
   (* Install generic signal handlers for other signals. *)
--- a/src/kind2.ml
+++ b/src/kind2.ml
@@ -145,14 +145,30 @@ let setup : unit -> any_input = fun () ->
      given. Measured on a four core runner, a run given 20s took 227s.
 
      So the last word on the wall clock there belongs to a thread the
-     operating system schedules, which the runtime cannot stop. It is a
-     backstop and not the timeout: a generous grace on top, so that it
-     only ever fires on a run that has already failed to stop itself. *)
+     operating system schedules, which the runtime cannot stop.
+
+     It is a backstop and not the timeout. Windows has the ordinary path
+     too -- the polling loop notices the clock and raises [TimeoutWall],
+     which unwinds through the exit path and reports what was proved,
+     the same as SIGALRM does elsewhere -- and that path works on every
+     run whose supervisor is running, which is every healthy one. The
+     grace is how long to wait for it before concluding it will not
+     come. Measured on a four core runner, a healthy run finishes about
+     ten seconds past its timeout: the polling loop looks at the clock
+     once an iteration, and the teardown prints what it found. Thirty is
+     three times that.
+
+     Not less, and never zero. This ends the process rather than
+     unwinding it, so it forfeits the results and forces its own status
+     -- fire it while an orderly teardown is still running and a run
+     that had disproved a property would report an incomplete analysis
+     instead. The grace is the width of that window, and it is worth
+     being generous with. *)
   ( if Sys.win32 then
       match Flags.timeout_wall () with
       | timeout when timeout > 0. ->
         NativeTimeout.arm
-          (int_of_float timeout + 60) ExitCodes.incomplete_analysis
+          (int_of_float timeout + 30) ExitCodes.incomplete_analysis
       | _ -> () ) ;
 
   (* Install generic signal handlers for other signals. *)
--- a/src/kind2.ml
+++ b/src/kind2.ml
@@ -153,10 +153,18 @@ let setup : unit -> any_input = fun () ->
      the same as SIGALRM does elsewhere -- and that path works on every
      run whose supervisor is running, which is every healthy one. The
      grace is how long to wait for it before concluding it will not
-     come. Measured on a four core runner, a healthy run finishes about
-     ten seconds past its timeout: the polling loop looks at the clock
-     once an iteration, and the teardown prints what it found. Thirty is
-     three times that.
+     come.
+
+     Sixty, from what CI measures rather than from a multiple. The
+     ordinary path is not quick and steady: it waits out the same
+     stalls the backstop exists because of, so how late it is has no
+     bound to take a multiple of. One test on the Windows runner came
+     in 4.2s past its timeout on one commit, 11.5s on another and 29.0s
+     on a third, all healthy runs reporting what they had proved. A
+     grace of thirty would have killed the third one had it been a
+     second slower, taking its verdict with it -- and the run that most
+     needs the grace is exactly the stalled one, which is the run most
+     likely to exhaust it.
 
      Not less, and never zero. This ends the process rather than
      unwinding it, so it forfeits the results and forces its own status
@@ -179,7 +187,7 @@ let setup : unit -> any_input = fun () ->
         if Float.is_nan timeout || timeout > ceiling then ()
         else
           NativeTimeout.arm
-            (int_of_float timeout + 30) ExitCodes.incomplete_analysis
+            (int_of_float timeout + 60) ExitCodes.incomplete_analysis
       | _ -> () ) ;
 
   (* Install generic signal handlers for other signals. *)
--- a/src/kind2_stubs.c
+++ b/src/kind2_stubs.c
@@ -59,32 +59,27 @@ static LONG kind2_timeout_solvers = -1;
    behind afterwards is not -- an idle solver reads the end of its
    standard input and goes whether or not anything killed it.
 
-   Room for one id is enough. The number of assigned processes comes
-   back in the header even when the buffer cannot hold the list, which
-   is the ERROR_MORE_DATA case, and the ids themselves are of no use
-   here. */
+   The accounting information is a fixed size structure, so there is no
+   buffer that can be too small and nothing to distinguish a short
+   answer from a failed one: -1 here means the query failed and zero
+   means the job really did hold nothing else. */
 static LONG kind2_solvers_in_job(void)
 {
-  JOBOBJECT_BASIC_PROCESS_ID_LIST ids;
+  JOBOBJECT_BASIC_ACCOUNTING_INFORMATION acc;
   DWORD returned = 0;
 
   if (kind2_timeout_job == NULL) return -1;
-  ZeroMemory(&ids, sizeof ids);
+  ZeroMemory(&acc, sizeof acc);
   if (!QueryInformationJobObject(kind2_timeout_job,
-                                 JobObjectBasicProcessIdList,
-                                 &ids, sizeof ids, &returned)
-      && GetLastError() != ERROR_MORE_DATA)
+                                 JobObjectBasicAccountingInformation,
+                                 &acc, sizeof acc, &returned))
     return -1;
-  if (ids.NumberOfAssignedProcesses == 0) return -1;
-  return (LONG) ids.NumberOfAssignedProcesses - 1;   /* less this one */
+  /* Cannot happen while this process is in the job, so treat it as a
+     query that did not answer rather than as an answer of none. */
+  if (acc.ActiveProcesses == 0) return -1;
+  return (LONG) acc.ActiveProcesses - 1;     /* less this one */
 }
 
-/* Say what is happening, from a thread of its own.
-
-   A write to the standard error of Kind 2 blocks rather than failing
-   when it is a pipe whose reader has stopped, and this is the path that
-   must not be stoppable. The OCaml last resort bounds its own writes
-   for the same reason. */
 static DWORD WINAPI kind2_timeout_announce(LPVOID unused)
 {
   (void)unused;
@@ -93,6 +88,10 @@ static DWORD WINAPI kind2_timeout_announce(LPVOID unused)
             "Kind 2 ran past its timeout without stopping, terminating it "
             "and the %ld solver processes its job holds.\n",
             kind2_timeout_solvers);
+  else if (kind2_timeout_solvers == 0)
+    fprintf(stderr,
+            "Kind 2 ran past its timeout without stopping, terminating it. "
+            "Its job holds no other process.\n");
   else
     fprintf(stderr,
             "Kind 2 ran past its timeout without stopping, terminating.\n");
--- a/src/kind2_stubs.c
+++ b/src/kind2_stubs.c
@@ -47,16 +47,48 @@
 
 static DWORD kind2_timeout_ms = 0;
 static UINT kind2_timeout_status = 0;
+static HANDLE kind2_timeout_job = NULL;
 
-static DWORD WINAPI kind2_timeout_thread(LPVOID unused)
+/* Say what is happening, from a thread of its own.
+
+   A write to the standard error of Kind 2 blocks rather than failing
+   when it is a pipe whose reader has stopped, and this is the path that
+   must not be stoppable. The OCaml last resort bounds its own writes
+   for the same reason. */
+static DWORD WINAPI kind2_timeout_announce(LPVOID unused)
 {
   (void)unused;
-  Sleep(kind2_timeout_ms);
-  /* Nothing below enters the runtime, so a rendezvous holding every
-     domain does not hold this. */
   fprintf(stderr,
           "Kind 2 ran past its timeout without stopping, terminating.\n");
   fflush(stderr);
+  return 0;
+}
+
+static DWORD WINAPI kind2_timeout_thread(LPVOID unused)
+{
+  HANDLE announce;
+  (void)unused;
+  Sleep(kind2_timeout_ms);
+
+  /* Nothing below enters the runtime, so a rendezvous holding every
+     domain does not hold this. */
+  announce = CreateThread(NULL, 0, kind2_timeout_announce, NULL, 0, NULL);
+  if (announce != NULL) {
+    WaitForSingleObject(announce, 1000);
+    CloseHandle(announce);
+  }
+
+  /* The job first, when there is one: it holds this process and the
+     solvers it started, and ends them together. Ending only this
+     process would leave a solver that is inside a query -- one that is
+     idle reads the end of its standard input and goes, but one that is
+     working does not read it until the query returns, and that is not
+     bounded. Every other exit path kills the solvers for that reason.
+
+     `TerminateJobObject` does not return, so the fallback below runs
+     only if there is no job to end. */
+  if (kind2_timeout_job != NULL)
+    TerminateJobObject(kind2_timeout_job, kind2_timeout_status);
   TerminateProcess(GetCurrentProcess(), kind2_timeout_status);
   return 0;
 }
@@ -67,6 +99,23 @@ CAMLprim value kind2_arm_native_timeout(value seconds, value status)
   HANDLE thread;
   kind2_timeout_ms = (DWORD) (Long_val(seconds) * 1000);
   kind2_timeout_status = (UINT) Long_val(status);
+
+  /* A job this process and its children belong to, so that the thread
+     can end them together. A child joins the job of its parent unless
+     it asks not to, and the solvers do not ask. Nesting has been
+     allowed since Windows 8, so being in a job already -- which a CI
+     runner arranges -- is not a reason this fails.
+
+     The handle stays open for the life of the process, and the job is
+     deliberately not created with JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE:
+     closing it must not be what ends the run. */
+  kind2_timeout_job = CreateJobObject(NULL, NULL);
+  if (kind2_timeout_job != NULL
+      && !AssignProcessToJobObject(kind2_timeout_job, GetCurrentProcess())) {
+    CloseHandle(kind2_timeout_job);
+    kind2_timeout_job = NULL;   /* fall back to ending this process alone */
+  }
+
   thread = CreateThread(NULL, 0, kind2_timeout_thread, NULL, 0, NULL);
   if (thread != NULL) CloseHandle(thread);
   CAMLreturn(Val_unit);
--- a/src/kind2_stubs.c
+++ b/src/kind2_stubs.c
@@ -74,7 +74,15 @@ static DWORD WINAPI kind2_timeout_thread(LPVOID unused)
      domain does not hold this. */
   announce = CreateThread(NULL, 0, kind2_timeout_announce, NULL, 0, NULL);
   if (announce != NULL) {
-    WaitForSingleObject(announce, 1000);
+    /* Above the rest, and given several seconds. This fires on a
+       machine whose every core is held by domains spinning at a
+       rendezvous, and a thread of ordinary priority given a second did
+       not get to run at all: the process was ended on time and said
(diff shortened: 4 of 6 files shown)


daniel-larraz:
Reviewed at `618349ad`. Built with `--profile strict` (clean), ran the regression suite (486 passed, 35.7s), and mutation-tested the claims below.

## Findings

### 1. The macOS release build will not link — `scripts/gen-linking-flags.sh:18-21`

The macOS static path sets `-noautolink` and then enumerates by hand every C stub archive the binary needs (`-lthreadsnat -lunix$NAT -lcamlstr$NAT -lnums`). `foreign_stubs` in `src/dune:24` adds a new one, `libkind2dev_stubs.a`, and it is not on that list.

Reproduced by adding `-noautolink` plus that same `-cclib` list to the `kind2` executable stanza:

```
/usr/bin/ld: src/kind2dev.a(nativeTimeout.o): in function `camlNativeTimeout':
:(.data+0x18): undefined reference to `kind2_arm_native_timeout'
```

Adding `-cclib -lkind2dev_stubs` links clean. Both executables need it and both take their flags from that file, so one entry in the `macosx` `CCLIB` covers `kind2` and `kmoxi`.

This pull request's CI will not catch it: `build-target: static` only appears in `.github/workflows/kind2-release.yml:132`, which does not run on pull requests. It would surface at the next nightly or tag.

### 2. The backstop is the one exit path that leaves the solvers — `src/kind2_stubs.c:60`

`post_clean_exit` calls `SMTSolver.destroy_all_of_process ()` (`src/kind2Flow.ml:533`), and the last resort calls it *first*, before anything that writes (`src/kind2Flow.ml:495`), on the stated grounds that the solvers are children and killing this process does not kill them. `TerminateProcess` does neither.

Measured how much that costs, by SIGKILLing Kind 2 on this test with 30 solvers live: all 30 were gone within a second — an idle solver reads EOF on its stdin and exits, which is what #1447 bought. But a solver inside a query does not read stdin until the query returns. A `z3 -smt2 -in` given a query taking 3.4s outlived its stdin EOF by the full 3.4s; in general that is unbounded, which is why `destroy_all_of_process` kills rather than asks.

On the stall this targets the solvers are near-idle (the description measures 0.1 cores across 32), so the common case is covered by EOF. The uncovered case is the other reason a run overruns — a solver genuinely churning — and there the backstop leaves it behind with nothing to notice it. The stub cannot call into the runtime to enumerate solvers, but it does not have to: assign the process to a job object at arm time (children join automatically) and have the thread call `TerminateJobObject`, which takes Kind 2 and every solver it started, atomically, without touching OCaml.

### 3. The stderr write can defeat the termination it announces — `src/kind2_stubs.c:57-60`

`fprintf`, then `fflush`, then `TerminateProcess`, in that order, with nothing bounding the first two. The OCaml last resort wraps every write on that path in `within 1.0` (`src/kind2Flow.ml:466`, `src/kind2Flow.ml:500`) precisely because "the standard output of Kind 2 may be a pipe whose reader has stalled, and a write to it blocks rather than failing." The same is true of stderr here, and this is the mechanism that is supposed to be undefeatable. The harness drains concurrently via `communicate`, so CI is safe; a user piping Kind 2 into something that stops reading is not. Writing from a helper thread and terminating after a bounded wait keeps the message without making it a precondition.

### 4. The description describes a grace the head commit no longer ships

It says "Sixty seconds of grace", puts the backstop at 80s for `--timeout 20`, and quotes a bound of 144s inside the harness budget. `618349ad` made it thirty: 50s and 114s. The whole validation table was measured at 60. Worth updating before merge, since the body becomes the commit message.

On the change itself: 30 is well-argued and I think right. The orderly teardown is not the unbounded part — it is already capped by the watchdog's `Thread.delay 10.0` at `src/kind2Flow.ml:481` — so the real question is how late the polling loop notices the clock, and the cost of guessing low is quiet: a run terminated mid-teardown reports 30, and `tests/conftest.py:318-321` accepts 30 for any non-IC3IA test, so a forfeited `falsifiable` verdict passes without a trace. Note that at 60 one of the three validation runs still hit the backstop. That is an argument for landing 30 and watching, not against it, but it is worth naming that no run was measured at 30.

### 5. Nit: an unvalidated float reaching a `DWORD` — `src/kind2.ml:171`, `src/kind2_stubs.c:68`

`--timeout inf` passes `timeout > 0.`, and `int_of_float infinity` is `0` (measured), so the backstop arms at 30 seconds on a run that asked for no limit. `--timeout 4294968` wraps the millisecond product to 704ms. `--timeout 1e18` is signed overflow in `Long_val(seconds) * 1000`. Nobody types these, and `inf` already fails loudly on POSIX (`setitimer` raises `EINVAL`), but on Windows this turns it into a silent kill. A clamp at the OCaml call site closes all three.

## Verified

- `dune build --profile strict` clean; `cd tests && ./run` 486 passed in 35.7s, no stray solvers left behind.
- The `tests/conftest.py` change is an exact revert of ba3c488c (`git diff ba3c488c^ HEAD -- tests/conftest.py` is empty); `run_timeout`, `LustreTimeout` and every import it touched are still used.
- `timeout_wall` defaults to `0.`, so nothing arms unless `--timeout` is given; the arm sits within microseconds of `Stat.start_timer Stat.total_time`, so its clock and the polling loop's agree.
- I could not build the `_WIN32` branch — no mingw available — so findings 2, 3 and 5 are read from the source and from the POSIX analogues I could measure.

daniel-larraz:
Re-verified at `60fa76c8`.

## Fixed

**1. The macOS static link — confirmed fixed.** `scripts/gen-linking-flags.sh` now emits `-lkind2dev_stubs` at the head of the list. Re-ran the reproduction with the exact flags the script produces (`bash scripts/gen-linking-flags.sh static macosx 5.2.0`), on both executables:

```
dune build src/kind2.exe src/moxi.exe   ->   EXIT 0
```

The same test against `618349ad` failed with `undefined reference to kind2_arm_native_timeout`. The archive lands after the OCaml objects that reference it on the link line, so naming it first in `CCLIB` is fine.

**2. The solvers — the job object is right.** Created at arm time, before any solver exists, so children join by themselves; a failed `AssignProcessToJobObject` closes the handle and falls back to ending this process alone; no `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`, so a normal exit closing the handle changes nothing. `TerminateJobObject` followed by an unconditional `TerminateProcess` covers both the no-job case and the case where the terminate itself fails, so there is no path where the thread reaches the end having done nothing. Nesting on Windows 8 and later is correct, and the runner's own job does not block it. I could not execute this here — no mingw — so it rests on reading plus the Windows job.

**3. What it writes — bounded.** The line goes out on its own thread with a second, and the process ends whether or not it lands. The bound is now `--timeout` plus thirty, plus up to a second.

**5. The ceiling — correct.** `inf` and `1e18` do not arm; `1000000` arms at 1000030s, which is 1,000,030,000 milliseconds, under 2^32 and not `INFINITE`; `nan` never gets past the outer `timeout > 0.` guard.

## Still outstanding

**4. The description is unchanged.** It still says "Sixty seconds of grace", puts the backstop at 80s for `--timeout 20`, quotes the 85s run, and gives the bound as 144s. Flagging it only so it does not merge that way.

## One correction to the commit message of 60fa76c8

> `inf` and `nan` are refused by the flag parser before they reach here, so the check on them is defensive.

They are not refused. `Arg.Set_float` takes whatever `float_of_string` accepts. Measured on this build:

```
--timeout inf   -> Fatal error: exception Unix.Unix_error(Unix.EINVAL, "setitimer", "")   exit 2
--timeout nan   -> runs to completion normally                                            exit 0
--timeout 1e18  -> runs to completion normally                                            exit 0
```

`inf` only dies on POSIX because it reaches `setitimer`, and `src/terms/termLib.ml:481` skips `setitimer` on Windows — so there `inf` reaches the arm site with nothing else in the way. The ceiling check is load-bearing on Windows rather than defensive, which is a better story than the one the message tells. (`--timeout inf` crashing on POSIX is pre-existing and not this pull request's business.)

## Verified

- `dune build --profile strict` clean; `cd tests && ./run` 486 passed in 37.7s; no stray solvers afterwards.
- `kind2-build (windows-x86_64)` on `60fa76c8` was still running when I wrote this; every other build job passed. `618349ad`'s Windows job passed, so the earlier stub compiled there — the job object APIs are new surface that this run is the first to cover, and it is also the evidence for the reverted harness workaround.

daniel-larraz:
Findings 1, 2, 3 and 5 were fixed in `60fa76c8`; finding 4 is now done, and re-verification turned up a mistake of mine that is worth recording because it changes what one of those fixes rests on.

## 4. The description — rewritten

At grace 30 throughout: backstop due at 50s for `--timeout 20`, 114s against the harness's budget of 204s, and a validation table measured at 30 rather than the old one measured at 60. It also now says what the earlier body did not: **the bound is 50s nominal and 63s observed**, because the backstop's thread is an ordinary-priority native thread on a machine whose cores are all held by spinning domains, so `Sleep` returning is not the same as it running. Bounded, not punctual.

## The mistake

I reported after the first re-verification that `Kind 2 ran past its timeout without stopping, terminating.` appeared **zero** times, concluded that the one-second bound from finding 3 had swallowed it, and changed the announce thread to `THREAD_PRIORITY_HIGHEST` with a five-second bound to fix that.

The zero was my harness. It captured every line Kind 2 wrote and then printed only the ones mentioning a stall — the message was never printed *by construction*, so its absence was evidence about the Python and not about this branch. The same filter produced the other zero I reported, for `did not exit in time` across five earlier runs.

Fixed the harness to print both lines with the time they were written, or say explicitly that they were never written. Re-ran:

```
[native-timeout 2] exit 30 after 53s
    said at 51.7s: Kind 2 ran past its timeout without stopping, terminating.
    solvers left behind: 0
```

So the message does land. Whether it would have landed at one second is now untested, and the priority change is a precaution rather than a repair — kept, since a starved thread plausibly needs more than a second and five seconds against a thirty-second grace costs nothing.

## What the corrected harness did show

The OCaml watchdog of #1450 **never fires in this failure mode**. Across nine runs without the backstop, the three that hung wrote neither line and none exited; it needs the runtime lock that nothing can release. That is the case for this pull request stated more sharply than the original description managed, and it is now in the body.

## 3, restated

The line still goes out on its own thread and the process still ends whether or not it lands. Bound is five seconds, not one. Worst case is `--timeout` plus thirty, plus the scheduling slack above, plus up to five.

## Related

The same one-second bound sits in the OCaml watchdog on `main`. #1462 raises it to five there, on the strength of the bound's measured behaviour rather than of any Windows observation — that path is exactly the one shown above never to run on Windows. It is independent of this pull request.

## Verified

- `dune build --profile strict` clean; `cd tests && ./run` 486 passed in 36.1s.
- Nine Windows runs each way across three jobs; every run with the backstop exited 30, none over 63s; solvers left behind 0 in all eighteen. That last number does not discriminate — idle solvers exit on EOF anyway — so the job object remains reasoned rather than demonstrated, as it was at the first review.

daniel-larraz:
Reviewed `5fd366cb` and the thread. The one-second bound I asked for was too tight — the measurement settled that, and the fix is right. All checks are green on `5fd366cb`, Windows included.

## The main thing

**The priority raise went on the announce thread, not on the thread that has to wake up — `src/kind2_stubs.c:84` against `src/kind2_stubs.c:127`.**

The description now has to advertise "50s nominal and 63s observed", and attributes the gap to exactly this: "a native thread of ordinary priority on a machine whose cores are all held by spinning domains, so `Sleep` returning is not the same as it running." That is a description of `kind2_timeout_thread`, which is still created at ordinary priority. It pays the slack twice — once waking from `Sleep`, then again after `WaitForSingleObject` returns and it has to reach `TerminateJobObject`.

The same one-line call that fixed the announce thread applies, at the creation site in `arm` where the handle is in hand:

```c
thread = CreateThread(NULL, 0, kind2_timeout_thread, NULL, 0, NULL);
if (thread != NULL) {
  SetThreadPriority(thread, THREAD_PRIORITY_HIGHEST);
  CloseHandle(thread);
}
```

It costs nothing — the thread is asleep for the whole run — and it is the difference between a bound of 63s and one near 50s. Setting it from inside the thread after `Sleep` returns would not work, since making that call is itself what needs a slice.

## The Windows job is already better evidence than the description claims

`kind2-build (windows-x86_64)` passed on `5fd366cb`: 486 passed in 650.71s, no failures, with #1451's workaround gone. And the backstop visibly fired in it. From the job log timestamps, `test-issue-127.lus` took **117.9s** (22:51:55.25 to 22:53:53.18) against a backstop due at 84 + 30 = 114s, and it is one of the two entries under "Kind 2 ran out of time". Thirty-four seconds past its timeout is the grace plus about four seconds of slack, not the ordinary path.

For contrast in the same log, `quant_finite_nested_binders.lus` took 88.3s — 84 plus a normal teardown — which is the ordinary path. That one also exits 30 on `main`'s Windows job (run `33016237489`), so it is pre-existing and nothing to do with this branch. `test-issue-127` finished normally on that main run, which is the intermittency the whole thing is about.

So this pull request's own CI, and not only the throwaway harness, shows a stalled run bounded at 114s and landing inside the 204s budget. Worth citing in the body — it is the evidence the description says the Windows job would be.

## Two smaller things

**The backstop is invisible in CI, and will stay that way after merge.** The harness prints a run's output only on failure, so `Kind 2 ran past its timeout without stopping, terminating.` never reaches the log, and the summary at `tests/conftest.py:368-377` cannot distinguish a run the polling loop ended from one the backstop killed — I had to do it by subtracting log timestamps. Adding the elapsed seconds to each line of that list would separate them at a glance, and would still be there once the hunt branches are deleted.

**A discriminating test for the job object exists and is cheap.** You are right that "solvers left behind: 0" proves nothing, since idle solvers exit on EOF anyway. `QueryInformationJobObject(job, JobObjectBasicProcessIdList, ...)` on the timeout thread, with the count folded into the announce line, would show the solvers are in the job at the moment it fires. It touches nothing in the runtime, and it turns the mechanism from reasoned into demonstrated.

**Nit:** `CreateThread` followed by `SetThreadPriority` leaves a window in which the announce thread can start at ordinary priority. `CREATE_SUSPENDED` plus `ResumeThread` closes it. Unlikely to matter on a machine where it would not be scheduled in that window anyway.

## Verified

- `dune build --profile strict` clean; `cd tests && ./run` 486 passed in 36.1s; no stray solvers afterwards.
- All checks green on `5fd366cb`, including `kind2-build (windows-x86_64)` at 18m6s and `macos-x86_64` and `macos-arm64`, which is the first CI confirmation of the linking fix on a real macOS build.

daniel-larraz:
All four addressed in `77cd7e3a`. The main one was a real defect and the mechanism given for it was right, so it got tested rather than taken on trust.

## The priority raise, on the thread that sleeps

Moved to the creation site in `arm`, where the handle is already in hand. It cannot raise itself after waking, since making that call is already the part that needs a slice.

The claim attached to it is causal — ordinary priority is *why* the bound sat thirteen seconds past nominal — so I ran an arm with the raise switched off. Both arms at a grace of five rather than thirty, so the backstop fires on every run instead of the occasional stalled one, with the line timestamped against the second it was due:

| | fired at | slack |
|---|---|---|
| raised | 25.0s × 5 | **+0.0s, five of five** |
| ordinary priority | 25.0s, 29.6s, 30.7s | +0.0s, **+4.6s, +5.7s** |

Confirmed. The two probe switches (`KIND2_PROBE_GRACE`, `KIND2_PROBE_NO_PRIORITY`) live only on the throwaway branch and never merge.

## The job object, demonstrated

`QueryInformationJobObject` on the timeout thread, before anything is ended, folded into the line:

```
Kind 2 ran past its timeout without stopping, terminating it and the
31 solver processes its job holds.
```

Eight of those ten runs carried a count, of 29 to 33. Two printed the plain line — the fallback when the query gives nothing usable — and I have not established which failure mode that was. The count is diagnostic and best-effort; the process is ended either way. Room for one id is enough, as you said: `NumberOfAssignedProcesses` comes back in the header on `ERROR_MORE_DATA`.

## The nit

Taken. `CREATE_SUSPENDED`, raise, `ResumeThread`.

## Elapsed seconds in the summary

Added at `tests/conftest.py:377`. Linux never produces a timed-out run, so I exercised the path by running the suite at a one-second budget:

```
2 test(s) exited 30 after the 1s budget (not counted as failures):
     1.0s  regression/success/bv_compression.lus::bv_compression [slice_on]
```

It earned its keep immediately. The Windows job on `77cd7e3a` reports:

```
1 test(s) exited 30 after the 84s budget (not counted as failures):
   113.0s  regression/success/quant_finite_nested_binders.lus [slice_on]
```

113.0s against a backstop due at 114s — a run that stalled and came back one second before it would have been killed, which the old summary would have shown as indistinguishable from the 88.3s ordinary path.

## Your CI reading, checked

Both numbers reproduce from the job log on `5fd366cb`: `test-issue-127` 22:51:55.255 → 22:53:53.179 = **117.9s** against 114s due, `quant_finite_nested_binders` 22:57:37.193 → 22:59:05.439 = **88.3s**. Cited in the body now, and the validation table is twelve runs each way rather than nine.

## One thing worth flagging about the verification itself

The re-run immediately before this one was **inconclusive and I nearly reported it as support**. All three runs with the backstop finished on their own, under the due time, so it never fired — no evidence either way about the priority or the count. The controlled arms above exist because of that. It is the same failure as the filtered harness earlier in this thread: a run in which the mechanism never fires says nothing about the mechanism.

## Verified

- `dune build --profile strict` clean; `cd tests && ./run` 486 passed in 35.7s.
- All checks green on `77cd7e3a`, Windows included — 486 passed in 647.6s with #1451's workaround gone.
- No Windows cross-compiler here, so CI remains the first compile of the C changes.

daniel-larraz:
Reviewed the updated description against `77cd7e3a` and the three Windows jobs. **Nothing below is blocking** — the code is correct as far as I can check it, and every check is green with #1451's workaround gone, which is the bar this pull request set for itself.

The body is much stronger. The punctuality experiment is the right shape — both arms, the raise switched off in one — and the job count is real evidence. Two things it now contains that it does not say, and one it says that does not add up.

## The grace has about a second of margin, and the reason given for thirty no longer holds

The body quotes this from the `77cd7e3a` job:

```
113.0s  regression/success/quant_finite_nested_binders.lus [slice_on]
```

That number is the interesting one, and the body files it under the new summary format rather than reading it. **113.0s is not the backstop.** The backstop is due at 84 + 30 = 114s measured from `arm`, which is a few tenths after the harness starts its clock, so anything it kills lands at 114.1s or later. 113.0s is the ordinary path, arriving 29.0s past the budget — about a second short of being killed mid-teardown.

The same test across the three jobs, from the log timestamps of the neighbouring test lines:

| job | elapsed | past budget |
|---|---|---|
| `main` 6ac334cb | 95.5s | +11.5s |
| `5fd366cb` | 88.2s | +4.2s |
| `77cd7e3a` | 113.0s | **+29.0s** |

`src/kind2.ml:156-159` and `618349ad` justify thirty as "a healthy run finishes about ten seconds past its timeout ... Thirty is three times that." Real CI has now produced 29. I endorsed 30 on that reasoning; the new instrumentation in this branch is what changed it.

Nothing was lost this time — that test exits 30 on `main` too, so the forfeit the code warns about had no verdict to take. But the grace has to wait out the whole OCaml path, and that path is subject to the same stalls it exists to survive, so its lateness has no bound to be three times of. 45 or 60 would bound the run at 129s or 144s, still comfortably inside the harness's 204s. Either raise it, or restate the rationale on the measured numbers rather than on three times ten. Not blocking: the exposure is a run that stalls past its timeout *and* had a verdict to report, which is the trade the body already documents and accepts.

## The message cannot tell "no solvers" from "could not tell"

The body says of the two runs that printed the plain line: "I have not established which of its failure modes that was." There is a third possibility that is not a failure mode. `kind2_solvers_in_job` returns `-1` for unknown and `0` for a job that legitimately held only Kind 2, and `src/kind2_stubs.c:91` tests `> 0`, so both print the plain line. Changing that to `>= 0` would have separated them in the same run — "the 0 solver processes its job holds" is an answer, and the plain line then means only that the query failed.

Which also makes the earlier suggestion less cosmetic than I pitched it: `JobObjectBasicAccountingInformation` has a fixed-size struct and an `ActiveProcesses` field, so there is no too-small-buffer case to have been the cause of the two. Between the two changes the open question closes by itself.

## The punctuality table accounts for eight of the ten runs

`hunt-the-windows-hang.yml` runs `rounds` per arm, so ten runs is five and five. The raised row has its five; the ordinary row lists three (25.0s, 29.6s, 30.7s). Either two are missing from that row or the ten is off — and the row that is short is the one carrying the argument.

**Nit:** the quoted summary line drops `::quant_finite_nested_binders` from the node id. It is shown as literal output; the real line is `113.0s  regression/success/quant_finite_nested_binders.lus::quant_finite_nested_binders [slice_on]`.

## Verified

- 113.0s and 647.55s are exactly what the `77cd7e3a` Windows job printed.
- Built `77cd7e3a` in a clean worktree: `dune build --profile strict` clean, 486 passed in 35.9s, and the new summary format reproduces exactly as quoted when the budget is shrunk to 1s.
- All checks green on `77cd7e3a`, including `kind2-build (windows-x86_64)` at 17m20s.

If anything here is worth five minutes before merging, it is the table row — it becomes the merge commit message — and the `>= 0` one-liner, which closes the open question in the body without another run. Everything else can follow.

daniel-larraz:
All four addressed in `21bd1a01`, and the grace one changed the code rather than the wording.

## The grace is sixty

Your three timings reproduce exactly: `main` 22:09:21.731 → 22:10:57.185 = **95.5s**, `5fd366cb` = **88.2s** (I had quoted 88.3), `77cd7e3a` 23:36:59.058 → 23:38:52.102 = **113.0s**. So a healthy run reporting its verdict came within one second of being killed and having that verdict replaced.

You offered raise-it or restate-it. Raised, because the argument you make against the multiple also argues against thirty: the ordinary path waits out the same stalls the backstop exists for, so its lateness has no ceiling to take three times of, and the run that most needs the grace is the stalled one most likely to exhaust it. The comment at `src/kind2.ml:156` now gives the three measurements instead of the multiple. Sixty bounds a wedged run at 144s against the harness's 204s.

## Zero, told from unknown

Took the `JobObjectBasicAccountingInformation` suggestion rather than only relaxing the test, since as you say it removes the short-buffer case from the failure modes. It earned itself immediately — one run in the new job:

```
Kind 2 ran past its timeout without stopping, terminating it. Its job
holds no other process.
```

so the plain line now means only that the query failed. Counts elsewhere ran 7 to 36, the low ones from late-firing runs whose solvers had already begun to go — which is the argument for asking at the moment it fires rather than counting what is left behind.

## The table row

You were right that it was short. Two of the five ordinary runs finished before the backstop was due and never fired. Rather than annotate three samples I ran the whole thing again, so it is twenty runs over two jobs:

| | fired | slack |
|---|---|---|
| raised | 8 of 10 | **+0.0s, eight of eight** |
| ordinary priority | 8 of 10 | +0.0s, +1.3s, +2.4s, +2.7s, **+4.6s, +4.8s, +5.7s, +6.0s** |

Seven of the eight ordinary runs were late, by up to six seconds; every raised one was exact.

**Nit:** fixed, the node id is whole.

## One thing the new job raises that I cannot settle

Two raised runs ended at about thirty seconds with no line at all. That is consistent with the backstop firing on time and the write being starved through its five second bound, and equally consistent with the run finishing by itself just after — both exit 30, and nothing in the log separates them. I am not claiming either. What is not in doubt is that the bound holds: the write never becomes the reason the process stays alive.

If it is worth settling, the cheap discriminator is a probe that gives the backstop a distinct exit status, so the code says which path ended the run. That is another Windows round and nothing here depends on it, so I have left it.

## Verified

- `dune build --profile strict` clean; `cd tests && ./run` 486 passed in 35.6s.
- The counts, the zero line and the punctuality table are from job `33025360219` on the throwaway branch, built from this branch plus the two probe switches.
- CI on `21bd1a01` is running; `77cd7e3a` was green throughout, Windows included.

PR #1460, opened 2026-08-26
Let the last resort report the verdict the run reached

daniel-larraz:
A regression I introduced in #1450, caught by the Windows job of #1459.

## What happens

The last-resort watchdog exits with `ExitCodes.error` unless `post_clean_exit` has already told it the verdict. On Windows it often has not: the teardown there can outlast the watchdog's ten seconds — the stalls of #1449 make `slaughter_kids` slow — so it fires mid-teardown and reports **1**. A run that had completed its analysis, printed `Incomplete analysis result` and reached verdict **30** exits as though it had crashed.

It is not hypothetical. The Windows regression job failed on it:

```
FAILED regression/success/quant_finite_nested_binders.lus::quant_finite_nested_binders [slice_on]
       - Expected: success, got Unknown return code: 1
```

That test ordinarily runs out of its budget and is forgiven for it, since 30 is a timeout rather than a failure. Reported as 1, it is a failure.

And nothing in the output explained why: the line the watchdog prints goes through a bounded helper thread, which a stalled process does not schedule either, so the run exits silently with the wrong code. Both halves of that come from the same commit of mine.

## The fix

Arm with what the results already say. `on_exit` is handed the analysis verdict, and `status_of_exn` leaves it alone for the exceptions that end a run normally — a wall clock timeout among them — so it is already the right answer before `post_clean_exit` confirms it.

**Success still becomes an error**, which is what that rule was for: a run the watchdog had to kill did not finish, and a zero would pass silently. Every other verdict is now the one the run actually reached.

## How it got in

The review of #1450 asked that a killed run not report a status that reads as a normal verdict. What it meant was *success*; I applied it to every status. The measurement was in front of me at the time — I recorded that the wedged run went from exiting 30 to exiting 1, and called that the fix rather than the regression it was.

## Validation

Wedging the teardown past the watchdog's ten seconds:

| wedged run | before | after |
|---|---|---|
| times out (verdict 30) | **1** | **30** |
| falsifies a property (verdict 40) | 1 | **40** |
| would have succeeded (verdict 0) | 1 | **1** — never success |

The 486 regression tests pass in 36.0s and the strict profile builds clean.

## Relation to the other work

Independent of #1459 and of the reverted workaround in #1451, which never covered this test. It has been latent on main since #1450 and shows only when a teardown outlasts ten seconds, which is a Windows symptom of #1449.

### DIFF
--- a/src/kind2Flow.ml
+++ b/src/kind2Flow.ml
@@ -420,15 +420,18 @@ let slaughter_kids ?(exiting = false) process sys =
    only the tail of what can wedge.
 
    The status the watchdog exits with is kept in a cell rather than
-   captured, and it is not lowered before [status_of_exn] has run: until
-   the exception has been accounted for, the only honest thing to report
-   is that Kind 2 had to be terminated. [post_clean_exit] sets the
-   verdict of the run once it knows it.
+   captured, and it carries the best verdict known so far.
+   [post_clean_exit] sets the final one once [status_of_exn] has run;
+   before that, [on_exit] sets what the results already say, which for
+   a run out of time is [incomplete_analysis] and for a falsified one
+   is [unsafe_result]. Those are the answers the run had reached, and
+   reporting them beats reporting an error the run did not have.
 
-   It never reports success. A run the watchdog had to kill did not
-   finish, whatever it had proved by then, and the output saying so is
+   It never reports success, though, whatever the results say: a run
+   the watchdog had to kill did not finish, and the output saying so is
    the output that was cut short. A hang fails loudly wherever Kind 2
-   runs; a zero passes silently. *)
+   runs; a zero passes silently. So success becomes [error], and
+   nothing else is touched. *)
 let watchdog_status = Atomic.make ExitCodes.error
 let watchdog_armed = Atomic.make false
 
@@ -482,9 +485,11 @@ let arm_exit_watchdog () =
       with _ -> () )
 
 (* The verdict of the run, for the watchdog to exit with if it has to.
-   Never success: see [watchdog_status]. *)
+   Success becomes an error, since a killed run did not finish; every
+   other verdict is what the run had reached. See [watchdog_status]. *)
 let watchdog_reports status =
-  if status <> ExitCodes.success then Atomic.set watchdog_status status
+  Atomic.set watchdog_status
+    (if status = ExitCodes.success then ExitCodes.error else status)
 
 (** Called after everything has been cleaned up. *)
 let post_clean_exit process base_status exn =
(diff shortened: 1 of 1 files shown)


daniel-larraz:
Reproduced the validation table on Linux by wedging `slaughter_kids ~exiting:true` with a 15s sleep, and mutation-tested against the pre-PR code. The fix does what it claims, and the reasoning in the comments is accurate:

| wedged run | pre-PR (measured) | with PR (measured) |
|---|---|---|
| times out, verdict 30 | 1 | **30** |
| proven, then `TimeoutWall` (base 0) | 1 | **1** — never success |

486 regression tests pass in 36.4s, ounit clean, the strict profile builds clean, and no stray `z3` is left after any run.

One gap.

## The rule now leaks in the other direction

`src/kind2Flow.ml:531` arms with the raw `status`, but the comment right above it states the precondition without the code enforcing it: `status_of_exn` leaves `status` alone *only* for the exceptions that end a run normally. For `Signal`, `Failure`, `Parser_error` and the catch-all it substitutes its own code — and the supervisor's `| e -> handle_exception …; on_exit_safety_results in_sys \`Supervisor e` at `src/kind2Flow.ml:1307-1309` reaches `on_exit` with exactly those.

So in the watchdog's own window — `on_exit` armed, `slaughter_kids` wedged, `post_clean_exit` not yet reached — a run that ended abnormally now exits with a verdict that reads as normal. Measured, same wedge, same binary:

| wedged run | correct | pre-PR | with PR |
|---|---|---|---|
| runtime failure after a falsification | 1 | 1 | **40** |
| SIGINT with partial results | 122 | 1 | **30** |

In the first case the crash message is cut short as well, so the run exits 40 and says nothing about why. And `tests/conftest.py:345-348` returns early on 30 for every non-IC3IA test, so a crashed-and-wedged run would be counted in `unfinished_runs` and pass. That is the hazard `watchdog_status`'s own comment argues against: a hang fails loudly wherever Kind 2 runs, a zero passes silently. Here 30 passes silently just as surely as 0 does.

The window is narrow — a Windows teardown outlasting ten seconds, per #1449 — but Ctrl-C during a stall is not exotic.

## Suggested fix

Report the results verdict only for the exceptions `status_of_exn` passes through. The reported failure (`| TimeoutWall -> on_exit_safety_results …`, `src/kind2Flow.ml:1306`) is covered exactly, and everything else keeps the conservative `error` default:

```ocaml
  arm_exit_watchdog () ;
  ( match exn with
    | Exit | KEvent.Terminate | TimeoutWall | TimeoutVirtual ->
      watchdog_reports status
    | _ -> () ) ;
```

Built and measured with that guard in place, same wedge:

| wedged run | with guard |
|---|---|
| times out, verdict 30 | **30** — fix preserved |
| proven, then `TimeoutWall` | **1** |
| runtime failure after a falsification | **1** — restored |
| SIGINT with partial results | **1** — restored |

daniel-larraz:
Fixed in 3dc43f7b, and the gap was real — reproduced both cases before fixing them.

## Confirmed

Wedging `slaughter_kids ~exiting:true` and mutating the guard away, to check the guard is what does the work rather than assuming it:

| wedged run | without the guard | with the guard |
|---|---|---|
| times out, verdict 30 | 30 | **30** — the fix preserved |
| injected runtime failure | **30** | **1** |
| SIGINT mid-analysis | **30** | **1** |

Your reading of the mechanism is exactly right, including why it is worse than it looks: a 30 is accepted as a timeout by the harness, so a crashed-and-wedged run would have passed quietly, with its own message cut short by the same stall that caused the wedge.

## What I did differently from the suggestion

Two changes to the shape, same behaviour on the four cases you measured.

**A named predicate beside `status_of_exn` rather than an inline match.** The two have to agree, and a bare `match` at the arming site is a second copy of that function's dispatch with nothing to connect them — the next person to edit `status_of_exn` has no reason to look at `on_exit`. It now reads `if keeps_the_given_status exn then watchdog_reports status`, with the predicate directly under the function whose behaviour it mirrors and a comment saying so.

**Two more constructors.** `SMTSolver.Unknown` and both `UnsupportedFeature`s also return the given status — they log a warning and leave the verdict alone — so they belong with the four you listed. `Failure` stays out even though two of its messages pass through, because which they are depends on the text; the comment records that as a deliberate omission rather than an oversight.

## Method note

Two of my attempts at your SIGINT case measured nothing, and both were my error rather than a real result: first I signalled the `timeout` wrapper instead of Kind 2 (exit 130 is `timeout` dying, not a verdict), then I used a configuration that falsified in a second so the signal arrived after the run was over. The implausible exit code is what gave each away. Worth recording since the same trap is there for anyone re-running these.

486 regression tests pass in 35.6s, strict profile clean, no scaffolding left behind.

PR #1461, opened 2026-08-26
Show a reachability property as it was written

daniel-larraz:
The summary printed the query rather than the property.

Asked to reach `c = 3`, Kind 2 answered about `(not (c = 3))`; asked to reach `c = 5` from step 2, about `((not (c = 5)) or (*counter < 2))` — the negation it proves, and the counter it instruments the system with, neither of which appears in the user's model.

```
--%PROPERTY reachable c = 3;
--%PROPERTY reachable c = 5 from 2;
check y < 50 provided y < 25;      -- with --check_nonvacuity
```

| | before | after |
|---|---|---|
| `reachable c = 3` | `(not (c = 3))` | `(c = 3)` |
| `reachable c = 5 from 2` | `((not (c = 5)) or (*counter < 2))` | `(c = 5)` |
| guard of `provided y < 25` | `(not (y < 25))` | `(y < 25)` |

## Why it happened

A reachability query is checked by asking whether its negation is invariant. `LustreAstNormalizer` builds that negation and then hands the *same* rewritten expression to `record_source_expr`, so the property's recorded source is the thing to prove rather than the thing that was asked.

The fix names the rewritten expression separately and records the annotation's own expression. The bounded forms had it twice over, since their query also carries the instrumentation counter; they now print the expression alone, which is what the bound in the heading already describes. The guard of a `provided` check is a reachability property of the same kind and had the same problem.

## Scope

Only the plain text summary changes. The XML and JSON outputs do not carry the expression, and `expr_source_map` is read in exactly two places, both of which format it for printing — so nothing that reasons about the property is affected, only what is shown.

## Validation

- The three cases above, measured before and after.
- The 486 regression tests pass in 37.2s, and the strict profile builds clean.

### DIFF
--- a/src/lustre/lustreAstNormalizer.ml
+++ b/src/lustre/lustreAstNormalizer.ml
@@ -309,6 +309,18 @@ let record_source_expr gids normalized_expr original_expr =
   let key = HString.mk_hstring (A.string_of_expr normalized_expr) in
   { gids with expr_source_map = StringMap.add key original_expr gids.expr_source_map }
 
+(* What a property was written as, under its own name.
+
+   Not [record_source_expr]: that is keyed by the variable an expression
+   was abstracted to, and a reachability query for [P] normalizes to
+   [not P], which is the same variable as a property written [not P].
+   A property name is its own. *)
+let record_prop_source gids name original_expr =
+  match name with
+  | None -> gids
+  | Some name ->
+    { gids with prop_source_map = StringMap.add name original_expr gids.prop_source_map }
+
 let get_inductive_vars ind_vars expr =
   let vars = AH.vars_without_node_call_ids expr in
   let ind_vars = List.map fst (StringMap.bindings ind_vars) in
@@ -1476,7 +1488,8 @@ and normalize_item info node_id map = function
                               Const (dpos, Num (HString.mk_hstring (string_of_int b)))))
         in
         let nexpr, gids, warnings = abstract_expr false info (Some node_id) map query in
-        let gids = record_source_expr gids nexpr expr in
+        let gids = record_source_expr gids nexpr query in
+        let gids = record_prop_source gids name' expr in
         [A.AnnotProperty (pos, name', nexpr, k)], gids, warnings
 
       (* expr or counter != b *)
@@ -1487,7 +1500,8 @@ and normalize_item info node_id map = function
                               Const (dpos, Num (HString.mk_hstring (string_of_int b)))))
         in
         let nexpr, gids, warnings = abstract_expr false info (Some node_id) map query in
-        let gids = record_source_expr gids nexpr expr in
+        let gids = record_source_expr gids nexpr query in
+        let gids = record_prop_source gids name' expr in
         [AnnotProperty (pos, name', nexpr, k)], gids, warnings
 
       (* expr or counter > b *)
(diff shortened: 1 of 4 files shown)


src/lustre/lustreAstNormalizer.ml
```diff
@@ -1512,13 +1512,13 @@ and normalize_item info node_id map = function
...
+        let nexpr, gids, warnings = abstract_expr false info (Some node_id) map query in
         let gids = record_source_expr gids nexpr expr in
         [AnnotProperty (pos, name', nexpr, k)], gids, warnings
 
       | Reachable _ -> 
-        let expr = A.UnaryOp (pos, A.Not, expr) in
-        let nexpr, gids, warnings = abstract_expr false info (Some node_id) map expr in
+        let query = A.UnaryOp (pos, A.Not, expr) in
+        let nexpr, gids, warnings = abstract_expr false info (Some node_id) map query in
         let gids = record_source_expr gids nexpr expr in
```
daniel-larraz: `expr_source_map` is keyed by the *abstracted variable name*, and `mk_fresh_local`'s `LocalCache` reuses a single variable whenever two items normalize to the same expression. A reachability query for `P` normalizes to `not P`, which collides with any ordinary property or contract item written as `not P`. Before this change both calls stored the same source, so the collision was harmless; now they store different sources under the same key, and whichever runs second wins.

Checked against the built binary:

```
--%PROPERTY not (x > 5);
--%PROPERTY reachable x > 5;
```

The invariant is printed as `(x > 5)` — wrong, and a regression: before this commit it correctly printed `(not (x > 5))`. With the two lines in the opposite order the reachable property prints as `(not (x > 5))`, i.e. the fix silently does not apply. The same collision happens between a contract `guarantee not (x > 5);` and a body `--%PROPERTY reachable x > 5;`.

Two ways out: key the source map per property (e.g. by the generated property name) rather than by abstracted variable, or force a fresh abstraction for reachability queries so they never share a cache entry.

daniel-larraz:
Fixed in 2edb5104. Reproduced your case first — it is exactly as you describe, and the regression is mine.

| | before this PR | with e8c2af93 | now |
|---|---|---|---|
| `not (x > 5)` then `reachable x > 5` | `(not (x > 5))` / `(not (x > 5))` | **`(x > 5)`** / `(x > 5)` | `(not (x > 5))` / `(x > 5)` |
| `reachable x > 5` then `not (x > 5)` | — | `(not (x > 5))` / `(not (x > 5))` | `(x > 5)` / `(not (x > 5))` |
| `guarantee not (x > 5)` with `reachable x > 5` | — | collided | `(not (x > 5))` / `(x > 5)` |

## Which of the two ways out

I took the first — a property's source kept under its own name, with `expr_source_map` left to go on answering about expressions.

**The second would not have been enough on its own.** `mk_fresh_local` adds to `local_cache` even when `force` is set:

```ocaml
  | Some nexpr, false -> nexpr, empty ()
  | _ -> ... LocalCache.add local_cache expr nexpr; nexpr, gids
```

So a forced reachability query still leaves `not P -> V` in the cache, and the next unforced property or contract item written `not P` picks up that same variable — the collision returns, just in the other order. It would have fixed the case you reported and left its mirror image.

## What changed

`prop_source_map` on `GeneratedIdentifiers`, keyed by property name, written by `record_prop_source` in the reachability branches and in the `provided` guard, and preferred by `lustreNodeGen` when it has a name. `record_source_expr` goes back to recording the query, so every non-reachability path is byte-for-byte as it was before this PR.

Property names are unique, so there is nothing left for it to collide with.

## Validation

Both orders of the collision, the contract variant, the original reachability cases and the `provided` guard — all measured, all in the commit message. 486 regression tests pass in 35.6s and the strict profile builds clean.

Thank you for catching it: recording a different value under a shared key is the kind of thing that looks right in the diff and only shows when two features meet.

PR #1471, opened 2026-08-27
Let the logic decide which arithmetic invariant generators run

daniel-larraz:
`process_invgen_mach_modules` asks the system what it is made of before it spawns anything. It reads the inferred logic and turns `INVGENMACH` into one generator per bit width actually present — or, if there are no bit vectors at all, into nothing:

```ocaml
match TransSys.get_logic sys with
| `Inferred fs when mem BV fs -> (* one generator per width in the system *)
| _ -> other_modules
```

Nobody ever asked the same question of the integer and real generators. They are spawned on every analysis whenever they are enabled, and `INVGENINTOS` and `INVGENREALOS` are enabled by default. So a system with no integer in it runs an integer invariant generator.

Both halves of the contrast are visible in one run. `tests/regression/falsifiable/bv-lia-ex.lus` is a `uint8` counter and an `int8` counter, and nothing else:

| | invariant generators spawned |
| --- | --- |
| before | bool ×2, bv8, ubv8, **int**, **real** |
| after | bool ×2, bv8, ubv8 |

The machine-integer path had already narrowed itself to the two widths in the file. The int and real generators sat next to it with nothing to mine.

## The change

```ocaml
let process_invgen_arith_modules: TSys.t -> kind_module list -> kind_module list
= fun sys modules ->
  match TSys.get_logic sys with
  | `Inferred fs -> (
    let open TermLib.FeatureSet in
    modules |> List.filter (
      function
      | `INVGENINT | `INVGENINTOS -> mem IA fs
      | `INVGENREAL | `INVGENREALOS -> mem RA fs
      | _ -> true
    )
  )
  | _ -> modules
```

applied in `analyze` right after the machine-integer pass. `IA` covers `int`, subranges and enums, so those keep the integer generator; `RA` covers `real`. When the logic was not inferred — `--smt_logic` was given, so `get_logic` returns `` `SMTLogic `` or `` `None `` — the modules are kept: having no feature set to consult is not a reason to drop an engine.

## Why the logic is the right thing to ask

The gate is sound only if the inferred logic covers everything the miner can reach, and for these two generators there is no exception to check:

- `IntRules.comp_set` and `RealRules.comp_set` are `Set.empty` (`invGenMiner.ml:599`, `:825`), so the int and real generators mine strictly from `Sys.state_vars sys` and the init/trans terms of each system in the subsystem map. Only `BoolRules` has a non-empty `comp_set` — the user candidates from `get_unknown_candidates` — and the bool generator is not gated.
- `mk_trans_sys` infers the logic from exactly that material: init, trans, the property terms, the types of the state variables and global constants, and the already-inferred logic of every subsystem (`transSys.ml:1899-1976`).

The logic is therefore a superset of what the miner walks — it over-approximates in the safe direction. Abstraction opens no hole either: an abstracted subsystem is still in the list the logic is unioned over.

## Selection, by type

One node each, default flags, counting the spawn at `-vv`:

| system | generators spawned |
| --- | --- |
| bool only | bool ×2 |
| `int` counter | bool ×2, int |
| `real` accumulator | bool ×2, real |
| `enum` | bool ×2, int |
| `subrange [0,3] of int` | bool ×2, int |
| `uint8` counter | bool ×2, ubv8 |

## Across the regression tree

All 472 files in `tests/regression/{success,falsifiable}`, run under both binaries:

| | files |
| --- | --- |
| spawned an int generator on a system with no integer arithmetic | 59 |
| spawned a real generator on a system with no real arithmetic | 382 |
| **at least one of the two** | **397** |
| both, on a system that had neither | 44 |
| unchanged | 75 (42 of which spawn no generator at all) |

The before and after sets differ only by removals, and only ever of `int` and `real`: no file loses a bool or bit-vector generator, and none gains anything.

One note on the method, because the obvious way to measure this is wrong. `Child process N (...) terminated normally` is written only for a child that exits on its own, so every engine killed once a property was decided goes uncounted; measured that way even the bool generator appears to come and go between runs of the same file. The number above counts `Starting new process ...`, logged at `-vv` by the child at spawn, which has no such race.

## Verified

- `make check` (`--profile strict`) clean.
- `cd tests && python3 -m pytest -q` — 486 passed in 36.1s.

## Where it does not apply

The `Running in parallel mode: ...` banner still lists every enabled module, including generators that will not be spawned. It is printed once, before any analysis, while the selection is per-analysis — the machine-integer gating has had the same discrepancy since it was added, and I left it alone rather than widen this change.

Enabling only `INVGENINTOS` on a system with no integers now leaves an analysis with no children at all. It ends at once and reports `unknown`, which is what `--enable INVGENMACH` on a system without bit vectors already did: `wait_for_children` returns as soon as `child_pids = []`.

### DIFF
--- a/src/invgen/invGenMiner.ml
+++ b/src/invgen/invGenMiner.ml
@@ -919,6 +919,75 @@ end
 module Real = MakeCandGen (RealRules)
 
 
+(* |===| What a domain has to mine. *)
+
+(* Types [IntRules] relates. Enumerations are not among them: they are
+integers as far as [TermLib.logic_of_sort] is concerned, but the integer
+rules skip enum-typed state variables and subterms. *)
+let is_int_type ty =
+  match Type.node_of_type ty with
+  | Type.Int | Type.IntRange _ -> true
+  | _ -> false
+
+(* Types [RealRules] relates. *)
+let is_real_type ty =
+  match Type.node_of_type ty with
+  | Type.Real -> true
+  | _ -> false
+
+(* Is this flat term a literal of its domain? *)
+let is_literal is_lit_symbol = function
+  | Term.T.Const sym -> is_lit_symbol (Symbol.node_of_symbol sym)
+  | _ -> false
+
+let is_int_literal =
+  is_literal (function `NUMERAL _ -> true | _ -> false)
+
+let is_real_literal =
+  is_literal (function `DECIMAL _ -> true | _ -> false)
+
+(* [has_mineable is_type is_lit sys] is [true] when [sys] or one of its
+subsystems has a state variable, or a subterm of its init or transition
+predicate, of a type [is_type] accepts that is not a literal.
+
+This is what [mine] walks for the domains whose [comp_set] is empty --
+[Int] and [Real]: the state variables of each system in the subsystem map,
+and the flat subterms of its [init] and [trans]. Literals are left out
+because [post_rules] injects constants of its own no matter what is found,
+so a system whose only integers are the numerals in an array index has
+nothing the miner can relate to anything else. *)
+let has_mineable is_type is_lit sys =
+  let exception Found in
+  let check_term term =
+    try
+      Term.eval_t ~fail_on_quantifiers:true (
+        fun flat _ ->
+          if is_type (to_term flat |> Term.type_of_term) && not (is_lit flat)
+          then raise Found
+      ) term
+    with Invalid_argument _ ->
+      (* Quantified: the miner will not look inside it, but neither have we. *)
+      raise Found
+  in
+  try
+    sys |> Sys.iter_subsystems ~include_top:true (
+      fun sys ->
+        if Sys.state_vars sys |> List.exists (
+          fun svar -> SVar.type_of_state_var svar |> is_type
+        ) then raise Found ;
+        check_term (Sys.init_of_bound None sys zero) ;
+        check_term (Sys.trans_of_bound None sys zero)
+    ) ;
+    false
+  with Found -> true
+
+(** Is there anything for the integer candidate term miner to work on? *)
+let has_mineable_int_terms sys = has_mineable is_int_type is_int_literal sys
+
+(** Is there anything for the real candidate term miner to work on? *)
+let has_mineable_real_terms sys = has_mineable is_real_type is_real_literal sys
+
+
 
 
 
--- a/src/invgen/invGenMiner.mli
+++ b/src/invgen/invGenMiner.mli
@@ -38,6 +38,19 @@ module UBV (L : sig val length : int end): CandGen
 (** Real candidate term miner. *)
 module Real : CandGen
 
+(** Is there anything for the integer candidate term miner to work on?
+
+    [true] when the system or one of its subsystems has a state variable, or
+    a subterm of its init or transition predicate, of type int or integer
+    range that is not a numeral. Enumerations do not count: they carry [IA]
+    in the logic of the system, but the integer rules skip them. *)
+val has_mineable_int_terms : TransSys.t -> bool
+
+(** Is there anything for the real candidate term miner to work on?
+
+    As {!has_mineable_int_terms}, for type real and decimals. *)
+val has_mineable_real_terms : TransSys.t -> bool
+
 (* 
    Local Variables:
    compile-command: "make -C .. -k"
--- a/src/kind2Flow.ml
+++ b/src/kind2Flow.ml
@@ -876,12 +876,19 @@ let process_invgen_arith_modules: TSys.t -> kind_module list -> kind_module list
     (* Logic given rather than inferred: nothing to consult. *)
     | _ -> true
   in
+  (* One question per domain, not one per module: the one state and two state
+     generators of a domain stand or fall together, and the miner walks the
+     init and trans of every subsystem to answer. *)
+  let mineable_int =
+    lazy (logic_has TermLib.IA && InvGenMiner.has_mineable_int_terms sys)
+  in
+  let mineable_real =
+    lazy (logic_has TermLib.RA && InvGenMiner.has_mineable_real_terms sys)
+  in
   modules |> List.filter (
     function
-    | `INVGENINT | `INVGENINTOS ->
-      logic_has TermLib.IA && InvGenMiner.has_mineable_int_terms sys
-    | `INVGENREAL | `INVGENREALOS ->
-      logic_has TermLib.RA && InvGenMiner.has_mineable_real_terms sys
+    | `INVGENINT | `INVGENINTOS -> Lazy.force mineable_int
+    | `INVGENREAL | `INVGENREALOS -> Lazy.force mineable_real
     | _ -> true
   )
 
--- a/src/kind2Flow.ml
+++ b/src/kind2Flow.ml
@@ -859,23 +859,31 @@ let process_invgen_mach_modules: TSys.t -> _ ISys.t -> kind_module list -> kind_
     | _ -> other_modules
   )
 
-(* Drop the integer (resp. real) invariant generators when the system's
-   inferred logic has no integer (resp. real) arithmetic: their candidate
-   term miners would come up empty. Machine integers are handled separately
-   by [process_invgen_mach_modules]. *)
+(* Drop the integer (resp. real) invariant generators when the system holds no
+   integer (resp. real) term their candidate miner can relate to another one.
+   Machine integers are handled separately by [process_invgen_mach_modules].
+
+   The logic answers the cheap half of the question -- no [IA] anywhere means
+   nothing to mine -- and the miner the exact half, which the logic cannot:
+   [IA] also stands for enumerations, which the integer rules skip, and for
+   bare numerals, which give the miner constants and nothing to relate them
+   to. *)
 let process_invgen_arith_modules: TSys.t -> kind_module list -> kind_module list
 = fun sys modules ->
-  match TSys.get_logic sys with
-  | `Inferred fs -> (
-    let open TermLib.FeatureSet in
-    modules |> List.filter (
-      function
-      | `INVGENINT | `INVGENINTOS -> mem IA fs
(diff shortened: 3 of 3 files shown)


daniel-larraz:
## Code review

**Verdict: the change is sound.** I checked the direction that actually matters — never dropping an engine that would have had real candidates — and could not find a hole:

- `TransSys.mk_trans_sys` folds subsystem logics, property terms, `init`/`trans`, global constraints, state-var types and global-const types into the inferred feature set, so `IA`/`RA` are present whenever any int/real state var, numeral, decimal, `to_int`/`to_real`/`bv2nat`, array-of-int/real, or UF constant of those types exists anywhere in the analyzed system's tree. `InvGenMiner.IntRules`/`RealRules` mine strictly less than that.
- Non-`` `Inferred `` logics (`--smt_logic <s>`, or `none`) fall through untouched, so the gate is conservative there.
- Ordering with `process_invgen_mach_modules` is fine: the expanded `INVGENBV`/`INVGENUBV` modules hit the `_ -> true` branch.
- An empty module list is harmless: `create_processes` returns `([], [])` and `InvarManager.wait_for_children` returns immediately when `!child_pids = []`. Confirmed empirically with `--enable INVGENINT --enable INVGENINTOS` on a Boolean-only model — clean exit, `unknown`, no hang.
- `InvGenDomain.Int.is_os_running` reads `Flags.enabled ()` rather than the filtered list, but since `INVGENINT` and `INVGENINTOS` are gated on the same feature they are always dropped or kept together, so no divergence is introduced.

`dune build @check` is clean and all 486 regression tests pass.

### One finding

**`src/kind2Flow.ml:873` — the `IA` gate misses enum-only systems.**

`TermLib.logic_of_sort` maps `Enum` to `IA`, but `InvGenMiner.IntRules.svar_rules` skips enum state vars, so for a system whose only "integer" content is enums the gate lets `INVGENINT`/`INVGENINTOS` through with nothing to mine but the constants `0`/`1` that `post_rules` injects.

Verified: an enum-only model still logs `Child process 1 (one state invariant generator (int)) terminated normally`, while a Boolean-only model correctly spawns nothing.

Not a correctness bug — just the wasted process the PR is meant to eliminate. Same reason the block comment's "would come up empty" is slightly off: the mined set is never empty, only useless.

daniel-larraz:
Fixed in 9096a759, and the finding generalizes: enums are one of two ways `IA` shows up with nothing behind it. The other is a bare numeral — an array index, say — which reaches the miner as a constant it can only compare to the constants `post_rules` was going to inject anyway.

So the gate now asks the miner instead of inferring from the feature set. `InvGenMiner` grew two predicates, next to the rules they speak for:

```ocaml
val has_mineable_int_terms : TransSys.t -> bool
val has_mineable_real_terms : TransSys.t -> bool
```

Both ask exactly what `mine` walks for a domain whose `comp_set` is empty — the state variables of every system in the subsystem map, and the flat subterms of its `init` and `trans` — for a term of the right type that is not a literal. `iter_subsystems ~include_top:true` covers the same systems `mk_sys_map` recurses through, and an exception cuts the walk short at the first hit. The logic stays in front as the cheap half of the question: no `IA` anywhere means there is nothing to walk, and that is one lookup instead of a traversal.

I also updated the block comment, which as you say was describing an empty set rather than a useless one.

## Two things the new gate deliberately does not do

**It is not a check for a state variable of the type.** In

```
ok = (0.0 <= x and x <= 1.0) => (int x >= 0);
```

there is no integer state variable anywhere, and `IntRules.flat_rules` still has `(to_int x)` to mine. Gating on state-var types would drop that. Measured: `bool ×2, int, real` — both generators still spawn.

**It does not exclude every enumeration**, because `IntRules` skips *enum-typed* terms and Kind 2 does not always give an enum expression an enum type. Your reading of `svar_rules` is right, but enum values are encoded as numerals, so

```
c = red -> (if a then green else pre c);
```

puts `(ite top.usr.a@0 1 top.impl.usr.c@-1)` in `trans`, and `Term.type_of_term` calls that `Int`. `flat_rules` matches `Type.Int`, the kids are all var-or-const, and the term is mined. The generator is kept there, and dropped for the constants-only shape:

| model | int generator |
| --- | --- |
| `c = red -> pre c` | not spawned |
| `c = red -> (if a then green else pre c)` | spawned, mines `(ite a 1 c)` |
| `tests/regression/falsifiable/enum.lus` | not spawned |

The middle row is why I did not gate on "the system has no int state var and only enums": it would drop a generator with a real candidate. If you would rather the enum encoding never feed the integer generator, that is a change to `IntRules` — teach `flat_rules` to skip integer terms that are enum values in disguise — and I would keep it out of this PR.

## What it changes over the tree

Re-running the earlier survey against the new gate, same method (spawn-time `Starting new process ...` at `-vv`, 472 files in `tests/regression/{success,falsifiable}`):

| | old gate | new gate |
| --- | --- | --- |
| files that no longer spawn an int generator | 59 | **85** |
| files that no longer spawn a real generator | 382 | **384** |
| at least one | 397 | **403** |

28 files move, 26 of them for the enum-and-numeral reason you found; `falsifiable/enum.lus` is one of them. The two real-generator ones are datatype models like `falsifiable/adt_eq_diff_payload.lus`, where `RA` comes from a `real` field of a datatype that no term ever selects — the same over-approximation, one type constructor further out.

Still no additions and still no non-int/real removals: before and after differ only by removals of `int` and `real`.

`make check` clean, 486 regression tests pass.

daniel-larraz:
## Re-review of 9096a759

The fix is correct. I checked the direction that matters — dropping a generator that would have produced real invariants — and could not construct or find such a case.

- `is_int_type`/`is_real_type` match `IntRules.flat_rules` / `RealRules.flat_rules` dispatch exactly. `Type.Enum` is a distinct constructor that both `svar_rules` and `flat_rules` fall through, so excluding it is right, and the `(ite a 1 c)` caveat in the commit message holds: `Term.type_of_term` returns the type of `ITE`'s *second* argument (`src/terms/term.ml:801`), read identically by the predicate and by `flat_rules`.
- The traversal mirrors `mine`'s `mk_sys_map`: same system set (dedup by scope vs. by assoc, same coverage), literally the same `Sys.init_of_bound None sys zero` / `Sys.trans_of_bound None sys zero` / `Sys.state_vars sys` calls, same `Term.eval_t ~fail_on_quantifiers:true`.
- Excluding literals is sound. If the predicate returns false then no `Int`/`IntRange` state variable exists in any system and every int-typed flat subterm is a `` Const (`NUMERAL _) ``. Under those, `svar_rules` contributes nothing, `flat_rules` contributes only numerals, and `post_rules` adds only ground constants — so every candidate is ground, `Graph.has_svars` (`src/invgen/invGenGraph.ml:242`) is false, and `InvGen.main` already discards that graph at `src/invgen/invGen.ml:711`. Nothing is lost.
- The `logic_has IA` conjunct stays safe. Every way an int-typed term can arise — vars, numerals, `SELECT`, datatype selectors, `ITE`, `PLUS`/`TIMES`/`MOD`/`ABS`, `TO_INT`/`BV2NAT`/`UBV_TO_INT`/`SBV_TO_INT` — forces `IA` in `TermLib.logic_of_flat`. The one gap is `App(uf, args)` returning int, which adds only `UF`, and `filter_terms_with_unsupported_symbols` strips exactly those terms from the candidate set.

Empirical check, one engine enabled per run with a 5 s timeout (instant exit = dropped, ~5 s = kept):

| model | INVGENINTOS | INVGENREALOS |
|---|---|---|
| bool-only | dropped | dropped |
| enum-only (`c = Red` comparisons, no int `ite`) | **dropped** ← the fix | dropped |
| enum with `if a then Red else …` | kept | dropped |
| subrange counter | kept | dropped |
| `(int x >= 0)` over a real, no int svar | **kept** | kept |
| real accumulator | dropped | kept |
| int counter in a *subsystem*, bool-only top | **kept** | dropped |
| real accumulator in a subsystem, bool-only top | dropped | **kept** |
| int only in a `--%PROPERTY` / contract guarantee | kept | kept |
| fixed-size bool array | dropped | dropped |

No test in `tests/` references these modules.

### One finding

**`src/kind2Flow.ml:879` — the miner predicate is called from inside `List.filter`**, so it re-walks init and trans of every subsystem once per matching module — twice for int under `--enable INVGENINT --enable INVGENINTOS`, and again for real — serially in the supervisor before any engine spawns. Hoisting into two `lazy` bindings collapses it to at most one walk per domain.

### Two non-findings, both failing safe

- `check_term`'s `with Invalid_argument _ -> raise Found` makes the gate a no-op for any system whose init or trans contains a quantifier. Conservative and deliberate; I could not get Kind 2's array encoding to emit one in practice (`arr[i] = a` over `bool^n` still got int correctly gated), so it rarely costs anything.
- An int-typed nullary UF constant (`__function_of_inputs`) is not a numeral, so the predicate keeps the generator — but such candidates carry no state var and the graph is discarded by `has_svars`. Residual waste, not a correctness issue.

daniel-larraz:
Hoisted in 4e6eb1fd — one `lazy` per domain, forced by both of that domain's modules:

```ocaml
let mineable_int =
  lazy (logic_has TermLib.IA && InvGenMiner.has_mineable_int_terms sys)
in
...
| `INVGENINT | `INVGENINTOS -> Lazy.force mineable_int
```

Counted with a print at the top of `has_mineable`, on the enum-only model with `--enable INVGENINT --enable INVGENINTOS`: **two walks before, one after**. The `logic_has` conjunct sits inside the thunk, so a domain the logic has already ruled out still costs nothing at all.

Selection is unchanged across the ten models from your table, `make check` is clean and the 486 regression tests pass.

On the two non-findings, agreed on both, and one of them makes the other smaller than it looks: since `has_svars` discards a ground-only graph at `invGen.ml:711`, the int-typed nullary UF case you flagged costs a process spawn and a mining pass, not a run. Same for the quantifier fallback — it fails toward the behaviour on `main`, which is to always spawn. If either turns out to matter, the place to fix it is `IntRules`/`RealRules`, not the gate.

PR #1476, opened 2026-08-28
Keep ADT values in canonical form

daniel-larraz:
## Summary

Non-recursive ADTs are desugared to a record holding a tag and one payload slot per constructor field; the slots of the constructors other than the active one held junk. Keeping junk unobservable required canonicalizing it at every use: ADT equality was guarded field by field by the tag, and every set/map membership, insertion and update wrapped each payload slot in `ite(tag = C, slot, default)`. Those `ite`s sit in array index positions and defeat the solver's congruence reasoning: on a set keyed by a two-constructor ADT, k-induction took ~35s where the same model with a struct key takes 0.3s, and a single-constructor ADT timed out.

This PR keeps every ADT value in **canonical form** instead: the inactive payload slots always hold the default value of their type.

- Constructor applications already fill them with defaults; free values (inputs, oracles, undefined and abstracted outputs, free constants, quantified variables) are constrained to that form. The constraints (`LustreDesugarADTs.mk_canonical_exprs`) are conjoined into `init`/`trans` of every node instance, concrete or abstracted by its contract, next to the range constraints of enumerated types — not assumptions, and not proof obligations, since every source construct preserves the form. Quantified variables are relativized in `compile_quantifier`, so properties keep their source form.
- ADT equality and set/map membership, insertion and update become plain structural operations on the record; `adt_canonicalize_key` and the guarded equality are gone.

A **selector** applied to a value built with another constructor keeps yielding an arbitrary value that is a function of the ADT value, rather than the default the record now holds: it compiles to `ite(tag = C, payload, junk(value))` with `junk` an uninterpreted function of the record's components (`adt_junk_ufs`). The normalizer, which already knows the guards of every selector for its proof obligation, marks a selector whose constructor is established at its own `pre` depth as a plain projection: by a lazy guard, a match arm, or — for the value only, obligations are unchanged — the condition of an eager `if`/`and`/`or`/`=>`. In a model whose selectors are guarded no junk function is emitted at all.

## Results

`Lock_Server.lus` (Kondo benchmark, set keyed by a two-constructor ADT): ~35s → 0.3s, now by 2-induction like the struct version; the single-constructor variant: timeout → 0.2s. Reachability results and the other Kondo models are unchanged.

## Behavior changes

- Two containers holding the same elements now compare equal whatever the encoding held at non-canonical indices (`tests/regression/success/set_adt_key_junk_index_equality.lus`, which fails on `main`).
- A selector on the wrong constructor of a *constructed* value is arbitrary rather than the default, and its value is a function of the ADT value: two equal values (same constructor and payload) agree on it, where `main` could return two different values. Both differences are only observable next to a failing selector obligation.
- Recursive ADTs are unaffected (SMT-LIB datatypes have no inactive slots).

## Also in this PR

- `syn_expr_equal` distinguishes a selector from a plain projection of the same field, so the cache of generated locals cannot serve an unguarded read with a guarded one's local.
- The detected SMT logic includes `UF` whenever a system declares uninterpreted functions, and the sorts of their signatures.
- `Simplify.if_then_else` accepts branches of abstract, datatype, bit-vector and array sorts, needed once a conditional's predicate can mention an uninterpreted function (counterexample reconstruction).

## Tests

12 regression tests added (canonical free values, set/map key semantics, selector value semantics, eager/lazy guards, shared locals); full suite passes; `--profile strict` build is clean.


daniel-larraz:
## Code review — ADT canonical form + selector obligations

Reviewed at `8de46aa` against `main` (merge-base `501fa73`). Every correctness finding below was reproduced by building this branch and, where the claim is a regression, diffing behaviour against a `main` build. The full `success` (348) and `falsifiable` (163) suites show **no outcome regressions** — which is exactly why most of these survived: they are engine-level or diagnostic-level failures that don't flip a recorded test verdict.

---

### 1. Mangled payload-name collision makes the transition system unsatisfiable

`src/lustre/lustreDesugarADTs.ml:467` (and `:105`)

`payload_field_name_of` maps `A_B(x)` and `A(B_x)` to the same record field `A_B_x`. `all_payload_fields` concatenates the per-constructor lists with no dedup, so the record carries the slot twice, and the `HStringMap.fold` that finds the owning constructor returns only the first match. The emitted constraint is `t.T_tag <> A => t.A_B_x = 0` with nothing for `A_B`, so `t = A_B(5)` is unreachable and **every property proves vacuously**.

This is live on the PR's own `tests/regression/success/adt_selector_mangled_name_collision.lus`. Adding one line to it:

```
  check "should be falsifiable" b;   -- b is a free boolean input
```

```
should be falsifiable: valid (k=1)
```

The root cause is the first-match fold ignoring ambiguity. That same fold is written three times (here, plus `lustreAstNormalizer.ml:527` and `:633`) — a shared `ctor_of_payload_field : adt_info -> HString.t -> HString.t option` that returns `None` on ambiguity would fix all three sites at once.

### 2. Counterexample rendering aborts the whole analysis

`src/lustre/lustrePath.ml:1155`

The unconditionally emitted selector obligation turns previously-valid programs into ones that produce a counterexample, which makes the pre-existing assert in the ADT reconstruction newly reachable when a non-recursive ADT has a payload field of a recursive ADT type.

```lustre
datatype L = Nil | Cons (hd: int, tl: L);
datatype W = P (l: L) | Q (n: int);
node main(w: W) returns (ok: bool);
let
  ok = Q?(w) => w.n = w.n;
  check ok;
tel
```

`main`: `InvProp[L6C3]: valid (k=1)`, no error. This branch:

```
<Error> Runtime error in invariant manager: File "src/lustre/lustrePath.ml", line 1155, characters 22-28: Assertion failed
Selector[L5C24]: unknown
Selector[L5C18]: unknown
InvProp[L6C3]: unknown
```

A program that used to verify now returns no result at all.

### 3. `is_canonical_bound_var` is forgeable by a user identifier

`src/lustre/lustreDesugarADTs.ml:273`

The check recognises Kind 2's generated binders by the bare suffix `_canon`, which a source identifier can carry, so such a quantified variable silently loses its canonical-form restriction in `compile_quantifier` (`lustreNodeGen.ml:1217`).

```lustre
datatype D = A (a: int) | B (b: bool);
node main(i: int) returns (ok: bool);
let ok = true;
  check "canonical" forall (d: D) (A?(d) => d = A(d.a));
tel
```

`valid (k=1)`. Renaming the binder `d` to `d_canon`, changing nothing else: `invalid after 0 steps`. The `exists` direction is unsound rather than merely noisy — an existential over a `*_canon` binder can be witnessed by a value no canonical variable can equal.

The file already has the forge-proof convention this wants: `is_generated_bound_var` (`:836`, digit prefix, which source Lustre cannot produce), and the comment at `:434` even cites it by name. Generated names are `<int>_canon`, so requiring the digit prefix is a one-line fix.

### 4. `IsConstructor` / `Selector` reach `simplify`'s `assert false`

`src/simplify.ml:839`

For a recursive ADT, `mk_selector_obligation` (`lustreAstNormalizer.ml:2049`) emits the obligation as an `A.ADTTester`, so a property term now carries SMT `IsConstructor`/`Selector` symbols. The PR extended `simplify.ml` only at the ITE cases (`:1599`); the term dispatch at `:839` still asserts.

The PR's own `tests/regression/success/adt_selector_guard_pre.lus`:

```
<Error> Runtime error in property directed reachability (QE): File "src/simplify.ml", line 839, characters 26-32: Assertion failed
<Warning> Child process 6 (property directed reachability (QE)) terminated on exception
```

Clean on `main`. The test still passes because other engines discharge the properties, so the suite hides a QE engine that dies on every run over a recursive ADT.

### 5. Junk UFs take `Bool` arguments, killing IC3IA

`src/lustre/lustreNodeGen.ml:1420`

`args = X.values cexpr` takes *every* leaf of the compiled record, so an ADT with a `bool` payload yields a UF with a Bool argument.

```lustre
datatype D = A (a: int) | B (b: bool);
node main(x: D) returns (ok: bool);
var c: int;
let
  c = 0 -> (if x.a > 0 then pre c + 1 else pre c);
  ok = c >= 0;
  check ok;
tel
```

With `--enable IC3IA`:

```
<Error> Runtime failure in property directed reachability (IA): SMT solver failed: "Functions with Bool arguments not supported"
InvProp[L7C3]: unknown
```

Guarding the selector, or changing the payload to `B (b: int)`, gives `valid (k=1)` under IC3IA. It also fires on three of this PR's own `falsifiable` tests (`adt_selector_eager_guard.lus`, `adt_selector_shared_local_junk.lus`, `adt_selector_arbitrary_value.lus`); no such error occurs anywhere in the suite on `main`. An ADT with a boolean payload plus an unguarded selector read is a routine pattern, so this removes IC3IA from any such model.

### 6. Eager value facts are misapplied to selectors inside node-call arguments

`src/lustre/lustreAstNormalizer.ml:2031`

`selector_statically_guarded` consults `value_context` for selectors sitting inside call arguments, but a call's internal state still advances on steps where the eager condition is false, so compiling the selector as a plain payload projection there is observable later.

```lustre
node Sum(v: int) returns (s: int); let s = v -> pre s + v; tel
node main(x: D) returns (ok: bool);
var s1: int;
let
  s1 = Sum(x.a);
  ok = (not A?(x)) or (Sum(x.a) = s1);
  check ok;
tel
```

The second `x.a` is under the eager `or`, whose fact strips to `A?(x)`, so it compiles to the raw payload (the default value whenever `x` is `B`) while the `x.a` feeding `s1` compiles to the junk UF. The two `Sum` instances then accumulate different sequences: `invalid after 1 steps`. Replacing the guard with a non-tester makes the identical property `valid (k=2)`.

Lazy guards are safe here because they are also pushed into `mk_fresh_call`'s activation condition; eager value facts are not.

### 7. The junk value of an ADT-typed payload is not a legal ADT value

`src/lustre/lustreNodeGen.ml:1448`

One independent UF per scalar leaf, with no canonical-form constraint applied to the result — while `compile_quantifier` restricts quantified variables to canonical values only.

```lustre
datatype Inner = X (s: int) | Y (u: bool);
datatype D = A (a: Inner) | B (c: int);
node main(d: D) returns (ok: bool);
let ok = true;
  check "junk is a value" exists (v: Inner) (v = d.a);
tel
```

`d.a` is arbitrary but must still denote *some* `Inner`, so this is semantically valid; it reports `invalid after 0 steps` with counterexample `d = B(4)`, because the solver picks junk leaves (`tag = X`, `u = true`) that no canonical `v` can match. With an int-only `Inner` the same shape instead aborts with `Invalid_argument("numeral_of_term")` from `lustrePath.ml:1130`, which only guards `Not_found`. The same gap lets a non-canonical value be stored in or queried against a set or map.

---

### Performance / scoping

- **`lustreNodeGen.ml:3580`** — `compile_const_decl` emits `mk_canonical_exprs` for every free constant into `G.global_constraints`, which `lustreTransSys` asserts in every system and slicing never filters, unlike node-level `adt_constraints` (filtered against surviving state vars at `lustreTransSys.ml:2899`). A completely *unused* `const arr: D^3;` injects `forall i. 0<=i<3 => (tag <> A => A_a[i] = default)` everywhere: with `--enable IC3IA` the property reports `unknown` and `IC3IA disabled ... quantified formulas over arrays are not supported`; changing only that constant's type to `int^3` gives `valid (k=1)`.
- **`lustreTransSys.ml:3468`** — `ufs = function_ufs @ globals.G.adt_junk_ufs` attaches the program-wide junk UF list to *every* node's system, and the new `transSys.ml` logic code adds `TermLib.UF` whenever `ufs <> []`. So `N*U` UF declarations across solver instances, and every ADT-free node's logic gains UF (QF_LIA becomes QF_UFLIA). Filtering to the symbols the system's own init/trans terms mention — as `function_ufs` already is, per node — avoids both.
- **`generatedIdentifiers.ml:217`** — `selector_obligations` is combined with `ids1 @ ids2` inside `union`, invoked once per subexpression at 36 call sites, so obligations from deep in an expression are re-copied at every enclosing operator: O(k·d), quadratic in obligation count along selector-bearing chains.

### Duplication

- **`lustreNodeGen.ml:1441`** *(unverified, but cheap to pre-empt)* — the junk UF is cached by a pretty-printed type string, and `Type.pp_print_type` renders signed and unsigned bitvectors identically as `(_ BitVec n)` (`type.ml:240`, `:247`). Two instantiations of a polymorphic ADT differing only in signedness collide on one name; because the Hashtbl short-circuits, `UfSymbol.mk_uf_symbol`'s own signature check (`ufSymbol.ml:210-226`), which would raise `Invalid_argument`, never runs. Keying on the structured `(type_name, field, index, arg Type.t list)` — `Type.t` is hash-consed — and building the display string only on insertion removes both the collision and the masked check.
- **`lustreAstNormalizer.ml:2011`** — the local `strip_not` is byte-for-byte `process_for_display` (`:411`), added by this same PR. Guard recognition and guard display must agree for a displayed obligation to match the term the solver checks; two copies invite drift.
- **`lustreDesugarADTs.ml:484`** — `mk_canonical_exprs` re-defines `mk_is_default`'s `fresh_var` and repeats its Tuple/Group and Array branches verbatim (`:485-500` vs `:396-413`), including the `0 <= i < len` wrapper, plus the `RefinementType`/`History` branches. A fix to array bounds handling in one would miss the other.
- **`lustreAstNormalizer.ml:1998`** — the `UserType -> find_opt name adt_map` lookup is duplicated in `selector_statically_guarded` and `mk_selector_obligation` (`:2037`), and is a weaker form of `LDAT.adt_info_of_type` (`lustreDesugarADTs.ml:157`), which also chases type synonyms and refinement types. With an aliased or refined ADT type both return `None`, and `mk_selector_obligation` then hits its `| None -> assert false`. Resolving `adt_info` once at the single `FieldProject` call site (`:2661`) and passing it to both would fix this.
- **`lustreDesugarADTs.ml:461`** — hand-builds the tag projection despite `tag_of` (`:141`) in the same file; the tag-tester shape is now encoded at four sites across two modules (`:461`, normalizer `:2006` and `:2048`).

---

*Review produced with Claude Code.*

daniel-larraz:
Thanks for the thorough review — every finding reproduced. Addressed in the latest commit; per item:

**1. Payload-name collision** — fixed. `adt_info` now carries a `(constructor, user name, internal name)` table built with unique internal names (`A_B(x)` → `A_B_x`, `A(B_x)` → `A_B_x_1`); the prefix mangling and stripping are gone, so the three first-match folds can no longer be ambiguous. Regression test: `falsifiable/adt_payload_name_collision_not_vacuous.lus` (`check b` on a free Boolean, plus reachability of both constructors).

**2. Counterexample reconstruction abort** — fixed, two parts. A recursive ADT in a payload is now `AdtFieldPlain` (a single datatype-sorted variable, not a record with a tag). Independently, the `assert false` fired whenever a *nested* value's variables were sliced away (only the tag survives when nothing reads the payload) — it now prints `_` there. That second case is pre-existing: `main` at `8f8dc331` fails identically on `datatype Inner = X(s:int)|Y(u:bool); datatype D = A(a: Inner)|B(c:int); ok = B?(d)`. Test: `falsifiable/adt_recursive_payload_selector_cex.lus`.

**3. Forgeable `_canon` suffix** — fixed: `is_canonical_bound_var` also requires the digit prefix (`is_generated_bound_var`, moved up in the file). Test: `success/adt_canonical_binder_name.lus` (`d_canon`, `x_canon` binders).

**4. `simplify` assertion on `IsConstructor`/`Selector`** — fixed, but it is not new: `main`'s own binary hits `simplify.ml:839` on `adt_selector_guard_pre.lus` too (with IC3QE alone and with the default engines), since the obligation of a recursive-ADT selector is an `ADTTester` on `main` as well. Testers and Boolean selectors are now atoms for the negation and for `extract.ml`'s literal extraction, and evaluate on a datatype value (constructor application) in the simplifier. IC3QE no longer dies on that test; it runs without concluding within the timeout, which is a separate limitation.

**5. Boolean UF arguments** — fixed: Boolean components are passed as 0/1 and a Boolean result is read back from an integer function. IC3IA proves your `c >= 0` example.

**6. Eager value facts inside call arguments** — fixed: `handle_call` (and the inlined path) normalize arguments with an empty `value_context`. Your `Sum` example is valid: `falsifiable/adt_selector_eager_call_argument.lus`.

**7. Junk of an ADT-typed payload** — fixed: the arbitrary value is canonical (components of constructors its arbitrary tag does not select hold the default), and an enumerated-type result stays in its range, so `exists (v: Inner) (v = d.a)` is valid, also with an int-only `Inner` (no `numeral_of_term` abort; `lustrePath` now also guards `Invalid_argument`). Test: `falsifiable/adt_selector_junk_is_a_value.lus`. A refinement type's predicate is not enforced on such a value; that matches `main`, where the refinement constraint on a payload slot is guarded by its constructor.

**Performance/scoping**
- Free constants: their canonical-form constraints live in a separate `G.adt_global_constraints` and are only asserted for constants some node's equations mention; your unused `const arr: D^3` no longer disables IC3IA (`valid (k=1)` with `--enable IC3IA`).
- Junk UFs: a system's `ufs` only includes those its own init/trans terms apply, so ADT-free systems keep their logic (a plain set model declares none and stays `QF_LIRA`).
- `selector_obligations` `@` in `union`: left as is — that is #1437's code and the other gids fields are combined the same way; happy to change it separately.

**Duplication** — done: the junk cache is keyed on `(component, arg types, result type)` with `Type.equal_types`/`hash_type` and names get a unique counter (signed/unsigned BV instantiations no longer collide); `process_for_display` replaces the local `strip_not`; `mk_is_default` and `mk_canonical_exprs` share `over_components` and `fresh_canonical_var`; the ADT of a selector is resolved once at the `FieldProject` site with `LDAT.adt_info_of_type` (synonyms and refinement types included) and passed to both `selector_statically_guarded` and `mk_selector_obligation`; `tag_of` is reused.

Full suite: 539 passed; `--profile strict` clean; `Lock_Server` still 0.3s by 2-induction.

daniel-larraz:
## Re-verified at `6193e379`

Rebuilt the branch and re-ran each original reproduction. **All seven correctness findings are fixed.**

| # | Finding | Before | After |
|---|---|---|---|
| 1 | Payload-name collision | free input `b` **valid** (vacuous) | `b` **invalid after 0 steps**, `ok` valid |
| 2 | `lustrePath.ml:1155` abort | assertion, all props `unknown` | no abort, `InvProp` **valid (k=1)** |
| 3 | `_canon` binder | `d` valid / `d_canon` invalid | both **valid (k=1)** |
| 4 | `simplify.ml:839` assert | QE child killed | no error, both props valid |
| 5 | Bool-arg junk UF | IC3IA children die | **valid (k=1)** |
| 6 | Eager fact in call argument | `invalid after 1 steps` | **valid (k=1)** |
| 7 | Junk not a legal ADT value | `invalid after 0 steps` | **valid (k=1)** |

Full regression suite on this commit: **539 passed**, no failures.

Worth noting that (1) and (3) were fixed at the root rather than at the symptom. `payload_field_names` now builds a real `(ctor, user, internal)` bijection, and its uniquifier's `List.mem cand naive_names` guard correctly stops a disambiguated name from stealing a name another field will naively want — which retires the ambiguous first-match fold entirely rather than teaching it to detect ambiguity. `is_canonical_bound_var` now requires `is_generated_bound_var` *and* the suffix.

Four of the quality items are resolved too: the junk UF is keyed on a structured `JunkTbl` key (this also closes the signed/unsigned `(_ BitVec n)` name collision I'd flagged as unverified) and is now filtered per system to the symbols actually applied in that system's terms; `strip_not`, `tag_of`, the `adt_info` lookup, and the shared `fresh_var` / `in_bounds` traversal are all deduplicated.

---

### Still open

**Free-constant canonicity constraints are never sliced** — `lustreNodeGen.ml:3580`

Node-level `adt_constraints` got the surviving-state-var filter at `lustreTransSys.ml`; global constants never did, so an entirely unused constant still injects a quantified array formula into every system:

```lustre
datatype D = A (a: int) | B (b: bool);
const arr: D^3;                 -- never referenced
node main(i: int) returns (ok: bool);
var c: int;
let
  c = 0 -> pre c + 1;
  ok = c >= 0;
  check "ind" ok;
tel
```

```
$ kind2 --enable IC3IA f8.lus
<Warning> IC3IA disabled: the interpolating solver does not support the arrays this system uses.
ind: unknown
```

Changing only that line to `const arr: int^3;`, or dropping it, gives `ind: valid (k=1)`.

Two minor leftovers: `generatedIdentifiers.ml:217` still combines `selector_obligations` with `@` inside `union` (called once per subexpression), and the ctor-lookup fold is down from three copies to two (`lustreAstNormalizer.ml:527` and `:633`).

---

*Re-verification produced with Claude Code.*

daniel-larraz:
Thanks for re-verifying.

**Free-constant canonicity constraints** — the filter is in effect: an SMT trace of your `f8.lus` on this branch contains no occurrence of `arr` at all, so no constraint over it is asserted. What disables IC3IA there is the constant's *type*, not a constraint: `D^3` flattens to arrays including a Boolean-element one (`arr.B_b`), and global constants are never sliced, so that array sort is part of every system's signature and the interpolating-solver check rejects it. The same happens on `main` at `8f8dc331` with your file unchanged, and on both `main` and this branch with `const arr: bool^3;` in its place, while `int^3` is `valid (k=1)` on both:

| constant | main | this branch |
|---|---|---|
| `D^3` (unused) | IC3IA disabled, `unknown` | IC3IA disabled, `unknown` |
| `bool^3` (unused) | IC3IA disabled, `unknown` | IC3IA disabled, `unknown` |
| `int^3` (unused) | `valid (k=1)` | `valid (k=1)` |

So this is pre-existing and independent of the PR; slicing unused global constants would be the fix, and that is a separate change.

**Ctor-lookup folds** — the two remaining copies now use `LDAT.ctor_of_payload_field` (the payload-name table is the one place that knows a field's constructor).

**`selector_obligations` in `union`** — left as is: it is #1437's code and every other list field of `gids` is combined the same way; I'd rather not change one of them in this PR.

Suite: 539 passed; `--profile strict` clean.

daniel-larraz:
## Correction — the free-constant finding was mine, and it was wrong

You're right, and my control experiment was insufficient. I varied `D^3` against `int^3`, which changes two things at once: whether the constant involves an ADT, *and* whether its flattening produces a Boolean-element array. Holding the first fixed isolates the cause:

| unused constant | `main` @ `8f8dc331` | this branch |
|---|---|---|
| `const arr: D^3;` | IC3IA disabled, `unknown` | IC3IA disabled, `unknown` |
| `const arr: bool^3;` | IC3IA disabled, `unknown` | IC3IA disabled, `unknown` |
| `const arr: int^3;` | `valid (k=1)` | `valid (k=1)` |
| `const arr: real^3;` | `valid (k=1)` | `valid (k=1)` |

`bool^3` involves no ADT and so generates no canonical-form constraints at all, yet it fails identically — and `D^3` fails the same way on `main`, which has no `mk_canonical_exprs`. So the trigger is the Boolean-element array sort of an unsliced global constant, exactly as you said; it is pre-existing and independent of this PR. Slicing unused global constants is a separate change.

I should have run the `bool^3` control before posting that. **Withdrawn** — nothing from my review is outstanding against this PR.

## Re-verified at `fec7f21b`

- `ctor_of_payload_field` is now shared: `lustreDesugarADTs.ml:133`, exported at `.mli:88`, used at `lustreDesugarADTs.ml:535` and `lustreAstNormalizer.ml:527` / `:629`. The payload-name table is the single place that knows a field's constructor.
- All seven correctness reproductions still behave correctly at this head (F1 `should be falsifiable: invalid`, F2 no abort, F3 both binders `valid`, F4 no assertion, F5 IC3IA `valid (k=1)`, F6 `valid (k=1)`, F7 `valid (k=1)`).
- Regression suite: **539 passed**.

No objection to leaving `selector_obligations`'s `@` in `union` as is — it is #1437's code and consistent with the other `gids` fields; changing one of them here would be the odd choice.

LGTM.

---

*Re-verification produced with Claude Code.*

lorchrob:
Finding reported by Claude

### Eager value facts still reach the arguments of `condact`, `restart every` and `merge`/`activate`

`src/lustre/lustreAstNormalizer.ml:2382`, `:2396`, `:2411`, `:2427`

The fix for the earlier "eager value facts inside node-call arguments" finding clears
`value_context` in `handle_call` (`:2309`) and on the inlined path (`:2354`), but the four
other sites that normalize node-call arguments still pass `info` through unchanged:

| line | construct |
|---|---|
| `:2382` | `Condact` |
| `:2396` | `RestartEvery` |
| `:2411` | `Merge` / `Activate` arm |
| `:2427` | `Merge` / `Call` arm |

All four reach `abstract_node_arg` with the enclosing eager condition still in
`value_context`, so `selector_statically_guarded` marks a selector in the argument as a
plain projection. The reasoning that justifies the fix for `Call` applies verbatim here:
the callee's state advances on steps where the eager condition is false, so the argument
value on those steps is observable later.

```lustre
datatype D = A (a: int) | B (b: bool);

node Sum(v: int) returns (s: int); let s = v -> pre s + v; tel

node main(x: D; r: bool) returns (ok: bool);
var s1: int;
let
  s1 = (restart Sum every r)(x.a);
  ok = (not A?(x)) or ((restart Sum every r)(x.a) = s1);
  check "eager_restart" ok;
tel
```

The second `x.a` sits under the eager `or`, whose fact strips to `A?(x)`, so it compiles to
the raw payload while the `x.a` feeding `s1` compiles to the junk UF; the two `Sum`
instances then accumulate different sequences.

| call form | `main` @ `28c9138e` | this branch @ `b53f7a1e` |
|---|---|---|
| `(restart Sum every r)(x.a)` | `valid (k=1)` | **`invalid after 1 steps`** |
| `condact(c, Sum(x.a), 0)` | `valid (k=1)` | **`invalid after 1 steps`** |
| `merge c (true -> (activate Sum every c)(x.a)) (false -> 0)` | `valid (k=1)` | **`invalid after 1 steps`** |

(The `condact` and `merge` variants are the same node with `s1` and the guarded read
replaced by that call form.)

Clearing `value_context` inside `abstract_node_arg` itself, rather than at each call site,
would cover all five call forms by construction and keep a future call form from
reintroducing this.

daniel-larraz:
Thanks — reproduced all three, fixed in `1a46d534`.

`value_context` is now cleared inside `abstract_node_arg` itself rather than at the call sites, as you suggested: every call form normalizes its arguments through that one function, so `Call`, `Condact`, `RestartEvery` and both `Merge` arms are covered by construction and a future call form cannot reintroduce this. The clearing in `handle_call` is gone (`mk_fresh_call` does not read the field; `selector_statically_guarded` is its only reader). The inlined path keeps its own clearing, since it normalizes arguments directly. Lazy guards are unaffected — they stay in `call_context` and are still pushed into the call's activation condition.

| call form | before | after |
|---|---|---|
| `(restart Sum every r)(x.a)` | `invalid after 1 steps` | **`valid (k=1)`** |
| `condact(c, Sum(x.a), 0)` | `invalid after 1 steps` | **`valid (k=1)`** |
| `merge c (true -> (activate Sum every c)(x.a)) (false -> 0)` | `invalid after 1 steps` | **`valid (k=1)`** |

("before" is a build of this same tree with the change stashed.)

Regression tests, one per call form, next to the existing `adt_selector_eager_call_argument.lus` and in `falsifiable` for the same reason — the obligations of the unguarded reads are falsifiable:

- `tests/regression/falsifiable/adt_selector_eager_restart_argument.lus`
- `tests/regression/falsifiable/adt_selector_eager_condact_argument.lus`
- `tests/regression/falsifiable/adt_selector_eager_merge_argument.lus`

Full suite: 576 passed; `--profile strict` clean.

lorchrob (approved):

PR #1479, opened 2026-08-31
Keep branch conditions that carry proof obligations

daniel-larraz:
Follow-up to #1478, which fixed one way `simplify_tree` lost information. This fixes the other one.

`simplify_tree` collapses `Node (i, c, j)` into `i` when the two subtrees are equal, which deletes the condition `c` from the AST. If `c` contains a user-written ADT selector, the proof obligation that selector carries (`mk_selector_obligation`, `lustreAstNormalizer.ml`) is deleted along with it, and Kind 2 silently stops checking that the scrutinee was built with the constructor owning the field:

```lustre
datatype IntOpt = None | Some (val: int);

node main(p: IntOpt) returns (g: bool);
let
  if p.val > 0 then g = true; else g = true; fi
tel
```

Before: `System main has no property, skipping verification step.`
After: `Selector[L5C6]: invalid after 0 steps -- Some?(p)`

The same program written as a plain ITE (`g = if p.val > 0 then true else true;`) already reported the obligation, and so did the if block as soon as its two branches differed — only the collapsed case lost it. `when` blocks share `simplify_tree`, so they had it too.

## The fix

Collapse a branch only when its condition can be dropped without losing an obligation. `AH.expr_is_droppable` decides that with a whitelist, so an expression form added later defaults to keeping the condition, which is always sound. Node calls, binders, pattern matching, and constructs whose operands include a type (which may carry a refinement predicate) are all conservatively kept.

Keeping the condition costs only a redundant `ite`, so being conservative here is cheap.

## Tests

- `falsifiable/if_block_collapsed_cond_selector.lus` and `falsifiable/when_block_collapsed_cond_selector.lus` — both report `success` (exit 0) on `main` and `falsifiable` (exit 40) with this change.
- `success/when_block_collapsed_cond_selector_guarded.lus` — pins that keeping the condition does not turn into a false alarm: the inner `when` condition is kept, but its obligation is still discharged by the enclosing `Some?(p)` branch, so the guard context `mk_selector_obligation` builds is unaffected.

Full regression suite passes in all three slicing modes (1581 passed), as do the OUnit suites.

### DIFF
--- a/src/lustre/lustreAstHelpers.ml
+++ b/src/lustre/lustreAstHelpers.ml
@@ -181,10 +181,17 @@ let fold_label_or_index empty union f idx =
    LustreAstNormalizer), so an expression containing one has to be kept even
    where its value is irrelevant, or the obligation silently disappears.
 
-   The test is a whitelist: node calls, binders, pattern matching, and any
-   expression form added later are all reported as not droppable. The only
-   caller (LustreDesugarIfBlocks) uses the answer to decide whether it may
-   simplify away a branch condition, and keeping a condition is always sound. *)
+   The test is a conservative whitelist: only the forms matched below as
+   droppable are recognized as such, and every other form, including any added
+   later, is reported as not droppable. The only caller (LustreDesugarIfBlocks)
+   uses the answer to decide whether it may simplify a branch condition away,
+   and keeping a condition is always sound, so a too-strict answer costs no
+   more than a redundant 'ite'.
+
+   Only expressions are inspected. The types carried by the droppable forms
+   below (the type arguments of a record or ADT construction, and the type of
+   an abstract symbolic constant) are type instantiations: unlike a variable
+   declaration, they carry no refinement-type obligation of their own. *)
 let rec expr_is_droppable = function
   | Ident _ | ModeRef _ | Const _ | Last _ | AbstractSymConst _
   | EmptyMap (_, None) | EmptySet (_, None) -> true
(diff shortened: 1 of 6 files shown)


lorchrob:
Claude noticed 2 comments that do not quite match the code (pasted below). Once those are tweaked it looks good to me. 

1. .mli doc comment understates the check (lustreAstHelpers.mli:43)

▎ "…i.e. whether it contains neither a user-written ADT selector nor a node call"

The implementation also refuses Quantifier, Match, TypeAscription, AnyOp/ChooseOp, StructUpdate, ArrayConstr, IndexAccess, Merge/Condact/Activate/RestartEvery, and typed EmptyMap/EmptySet. Since this comment is the rule a future maintainer applies when adding a constructor, it should say it's a conservative whitelist rather than enumerate two cases.

2. The .ml comment's stated rule doesn't match the code (lustreAstHelpers.ml:203)

▎ "every construct whose operands include a type (which may carry a refinement predicate) are conservatively reported as not droppable"

But RecordExpr (type-param list), ADTTerm (type-arg list) and AbstractSymConst of position * lustre_type all carry types and are droppable. I checked this is harmless today: refinement obligations come from variable declarations (SubType[…]), not from construction expressions — R { x = k } with x: subrange[1,10] of int, and Some(k) at Opt<Pos>, produce no obligation in either the ITE baseline or the block form. So it's a comment/code mismatch, not a bug; fix the comment (or move the three).

lorchrob (commented):
See my earlier comment (otherwise it looks good)

PR #1485, opened 2026-09-04
Let the runtime keep the wall clock again on Windows

### LINKED ISSUE #1477: Windows: multi-second stop-the-world GC stalls delay --timeout (blocks removing the native backstop)
#1459 added a native Windows thread that ends the process `--timeout` plus sixty seconds in, because on Windows nothing else could: the supervisor's polling loop, which is where `--timeout` is enforced there, could stop running for minutes at a time.

That was [ocaml/ocaml#15028](https://github.com/ocaml/ocaml/issues/15028) — on Windows from OCaml 5.5.0, a thread that never yields keeps its domain's runtime lock and the other threads never run. Fixed by [ocaml/ocaml#15029](https://github.com/ocaml/ocaml/pull/15029), which is on the 5.5 branch for 5.5.1. See #1449 for the whole story.

**When 5.5.1 ships and CI moves to it, the backstop's motivating bug no longer exists on any supported compiler**, and it can be removed.

## What to remove

- `src/kind2_stubs.c` — the whole file, the tree's only C stubs
- `src/nativeTimeout.ml` / `.mli`
- the `foreign_stubs` stanza in `src/dune`
- the `-lkind2dev_stubs` entry that `scripts/gen-linking-flags.sh` puts at the head of the macOS static `CCLIB`
- the arming site in `src/kind2.ml`, including its ceiling check on `--timeout`

## What to keep

- The OCaml last resort in `kind2Flow.ml` (#1450, #1460, #1462). It is not Windows-specific and it guards a different failure — a teardown that wedges — which nothing upstream has fixed.

## Before removing

- Confirm the CI compiler is ≥ 5.5.1 on **every** platform, not just Windows, and that no supported build still uses 5.5.0.
- Re-run the Windows job with the backstop gone and check `test-issue-127.lus [slice_on]` comes in under the 84s budget rather than being killed at 204s. That test is the one this was all about, and it is the evidence either way.
- The elapsed seconds in the harness's timeout summary (`tests/conftest.py`) stay useful regardless — they are what distinguishes a run the supervisor ended from one something else killed.

## If it is kept instead

A defensible alternative is to keep it as a general guard against any future wedge, in which case the sixty second grace should be revisited: it was set from measurements of runs stalling under *this* bug (healthy runs finished 4.2s, 11.5s, 23.9s, 24.3s and 29.0s past their timeout), and those numbers will not describe a fixed runtime. The comment in `src/kind2.ml` explaining the grace would need to say so.

daniel-larraz:
Removes the native timeout backstop of #1459. Its motivating bug is fixed upstream and released. Closes #1477.

## Why it existed, and why it does not need to

On Windows there is no SIGALRM, so `--timeout` is enforced by a check inside the supervisor's polling loop — and that loop stopped running for minutes at a time. #1459 answered that with a native thread that ended the process `--timeout` plus sixty seconds in, because nothing written in OCaml could be relied on to run at all.

The cause was [ocaml/ocaml#15028](https://github.com/ocaml/ocaml/issues/15028): on Windows from OCaml 5.5.0, a thread that never yields voluntarily keeps its domain's runtime lock and the other threads of that domain never run. The backup thread's handle is closed right after creation, so the invalid handle reaching `caml_plat_thread_equal` came back as `-1`, which reads as true. Fixed by [ocaml/ocaml#15029](https://github.com/ocaml/ocaml/pull/15029), released in **5.5.1**, which also re-enables `testpreempt.ml` on Windows so it cannot come back quietly.

## Evidence that 5.5.1 carries the fix

The standalone reproducer ([repo](https://github.com/daniel-larraz/ocaml-windows-thread-starvation)), one domain and two threads, on `windows-latest`, with 5.5.0 in the same run as the control:

| | 5.2.1 | 5.3.0 | 5.4.0 | 5.5.0 | 5.5.1 |
|---|---|---|---|---|---|
| hangs | 0 | 0 | 0 | **3** | **0** |
| worst wait | 0.06s | 0.07s | 0.05s | — | **0.06s** |

[run](https://github.com/daniel-larraz/ocaml-windows-thread-starvation/actions/runs/33894638239)

## What goes

- `src/kind2_stubs.c` and `src/nativeTimeout.ml` — the tree's only C stubs, gone again
- the `foreign_stubs` stanza in `src/dune`
- `-lkind2dev_stubs` from the head of the macOS static `CCLIB` in `scripts/gen-linking-flags.sh`
- the arming site in `src/kind2.ml`, with the ceiling check on `--timeout` it needed

## What stays

The last resort in `kind2Flow.ml` (#1450, #1460, #1462). It is not Windows-specific, and it guards a teardown that wedges — a different failure, and not one anybody has fixed.

## The compiler floor moves to 5.5.1

In `dune-project`, which regenerates `kind2.opam` and `kmoxi.opam`, and in all four workflows.

This is the part that makes it a removal rather than a silent regression. #1477 asked only that CI be on 5.5.1, but CI says nothing about what users build with: a Windows build on 5.5.0 would go straight back to running past its timeout with nothing left to stop it. The floor is what prevents that, and 5.5.1 is a patch release of a compiler this project already required.

## Verified

- `dune build --profile strict` clean; `cd tests && ./run` — 555 passed in 36.7s.
- **The Windows job of this pull request is the test**: `test-issue-127.lus [slice_on]` has to come in under the 84s budget on its own, with nothing left to rescue it. That is the same test #1451 once had to make an exception for, and #1459 bounded at 144.1s.


daniel-larraz:
## Three samples of the Windows job, since one was not enough

`test-issue-127.lus [slice_on]`, the test this whole line of work was about, with nothing bounding it any more:

| attempt | elapsed | past the 84s budget |
|---|---|---|
| 1 | 159.1s | +75.1s |
| 2 | 144.9s | +60.9s |
| 4 | **87.6s** | +3.6s |

(Attempt 3 failed in `Build Kind 2` on `curl` exit 35 fetching opam sources — a network error, no sample.)

All three pass. The suite is green in each: 555 passed.

**What the spread says.** The first sample looked alarming on its own: every healthy overshoot measured while the bug was live sat between 4.2s and 29.0s, and 75s is well outside that. Two more samples make the honest reading a wide distribution rather than a residual defect — 87.6s is an ordinary run, and it rules out the idea that something still systematically delays the polling loop.

It also shows the backstop had been **censoring the upper tail**: anything past 144s was killed rather than measured, so the 144.1s figure quoted for #1459 was a kill, not a completion. This is the first look at what these runs actually want to take.

`quant_finite_nested_binders` was 101.8s, 105.6s and 96.6s across the same attempts, all inside its historical range.

**The residual risk, stated plainly.** The harness kills at `--timeout` + 120 = 204s. The worst sample here is 159.1s, so the margin is 45s. That is real but not alarming, and it is now visible rather than hidden: the elapsed seconds in the timeout summary (added in #1459 and kept here) mean any creep shows up in the log instead of arriving as a mystery kill.

If it does recur, the cheap fix is the harness's margin — `run_timeout` at `+120` was chosen when nothing bounded a wedged run and a hang had to be caught quickly. With the runtime fixed, a slow run is slow rather than stuck, and a larger margin costs only the time of a genuine failure. That would be a smaller change than reinstating a native thread to compensate for slowness that was never the bug.

PR #1489, opened 2026-09-08
Assume the one-state part of the transition relation at every state

daniel-larraz:
## What

The inductive step does not assert the transition relation at the first state of its path, so a variable defined from others is assumed there without its definition. A property abstracted into a state variable is assumed true while nothing ties it to what it stands for, and the step at a small `k` has a model that violates the definition. The step then fails on properties that are 1-inductive, and only succeeds at a larger `k`, if at all.

Every system now carries a one-state invariant of its own,

```
if init_flag then Init(x) else OneStateTrans(x)
```

where `OneStateTrans` is the conjunction of the conjuncts of the transition relation that only constrain the current state. Any reachable state that is not initial is the target of a transition and satisfies them; an initial state satisfies the initial state constraint. The existing invariant machinery asserts this at every bound of the step path, including the first, which recovers the definitions and every other stateless constraint there.

## How

`TransSys.one_state_conjuncts` selects the conjuncts of a transition relation that mention a single state. A subsystem called unconditionally contributes the same part of its own transition relation, with its formal parameters bound to the actual ones of the call; a call under a condition is left out, which is sound and only weaker.

The Lustre front end splits those conjuncts off before the let closures for inlined locals wrap the conjunction, and wraps them with the same closures, so the part it hands over mentions no variable the system does not declare. A conjunct can still cease to be one-state once those bindings are substituted into it, so the term is checked to be one-state rather than trusted, and dropped if it is not.

The invariant is registered with the certificate `(1, inv)`. That claim holds: the initial state constraint sets the init flag, so the then-branch reduces to the initial constraint itself, and the transition relation clears the flag and contains the one-state conjuncts, so the else-branch follows from the transition alone. Certificates generated for models with node calls check out, base case, inductive case and property subsumption all `unsat`.

## Effect

Every engine that reads the system's invariants sees it: BMC, k-induction, 2-induction, IC3QE, invariant generation, C2I, and the realizability and assumption-generation queries. IC3IA works from `init_of_bound` and `trans_of_bound` and does not read invariants, so it is unaffected.

On the Kondo protocol models, `kind2` with default options:

| Model | before | after |
| --- | --- | --- |
| Ring leader election | 16.6s, k=2 | 0.9s, k=1 |
| Simplified leader election | 0.7s, k=2 | 0.6s, k=1 |
| Lock server | 0.9s, k=2 | 0.5s, k=1 |
| Distributed lock | 0.2s, k=2 | 0.2s, k=1 |
| Client server | 0.4s, k=2 | 0.4s, k=1 |
| Two-phase commit | 1.3s, k=2 | 1.4s, k=2 |

Five of the six drop from 2-induction to 1-induction. Two-phase commit has a property that genuinely needs two steps and is unchanged.

## Tests

The regression suite passes (576). No test is added: the change strengthens what the step assumes rather than what it reports, so a pass/fail outcome does not separate it from `main`, and the suite has no way to pin a single engine for a test that would.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 # Unreleased
 
+Improvements:
+- The inductive step now assumes, at every state of its path, the part of the transition relation that only constrains one state. It previously had nothing to say about the first state of the path, where the transition relation is not asserted, so a variable defined from others was assumed there without its definition. Properties that are 1-inductive are now proved at k=1 rather than at a larger k, if they were provable at all.
+
 Breaking changes:
 - A user-written [ADT selector](https://kind.cs.uiowa.edu/kind2_user_doc/2_input/14_algebraic_datatypes.html#selectors) `e.f` now generates a proof obligation that the constructor owning `f` is active, reported as a property named `Selector[L<line>C<column>]`. Models that read a field without establishing its constructor will report a new failing property. The obligation is discharged by the short-circuiting operators (`and then`, `or else`, `==>`), `when ... then ... else` expressions, `when` blocks, and `match` arms; their eager counterparts (`and`, `or`, `=>`, `if` expressions and `if` blocks) do not discharge it.
 
--- a/src/lustre/lustreTransSys.ml
+++ b/src/lustre/lustreTransSys.ml
@@ -2293,7 +2293,14 @@ let constraints_of_arrays init terms eq_bounds =
 
     ) eq_bounds terms 
 
-let constraints_of_equations node init stateful_vars terms equations definition_set =
+(* The one-state part of a list of constraints, let-bound like them *)
+let one_state_term_of_conjuncts f terms bind_lets =
+  f terms |> Term.mk_and |> bind_lets
+
+(* [one_state] selects, among the constraints, those that only concern
+   the current state; if given, the conjunction of its result is returned
+   as well, wrapped in the same let bindings as the constraints. *)
+let constraints_of_equations ?one_state node init stateful_vars terms equations definition_set =
 
   (* make constraints for equations which do not redefine arrays first *)
   let terms, lets, eq_bounds, definition_set =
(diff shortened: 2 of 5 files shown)


daniel-larraz:
Fixed in 8b4f50cd.

The front end wraps the constraints of a node in the let bindings of all its stateless variables. A binding the constraints do not mention is harmless there, since it binds a variable that does not occur, but it is not harmless in the one-state part: the transition relation binds such a variable at the previous instant as well as the current one, the one-state conjuncts never read the previous one, and the binding survived into the invariant unused, carrying the previous state with it. Asserting the invariant at a bound then mentioned variables one instant before that bound, which are not declared there, hence `unknown constant %f112@-1`.

The one-state check did not catch it because the fold behind `Term.var_offsets_of_term` visits a binding only where the body uses the variable it binds, so a binding nothing reads is invisible to it while still being part of the term. The one-state part is now wrapped in the bindings it uses and no others, which also makes the check meaningful, since what is analysed is then what is emitted.

Checked locally against the same set (`FMCAD08/Bool`, `FMCAD08/Int`, `FMCAD08/Real_Int`, 859 models), head and base built from `05364f66`:

| | base | head |
| --- | --- | --- |
| Valid | 458 | 458 |
| Invalid | 345 | 345 |
| Timeout | 56 | 56 |
| Error | 0 | 0 |

Two benchmarks differ, `metros_4_e2_968_e5_991.lus` (Timeout to Invalid) and `metros_4_e3_1091_e4_232.lus` (Invalid to Timeout); both agree with the base when the wall limit is raised, so that pair is noise from the shorter limit I used locally. The regression suite passes at 576, and the Kondo models still prove at k=1 at the same cost.

daniel-larraz:
Added a test in dfd128e9: `tests/regression/success/one_state_part_of_trans.lus`.

It is a counter with a property stated over an output, so the property variable is not inlined and the step has to assume it at the first state of its path without its definition. The suite only checks that it stays provable. The difference is in the reported k, which is observed by hand with

```
kind2 --enable BMC --enable IND tests/regression/success/one_state_part_of_trans.lus
```

`counter is non-negative: valid (k=1)` on this branch, `valid (k=2)` on `main`. The two engines have to be named: under the default set another engine supplies what the step is missing and both report k=1. The header of the file says all of this, so it does not need to be rediscovered.

Suite is at 577 with it.

Issue #1468, opened 2026-08-27
Nested constructor pattern on a polymorphic ADT fails an assertion in lustreDesugarADTs


lorchrob:
A nested constructor pattern on a *polymorphic* ADT aborts with an assertion failure. The same nesting on a monomorphic ADT works.

### Reproducer

```lustre
datatype Opt<T> = Absent | Present (val: T);

node T (q: Opt<Opt<int>>) returns (y: int)
let
  y = match q with
      | Present (Present (z)) : z
      | Present (Absent) : 0
      | Absent : 0
      end;
tel

node main () returns ()
let check T(Present(Present(4))) = 4; tel
```

```
<Error> Error opening input file 'adt_poly_nested_constructor_pattern.lus':
        File "src/lustre/lustreDesugarADTs.ml", line 195, characters 17-23: Assertion failed
```

### Cause

`collect_pattern_constraints` (`src/lustre/lustreDesugarADTs.ml:189-195`) resolves a nested `Pat` sub-pattern by looking up the ADT info of the field's declared type:

```ocaml
        | LA.Pat (_, sub_name, _) ->
          (match adt_info_of_type ctx adt_map ftype with
          | Some sub_info when List.mem sub_name sub_info.ctor_variants ->
            let (c, s) = collect_pattern_constraints pos ctx adt_map sub_info field_expr sub_pat in
            (conds @ c, subs @ s)
          | _ -> assert false)
```

For `Opt<Opt<int>>`, the declared type of `Present`'s field `val` is the type parameter `T`, not an ADT name, so `adt_info_of_type` returns `None` and the catch-all asserts. The type argument is not substituted into `ftype` before the lookup, so the nested pattern is never resolved to `Opt`.

Nesting one level deeper or wider makes no difference; a single-level pattern on the same polymorphic type is fine, as is `Present (Present (z))` on a monomorphic `datatype Opt = Absent | Present (val: OptInt)`.

lorchrob:
The bug is wider than this reproducer suggests: nesting the polymorphic datatype inside itself is not needed, a constructor pattern in *any* type-variable field position is enough.

```lustre
datatype Opt<A> = Nn | Ss (val: A);
datatype B = T | F;
node N(x: Opt<B>) returns (y: int);
let y = match x with | Nn : 0 | Ss(T) : 1 | Ss(F) : 2 end; tel
```

[...]

PR #1490, opened 2026-09-08
Pattern match exhaustiveness and redundancy checks

lorchrob:
See Luc Maranget, "Warnings for pattern matching" (2007)

This PR required some updates to the type checker. Earlier, we noticed that it is useful for the type checker functions to return updated ASTs as output, so they can e.g. insert information about inferred types directly in the AST. In this PR I extended this same type of update to more functions in the type checker (particularly in functions that process globals) as this extra type information is needed for exhaustiveness and redundancy checks.


daniel-larraz:
## Review

Nice work. The `U_rec` implementation looks faithful to the paper, the AST traversal is unusually thorough, and the type checker refactor appears behavior-preserving. I did not find a correctness defect in the new checker itself.

<details>
<summary><b>What I checked empirically</b> (built the branch and compared against <code>main</code>)</summary>

* Front-end output is byte-identical to `main` on all 1170 existing `.lus` files under `tests/`.
* All ounit tests pass (`dune build @runtest`), and CI is green.
* The checker gives the right answer on nested patterns, type synonyms at both scrutinee and field positions, recursive datatypes, polymorphic datatypes instantiated at other datatypes, and depth-3 nesting.
* I probed ~25 syntactic positions where a `match` can hide (contract assumes/guarantees/modes, ghost vars, frame and `when` blocks, `any`, node and contract call arguments, array indices, record expressions, `pre`/`->`, properties, type arguments). Every one is caught, so the traversal looks complete.
* No performance concern: a 1024-arm match over a 10-field product checks in 0.11s. Triggering the exponential worst case seems to require a source file that is itself exponentially large.
* Constructor names are globally unique, so the name-based signature comparison in `complete_signature` is sound.
* I temporarily instrumented the `assert false` branches and the `scrut_ty_opt = None` branch: across all of the above, none of them was ever reached.

</details>

---

### 1. Polymorphic nested pattern still aborts the desugarer

A constructor pattern nested under a field whose **declared** type is a type variable hits `assert false` in `desugarADTs`:

```lustre
datatype B = T | F;
datatype Opt<A> = None | Some (val: A);
node N(x: Opt<B>) returns (y: int);
let y = match x with | None : 0 | Some(T) : 1 | Some(F) : 2 end; tel
```

```
<Error> Error opening input file 't.lus':
        File "src/lustre/lustreDesugarADTs.ml", line 260, characters 17-23: Assertion failed
```

**This is pre-existing** (`main` crashes identically), so it is not a regression from this PR. I am raising it here because the PR touches exactly this area:

* The comments at [`build_ite`](https://github.com/kind2-mc/kind2/blob/a0bea189/src/lustre/lustreDesugarADTs.ml#L281-L288) are updated to say the new pass now guards those assertions, which invites the reading that the desugarer's assertions are generally covered. This one is not.
* `tests/regression/success/match_polymorphic.lus` narrowly misses it: in `MkW(S(v))` the inner sub-pattern `v` is a variable, so it takes the `VarPat` branch and never calls `adt_info_of_type`.
* The new exhaustiveness check now *masks* the crash whenever the match also happens to be non-exhaustive (e.g. dropping the `Some(F)` arm above reports a clean error instead), which makes the remaining crash harder to notice.

Root cause is that `info.ctor_fields` holds the declared field type (`A`) rather than the instantiation (`B`), so [`adt_info_of_type ctx adt_map ftype`](https://github.com/kind2-mc/kind2/blob/a0bea189/src/lustre/lustreDesugarADTs.ml#L256-L260) returns `None`. Happy to split this into its own issue if you would rather keep it out of this PR.

### 2. `matches_of_type` duplicates `fold_lustre_ty`, and the difference is interesting

[`matches_of_type`](https://github.com/kind2-mc/kind2/blob/a0bea189/src/lustre/lustreCheckMatchExpressions.ml#L214) is `LH.fold_lustre_ty matches_of_expr [] (@)` case for case, with one exception: it descends into `UserType` type arguments.

That exception is the valuable part. Both [`fold_lustre_ty`](https://github.com/kind2-mc/kind2/blob/a0bea189/src/lustre/lustreAstHelpers.ml#L82) and `map_lustre_ty` return `init`/`ty` unchanged for `UserType _`, dropping its type arguments. Your own `non_exhaustive_in_type_argument.lus` shows those arguments carry expressions worth visiting, which means the helper's other callers miss them too:

* `check_clocks` folds over ty_args with `fold_lustre_ty` (`lustreTypeChecker.ml:423-452`), and ignores `Call` ty_args entirely at line 437.
* `lustreCheckADTDecreases.ml:128`, `:223`, `:314` use it to find node calls inside types.

Extending `fold_lustre_ty` to descend into `UserType` ty_args would close the same gap in those callers and let this function collapse to a one-liner.

### 3. Type arguments re-walked to recover warnings that were just dropped

[`check_instantiated_ty_args`](https://github.com/kind2-mc/kind2/blob/a0bea189/src/lustre/lustreTypeChecker.ml#L3084) re-runs `check_type_well_formed` over every type argument only to recover warnings the `UserType` branch already computed as `warnings0` and then discarded, as the comment at lines 3210-3212 acknowledges. Each explicitly instantiated `RecordExpr`/`ADTTerm` type argument ends up checked two or three times, and the re-walk passes `is_nested=false` regardless of the `arg_nested` decision made just above.

I confirmed the observable warning behavior is unchanged from `main` (a parameter mentioned twice reports twice, a phantom parameter loses the warning, identically on both branches), so this is duplicated work rather than a behavior change. Fixing the drop at the source would let this function just destructure the result of the single existing call.

### 4. `rloi` duplicates the exported `fold_label_or_index`

The local helper at [`lustreCheckMatchExpressions.ml:175`](https://github.com/kind2-mc/kind2/blob/a0bea189/src/lustre/lustreCheckMatchExpressions.ml#L175) is a verbatim copy of `LH.fold_label_or_index` (`lustreAstHelpers.ml:173`, exported). `lustreCheckADTDecreases.ml:210` already calls it as `LH.fold_label_or_index [] (@) go idx`. If a new `label_or_index` constructor is added later, the shared helper gets updated and this private copy does not, so matches inside that new index kind would silently escape the check.

### 5. New `ADT` case hand-rolls the `rlist`/`join` idiom twice

In [`syn_type_equal`](https://github.com/kind2-mc/kind2/blob/a0bea189/src/lustre/lustreAstHelpers.ml#L2147), the length-check + `List.map2` + `join` shape is spelled out by hand for the constructor list and again for each field list. The local `rlist` helper (line 2085) and the adjacent `RecordType` case (line 2111) already encode it, so there are now three copies in one function; a future change to one (say, treating a length mismatch as `Error` rather than `Ok false`) has to be replicated in each.

Behavior is correct as written. I checked specifically that recursive datatypes do not diverge here, since the `UserType` case compares names and arguments without expanding. Worth keeping in mind that most callers pass `depth_limit = None`, so there would be no bound if that ever changed.

### 6. Minor: non-exhaustiveness error does not name a witness

`IncompletePatternMatch` reports only "cases are not exhaustive" at the position of the whole match. For a wide datatype or nested patterns over a product, the user has to work out the missing case by hand. Maranget's algorithm I (§3.2) gets a witness out of the `U_rec` already implemented here with a fairly small change, and the neighbouring `RedundantPattern` already prints the offending pattern, so it would match the module's own standard. This could be implemented in a follow-up PR.

---

<sub>Reviewed with the help of Claude Code. Findings verified against a local build of a0bea189; the repros above are reproducible from a clean tree.</sub>

lorchrob:
Here is how each point was handled. 

### 1. Polymorphic nested pattern still aborts the desugarer — left out of this PR

Confirmed pre-existing (`main` crashes identically), so it stays out of this PR. Already tracked in #1468.

The `build_ite` comments were left as written. They annotate the two assertions at `lustreDesugarADTs.ml:286` and `:288`, and for those two the claim is accurate — a catch-all arm is now guaranteed to be last, and the last arm is now guaranteed to cover what is left. The assertion in `adt_info_of_type` is a different one and is not claimed to be covered by either comment.

### 2. `matches_of_type` duplicates `fold_lustre_ty` — fixed

### 3. Type arguments re-walked to recover dropped warnings — comments corrected, structure kept

I didn't feel updating the code to allow for only a single pass was worth the (code) complexity hit.

### 4. `rloi` duplicates the exported `fold_label_or_index` — fixed

### 5. New `ADT` case hand-rolls the `rlist`/`join` idiom twice — fixed

### 6. Non-exhaustiveness error does not name a witness — follow-up

Agreed, and left for a follow-up PR as you suggest.

PR #1506, opened 2026-09-10
Array element struct update bug fixes

lorchrob:


src/lustre/lustreArrayDependencies.ml
```diff
@@ -174,9 +174,10 @@ and process_expr ind_vars ctx (ns:AD.node_summary) proj indices expr =
   )
   | GroupExpr (_, _, es) ->
     es |> (List.map r) |> (List.fold_left union_ empty_)
-  (* Update of structured expressions *)
-  | StructUpdate (_, e1, _, Some e2) -> union_ (r e1) (r e2)
-  | StructUpdate (_, e1, _, _) -> r e1
+  | StructUpdate (_, e1, idx, e2) ->
+    let idx_graph = AH.fold_label_or_index empty_ union_ r idx in
```
daniel-larraz, line 178: **New false cyclic-dependency error.**

Folding the index list with `r` carries the enclosing accumulated `indices` and `ind_vars`, so a read appearing inside the index is attributed to the slot currently being defined. That makes a legal definition fail:

```
node top(a: int^3) returns (y: int^3);
let
  y[i] = (a[y[i-1] := 1])[i];
tel
```

This is now rejected with `Cyclic dependency detected in definition of identifiers: y 0 -> y 0`, while the equivalent read `y[i] = a[y[i-1]]` is still accepted. Reverting only this hunk makes the model compile again, so the regression is isolated to it.

The index is evaluated independently of the slot being written, so it should be processed in a reset index context rather than with `r`, which inherits `indices` from the surrounding update.

src/lustre/lustreAstNormalizer.ml
```diff
@@ -2867,14 +2867,29 @@ and normalize_expr ?guard info (node_id : NI.t option) map =
       (* The MapIndex and SetIndex cases are handled specially *)
       then normalize_expr ?guard info node_id map (StructUpdate (pos, expr1, i, expr2)) 
     else 
+      (* An index expression is a subexpression like any other: it may hold a
+         node call or a read that later passes require to have been abstracted *)
+      let normalize_idx ?guard idx = match idx with
```
daniel-larraz, line 2872: **The node-call abstraction in an index is incomplete.**

`normalize_idx` correctly abstracts a node call appearing in the index, but `expand_node_calls_in_place` (lines 2963-2964) still passes the index list `u` through untouched:

```ocaml
| StructUpdate (p, e1, u, Some e2) -> A.StructUpdate (p, r e1, u, Some (r e2))
| StructUpdate (p, e1, u, None) -> A.StructUpdate (p, r e1, u, None)
```

So inside an array definition the abstraction is left referring to the array index variable. This input:

```
node Id(n: int) returns (o: int);
let o = n; tel

node top(x: int^3; v: int) returns (y: int^3);
let
  y[i] = (x[Id(i) := v])[i];
tel
```

normalizes to `y[i] = x[1_call := v][i]` and then aborts with `lustreNodeGen.ml:1189: Assertion failed`.

This is not a regression, since the same input asserted at `lustreNodeGen.ml:1893` before the change. But it is exactly the case this hunk sets out to fix, and `tests/regression/success/struct_update_index_expr.lus` covers the call-in-index only outside an array definition, so the gap is not caught by the new tests.

daniel-larraz (commented):
Reviewed the branch against `main`: 7 source files plus tests. I built it and exercised it with the full `tests/regression` success/falsifiable/error suites, `dune build @runtest`, `--smt_arrays true`, `--slice_nodes off/experimental`, and a set of hand-written adversarial models covering nested and chained updates, 2-D arrays, records, tuples, maps, sets, array literals and constructors, global const arrays, `pre`/`->`/`ite` contexts, frame conditions, and soundness counterexamples.

The core rewrite holds up well. Always emitting `E.mk_store` and expanding `select (store a k v) i` in `Term.push_select` behaved correctly on everything I tried, including out-of-bounds indices, temporal indices, and the nested updates the type checker previously rejected outright. Three specific things I checked and found fine: no store terms reach the solver under the default encoding (verified with `--smt_trace`); the `type_of_store` relaxation only widens within existing subtype relations and still rejects genuine mismatches; and the `lustreAstDependencies` change does not produce the false positive described below, because array definitions are exempt at that level.

Two findings inline.

src/lustre/lustreArrayDependencies.ml
lorchrob: Fixed.

Resetting only `indices` wasn't enough: by the time the update is reached the enclosing `[i]` read has already consumed `i` from `ind_vars`, so `y[i]` in the index no longer matches an index layer and `inductive_array12.lus` stops being detected as a cycle. `process_expr` now carries the equation's own index variables alongside the partially consumed ones, and an index is processed from those, with an empty index path and `proj = 0` — the same context the right-hand side starts in.

The new value has the same problem — `y[i] = (a[0 := y[i-1]])[i]` is rejected on main too — but it's pre-existing and outside this PR, so I filed it as #1516.

src/lustre/lustreAstNormalizer.ml
lorchrob: Fixed — `expand_node_calls_in_place` now recurses into the index expressions. Your reproducer is covered by a new node in `struct_update_index_expr.lus`.

A map or set update in that same position still fails, but it does so with or without a call in the key, so it's a separate pre-existing gap: #1517.

PR #1507, opened 2026-09-10
Define contract-less and transparent recursive functions with define-funs-rec

daniel-larraz:
## Summary

A recursive function is compiled by unrolling its definition once and abstracting its recursive calls by its contract, with the outputs of every instance tied to an uninterpreted functional symbol `f.out.__function_of_inputs`. A function without a contract is therefore unknown to the solver past its first unrolling: `Fact(4) = 24` cannot be established.

This PR gives that symbol a definition. When a recursive function has **no contract to abstract it with** (no guarantee or mode, explicit or from a refinement type of an output; assumptions and input types are obligations of the callers and do not count) or is declared **`transparent`**, its functional symbols are defined with an SMT-LIB `define-funs-rec` block built from the body of the function (new module `LustreFunDefs`), so that the solver unfolds the definition. The unrolling and the cutoff instance are kept as they are; only the symbol the cutoff is tied to changes from declared to defined.

- In the definition, a call of the same recursive group applies the callee symbol under the guard of the call's termination checks (and its context), and yields an arbitrary value `f.out.__undefined_call(args)` otherwise. The definitions are thus well-founded whether or not the measure actually decreases, always have a model, and cannot make the transition system inconsistent; the termination checks remain properties (see the new `falsifiable/rec_def_nonterminating.lus`).
- A recursive function of another group and an imported function are applied through their functional symbol (defined by its own block, or uninterpreted and tied to its instances); any other function is inlined. Every function of a group is defined, or none is; a group is left alone if a body is not a total function of its inputs (assertions, oracles, arrays, activation conditions).
- The blocks are carried by `TransSys.t` (`fun_defs`) in dependency order and emitted by `define_and_declare_of_bounds` / `declare_sorts_ufs_const` through a new `~define_rec` callback (`SMTSolver.define_funs_rec`, `define-funs-rec` in `SMTLIBSolver`); the certificate export writes them as well.
- A new logic feature `RF` marks such systems and prints the logic as `ALL`: both solvers refuse the definitions in a `QF_` logic, and Z3 accepts them under an erratic subset of the named logics only (`UFLIA` and `AUFNIA` yes, `UFNIA` and `AUFLIA` no).
- Z3's relevancy propagation diverges on the unfolding (the base case of `even_odd_sum_props.lus` runs for minutes with it and takes 0.4s with `smt.relevancy=0`), so the Z3 driver passes that option for such systems only.
- IC3IA turns itself off on these systems, as the interpolating solvers do not take the definitions (IC3QE already did for any function symbol).
- New option `--define_fun_rec <bool>` (default `true`); the encoding is only used with Z3 and cvc5.

Documentation (`lustre.md`, "How recursive functions are analyzed") and `CHANGES.md` are updated.

## Test plan

- [x] `dune build --profile release @runtest` (OUnit) passes
- [x] Full regression suite: 584 passed, no timeouts (Z3)
- [x] All 40 regression tests involving recursive functions give the same verdict with `--define_fun_rec false` as before
- [x] 7 new regression tests (`rec_def_*.lus`): concrete values of contract-less factorial, mutual even/odd, ADT sum, transparent function with contract, inlined helper, uninterpreted callee of another group; a non-terminating function is falsified rather than vacuously valid. They pass with both Z3 and cvc5, and (except the falsifiable one) fail with the previous encoding, as they need the definition.
- [x] `--certif true` on a defined function: the certificate replays as valid with Z3 and cvc5


daniel-larraz:
## Code review

Reviewed `25e7bb23..98f894e4`. Two findings, both reproduced by running the built binary.

### 1. Soundness: `transparent` + contract ⇒ vacuous proofs

[`lustreFunDefs.ml:122-124`](https://github.com/kind2-mc/kind2/blob/98f894e4666c29bc3935c4c699d8524fcd9bac2b/src/lustre/lustreFunDefs.ml#L122-L124) — `eligible` admits a transparent function *that has a contract*:

```ocaml
let eligible node =
  not (N.has_effective_contract node)
  || node.N.opacity = Opacity.Transparent
```

For that case `lustreTransSys` does two contradictory things to the same functional symbol: it defines it exactly with `define-funs-rec`, **and** still asserts the contract abstraction at the recursion cutoff ([`lustreTransSys.ml:2896`](https://github.com/kind2-mc/kind2/blob/98f894e4666c29bc3935c4c699d8524fcd9bac2b/src/lustre/lustreTransSys.ml#L2896) → `abstraction_of_contract`). When the guarantee is not actually valid, the two assertions contradict, the transition relation becomes unsatisfiable, and every property is proved vacuously.

Reproducer — a transparent `F(n) = n` carrying a false guarantee:

```
transparent function rec F(n:int) returns (f:int);
(*@contract guarantee "g" f < 10; decreases n; *)
let f = when n <= 0 then 0 else 1 + F(n-1); tel

node main(x:int) returns (y:int);
let
  y = F(20);
  check "p" y = 999;
tel
```

| | `g` | `p` (`F(20) = 999`) |
|---|---|---|
| default | **valid** | **valid** |
| `--define_fun_rec false` | invalid | invalid |

Both verdicts are wrong under the default: `F(20)` is 20, so `g` is false and `p` is false. Dropping `transparent` also gives the correct answer.

`guard_rec_call` does not help here — its guard is the termination checks, which genuinely hold in this example (`decreases n`). The comment at [`lustreTransSys.ml:979-992`](https://github.com/kind2-mc/kind2/blob/98f894e4666c29bc3935c4c699d8524fcd9bac2b/src/lustre/lustreTransSys.ml#L979-L992) anticipates exactly this vacuity hazard for an *unsatisfiable* guarantee; the definition extends it to a merely *false* one.

Suggested fix: stop doing both. For a transparent function with a contract, either the cutoff should not assert the abstraction (the definition already pins the symbol down), or such functions should be excluded from `eligible`. The guarantees need to remain proof obligations either way.

### 2. An explicit `--smt_logic` disables every engine

[`lustreFunDefs.ml:44-48`](https://github.com/kind2-mc/kind2/blob/98f894e4666c29bc3935c4c699d8524fcd9bac2b/src/lustre/lustreFunDefs.ml#L44-L48) — `enabled ()` checks the flag and the solver, but not `Flags.Smt.logic ()`. With an explicit `--smt_logic`, `mk_trans_sys` produces `` `SMTLogic _ `` rather than `` `Inferred _ ``, so the new `RF` feature is never recorded, and [`termLib.ml:296`](https://github.com/kind2-mc/kind2/blob/98f894e4666c29bc3935c4c699d8524fcd9bac2b/src/terms/termLib.ml#L296) emits the named logic verbatim instead of forcing `ALL`:

```
$ kind2 --smt_logic QF_UFLIA tests/regression/success/rec_def_even_odd_value.lus
<Error> Caught failure in invariant generator: SMT solver failed:
        "line 52 column 8: logic does not support theory recfun"
... every property: unknown
```

BMC, inductive step, 2-induction and the invariant generators all fail; the same file completes normally with `--define_fun_rec false`. The same `` `Inferred ``-only match also makes the new Z3 `smt.relevancy=0` workaround silently inert under an explicit logic ([`Z3Driver.ml:55-59`](https://github.com/kind2-mc/kind2/blob/98f894e4666c29bc3935c4c699d8524fcd9bac2b/src/smt/Z3Driver.ml#L55-L59)).

Simplest fix: have `enabled ()` return `false` when a non-`ALL` logic is explicitly set.

### What checked out

Everything else probed behaved correctly: all 7 new `rec_def_*` tests and the 33 pre-existing `function rec` tests pass with both Z3 and cvc5; block dependency ordering and the declaration-before-definition ordering in `TransSys.define_and_declare_of_bounds` / `declare_sorts_ufs_const` are sound; the let-nesting in `body_of_output` preserves shadowing when a helper is inlined twice; const inputs, record inputs, multiple outputs, global constants and imported callees all work; the certificate path emits `define-funs-rec` after the uninterpreted declarations and `certificate_checker z3` passes; the `%f<N>` state-var naming cannot collide with the `X<n>` de Bruijn let names; `Term.map` descends into let bodies, so `ufs_of_term` does not miss applied symbols; and the IC3IA and IC3QE opt-outs fire correctly.

daniel-larraz:
Both findings were real and are fixed in 45fe3873 and 484ad4c8. Thanks — the first one was a soundness hole.

### 1. `transparent` + contract ⇒ vacuous proofs

Fixed by not using a defined function's contract as an abstraction anywhere, which is the first of the two options you suggested.

The two assertions cannot coexist: a definition says exactly what the function is, so assuming its guarantees on top of it adds nothing when they hold of the definition and contradicts it when they do not. Tying the cutoff's output to a fresh uninterpreted symbol instead does not help either — the body equation of the expanded instance plus the definition force that symbol back to `F(a_cutoff)`, so the contradiction returns.

`use_contract_as_abstraction` is now `(reached_limit || param_scope_is_abstract) && not is_defined`. Your reproducer:

| | `g` | `p` |
|---|---|---|
| before | valid | valid |
| now | **invalid** | **invalid** |
| `--define_fun_rec false` | invalid | invalid |

The consequence, which is in `CHANGES.md` under breaking changes and in the user doc: a `transparent` recursive function is verified from its body alone, so its guarantees no longer serve as the induction hypothesis of the recursion. One that needs induction over the recursion (a guarantee over a symbolic argument) goes from valid to unknown; the remedy is to drop `transparent`. Nothing changes for a contract-less function, which has no guarantee to assume, or for a function that is not defined.

`falsifiable/rec_def_transparent_false_guarantee.lus` pins this down. I checked it against the pre-fix binary: it reported everything valid and exited 0, so it would have failed.

### 2. An explicit `--smt_logic` disables every engine

Fixed in `enabled ()`, which now also requires the logic to be inferred.

I went further than "non-`ALL`": `--smt_logic none` is equally broken, though differently. It prints as `ALL` so the solver accepts the definitions, but `mk_trans_sys` produces `` `None `` rather than `` `Inferred _ ``, so the `RF` feature is absent and the Z3 `smt.relevancy=0` workaround stays inert — `--smt_logic none` on `even_odd_sum_props.lus` took the whole 84s budget and returned unknown. Since only an inferred logic can carry `RF`, the condition is now simply `Flags.Smt.logic () = `detect`.

Verified on `even_odd_sum_props.lus` and `rec_def_even_odd_value.lus` with `none`, `ALL` and `QF_UFLIA`: each completes in 0.2s with the previous encoding, no solver error and no divergence.

### 3. cvc5 hangs on every counterexample search (found while testing the above)

Not in your review, but the new falsifiable test surfaced it. With the flags Kind 2 passes (`--mbqi --full-saturate-quant`), cvc5 does not terminate on a **satisfiable** query whose context holds a recursive definition — even one that never applies the function:

```smt2
(set-logic ALL)
(define-funs-rec ((f ((n Int)) Int)) ((ite (<= n 0) 0 (+ 1 (f (- n 1))))))
(declare-fun y () Int) (assert (> y 3)) (check-sat)
```

hangs; plain `cvc5` answers `unknown`. So every BMC query on such a system would hang under cvc5. The existing tests missed it because they are all valid-shaped, and `rec_def_nonterminating` falsifies through the `__undefined_call` branch without unfolding anything.

45fe3873 adds `--fmf-fun` (cvc5's bounded finite model finding for recursive functions) when the logic carries `RF`, mirroring the Z3 workaround. The query above answers `sat` with it, and the queries that were answered `unsat` still are. All 8 `rec_def_*` tests now give the expected verdict under cvc5 as well as Z3.

### Checks

585 regression tests pass, OUnit passes, `dune build @doc-private` passes, and all 50 files mentioning `decreases`, `lemma` or `transparent` give the expected verdict.

PR #1510, opened 2026-09-11
Primer

lorchrob:
Extend the primer to full language


daniel-larraz (changes requested):
Can you please split this PR into two: one with the relocation of `lustre-primer.md`, and another with the rest of the changes?

lorchrob:
I applied the requested changes; after rebasing with `main` this PR now only includes "the rest of the changes"

doc/content/docs/lustre-primer.md
```diff
@@ -588,6 +839,220 @@ let
...
+  | <cond_1>:
+     <equations>
+  | <cond_2>:
+     <equations>
+  otherwise:
+     <equations>
+end
+```
+
+In both, only the selected branch is evaluated, with the consequences
```
daniel-larraz, line 1009: This restriction appears to be out of date: lazy branches now support temporal operators and node calls (see `abstract_temporal_branch` in `src/lustre/lustreGenNodes.ml` and the lazy-guard handling of node calls in `src/lustre/lustreSyntaxChecks.ml`). The next paragraph also relies on it, since it explains what `pre x` means inside a lazy branch.

I suspect this was copied from the user documentation, which is itself outdated. `doc/content/docs/inputs-and-outputs/lustre.md` still lists these restrictions for `when ... then ... else ...` expressions, for the short-circuit operators, and under "Current restrictions for `when` blocks". Its `last x` warning also describes `init_x -> pre x` inside a `when` branch, so it contradicts itself. It might be worth updating those sections too, either in this PR or in a follow-up.

doc/content/docs/lustre-primer.md
```diff
@@ -588,6 +839,220 @@ let
...
+node Top() returns (y1: int; y2: bool);
+let
+  y1 = SafePre@<int>(0);
+  y2 = SafePre@<bool>(false);
+tel
+```
+
+The `@<...>` instantiation at the call site is usually optional: Kind 2 infers
+the type arguments *bottom-up*, by matching the declared parameter types
+against the types of the actual arguments, so `SafePre(x1)` (without the annotation)
```
daniel-larraz, line 899: There's no `x1` in this example. Did you mean `SafePre(0)`?

doc/content/docs/lustre-primer.md
```diff
@@ -588,6 +839,220 @@ let
...
+let
+  l = any { x: int | x mod 2 = 1 };
+  z = y + l;
+tel
+```
+
+Here `l` is some odd stream, with no further commitment as to which — it may
+take a different odd value at every step.
+
+The variant `choose { ... }` is similar to `any { ... }`, but functional
```
daniel-larraz, line 1038: "given the same type as input" isn't quite right. `choose` is functional in the free variables of its predicate: it returns the same value whenever those variables have the same values. Maybe something like: "(that is, it always produces the same value for the same values of the variables occurring in its predicate)".

doc/content/docs/lustre-primer.md
```diff
@@ -620,11 +1085,274 @@ let
...
+
+```bash
+kind2 --enable CONTRACTCK <file>.lus
+```
+
+Kind 2 also checks the realizability of a node's *environment*---that the
+assumptions themselves can be met. Assumptions that
+no input sequence can satisfy make a node's guarantees vacuous. 
+
+When a contract is found unrealizable, `--print_deadlock` shows a trace ending
```
daniel-larraz, line 1342: `--print_deadlock` is enabled by default (`src/flags.ml`), so the trace is shown without any extra flag. As written, the sentence suggests you need to pass the flag. Maybe: "When a contract is found unrealizable, Kind 2 shows a trace ending in … (disable with `--print_deadlock false`)."

doc/content/docs/lustre-primer.md
lorchrob: Should these restrictions also be lifted for the lazy boolean operators (`and then`, `==>`, `or else`)? I see our frontend is still applying them [here](https://github.com/kind2-mc/kind2/blob/main/src/lustre/lustreSyntaxChecks.ml#L1375)

daniel-larraz: Yes, but removing the frontend check alone is only enough for node calls. `pre`/`->` would also need a change in node generation. The user docs (`lustre.md`, section "Short-circuit Boolean operators") list the same outdated restrictions and would need updating along with the code.

lorchrob: We will address these changes (primer updates, user doc updates, code changes) in a follow-up PR

doc/content/docs/lustre-primer.md
lorchrob: Addressed

doc/content/docs/lustre-primer.md
lorchrob: Addressed

doc/content/docs/lustre-primer.md
lorchrob: Addressed

daniel-larraz (approved):

PR #1513, opened 2026-09-11
State the assumption history of a call as a silent candidate invariant

daniel-larraz:
## Summary

The guarantees of a node abstracted by its contract are asserted under its "sofar" flag, the history of its assumptions. Proving an assumption obligation of the callee therefore needs those guarantees, which need the flag, which needs the obligation to have held at every earlier step. k-induction sees the flag as a free variable at the start of its window and cannot close that loop; only the one-state invariant generator, when it discovers that the flag is invariant, does. A node called on its own previous output, `state = Step(init -> pre state)` with `assume Valid(state); guarantee Valid(state2)` on `Step`, is the typical shape: the top-level `check Valid(state)` is not k-inductive for any k, and when IC3 is out (quantifiers over sets, functional congruence instances) and the invariant generator has nothing to work with, it is never proved.

Two commits:

1. **State the assumption history of a call as a candidate invariant.** The flag is stated as a candidate property of the caller, proven before it is used, so that the history is visible to k-induction. This is the treatment the when-block ties already receive for the hold semantics of a clocked call. The candidate is a property of the callee lifted into the caller, so it joins the other lifted properties of the call and goes through the same instantiation and guarding as they do, including the clock of a clocked call. It does not always suffice: where the assumption obligation and the guarantee are mutually dependent through the callee's own state, the candidate is itself not k-inductive and is proven only once the invariant generator has delivered.

2. **Silence the candidate properties Kind 2 generates for itself.** Where a callee's assumptions are genuinely violated, the candidate above is disproved as well and Kind 2 printed a `Candidate ... disproved` warning next to the assumption failure it already reports. The `Candidate` property source now carries a `report` flag, and the three logging entry points (`log_proved`, `log_unknown`, `log_cex`) check it, so all output formats behave alike: a silent candidate produces nothing whether it is proved, disproved or unknown. Every candidate Kind 2 generates for itself is silent (the when-block ties, the assumption histories, the internal candidates of IVC/MCS computation and certificate checking); their failure is reported through the properties they support. A `:candidate` property of a native input file is still reported.

## Measurements (Z3 4.16.0)

- A five-line model `s = N(0 -> pre s); check s >= 0` with `assume x >= 0; guarantee y >= 0` on `N`, compositional, BMC and k-induction only: on `main` it is still open after 25000 steps; on this branch it is valid by 1-induction.
- A security-protocol model with sets of ADT messages and quantified invariants (compositional and modular, every node contract `assume Valid(state); guarantee Valid(state2)`, top node calling `Step` on its own previous output): on `main` the top-level property is open after 34 steps at a 170s timeout; on this branch it is valid by 2-induction in 0.4s and the whole model proves in 30s.
- The model in the first commit's original measurement (an `any` carrying an assumption from a refinement type) has its obligations drop from k=3 to k=2.

## Test plan

- [x] `dune build --profile release @runtest` (OUnit) passes
- [x] Full regression suite: 595 passed, no test reaching its time budget (each commit)
- [x] With a violated callee assumption, the output has the assumption and property failures only, no candidate warning
- [x] Plain text, XML and JSON outputs contain no candidate property; the JSON output parses; `isCandidate` remains on the real properties


daniel-larraz:
Reviewed both commits. `dune build src/kind2.exe` is clean and the full regression suite (595 tests) passes. I also exercised the new candidate on concrete and abstract (imported) callees, self-feedback models (`a = 0 -> pre v`), violated assumptions, `condact`, `restart every`, two calls on one line, and array-indexed calls: the candidate is silent in both plain-text and XML, and a violated assumption is still reported through the assume property.

Two findings.

### 1. Duplicate candidate names when a call expands into several instances

`src/lustre/lustreTransSys.ml:1062`

The candidate's name is built from the call position alone:

```
"Assumptions of call at l%dc%d have held so far"
```

A call that expands into several instances (an array-indexed call, for example) then yields several candidates with identical names over *different* `sofar` variables — I saw three properties named `Sub[L14C10].assume[L3C3]` in one run. `TransSys.set_prop_status` matches by name and marks all of them, but only the first one's term becomes an invariant, so instances 2..N are marked decided without their history ever having been established.

Adding a per-instance disambiguator to the name would fix it — the lifted state variable (`lift_state_var state_var_map_up sofar_assump`, already computed two lines below for the `Generated` source) rather than the callee-side identifier. Note the neighbouring when-block tie candidates are not a precedent here: their `'%s'` is the tie's source identifier, which separates the several ties of one call from each other but would repeat across expanded instances just the same.

### 2. `PropertyNotFound` newly reachable in `F_relay`

`src/kEvent.ml:1517`

`reports` calls `TransSys.property_of_name` before the log-format dispatch, so the lookup now runs in `F_relay` mode too, where `log_proved` / `log_unknown` / `log_cex` were previously complete no-ops. The `StepCex` branch at `src/kEvent.ml:2412` calls `log_cex` outside any handler, unlike every surrounding `PropStatus` branch, which all catch `TransSys.PropertyNotFound`. A message about a property absent from that system would now abort the event-handling pass instead of being ignored.

### Checked and fine

The clocked and restart branches pass `[]` as the property accumulator into `call_terms_of_node_call`, so the candidate goes through the same clock/restart guarding as the other lifted properties; the `lift_term` / `lift_state_var` of `sofar_assump` sits under exactly the same guard as the pre-existing `node_assumes`, so it adds no new `Invalid_argument` exposure; the summary and results paths already filter candidates (`get_prop_status_and_kind_all_nocands`, `split_properties_nocands`), and `all_props_proved` / `at_least_one_prop_falsified` still exclude them, so exit codes are unaffected.

daniel-larraz:
Both findings addressed in 687c9582.

**1. Duplicate candidate names.** The name now carries the lifted flag: `Assumptions of call at l13c10 have held so far (Main.res.inst_Sub_l13c10_Sub.res.sofar_2)`. Checked on `s[i] = Sub(0 -> pre s[i])` with `int^3`: three instances, three distinct candidates. The `Generated` source reuses the same lifted state variable. The three lifted assumption properties of that call still share the name `Sub[L13C10].assume[L3C3]`, as before this PR; I left that as is since it is `subrequirements_of_contract`'s naming and predates the change. The model is added as `tests/regression/success/array_indexed_call_assume_history.lus`. It is not a discriminating test: with the old naming k-induction still proved it, because the three candidates are proved in the same step and the bookkeeping loss only shows when the instances' histories are proved at different times.

**2. `PropertyNotFound` in `F_relay`.** `reports` is replaced by `silenced`, which returns false in relay mode before any lookup, so the three logging functions are again complete no-ops there and the `StepCex` branch is unaffected.

Full regression suite: 596 passed. OUnit passes.

Issue #1491, opened 2026-09-08
match on a node call duplicates the scrutinee, giving a spurious counterexample


lorchrob:
### Summary

A `match` expression whose scrutinee is a **node call** evaluates that call independently in every arm. Each syntactic copy of the scrutinee becomes its own node instance, so the arms can disagree about which constructor the value has. A projection then lands on the wrong constructor and yields the canonical default, producing a **spurious counterexample** for a property that is valid by construction.

Binding the call to a local first is proven valid, so the two formulations disagree.

### Reproducer

```lustre
datatype T = A (v: int) | B (w: int);

node imported nd () returns (t: T);
(*@contract guarantee t = A (1) or t = B (2); *)

node main () returns (ok: bool);
var r: int;
let
  r = match nd () with
      | A (v) : v
      | B (w) : w
      end;
  ok = true;
  check "shared scrutinee" r = 1 or r = 2;
tel
```

`nd` is guaranteed to return `A(1)` or `B(2)`, so `r` must be 1 or 2 under any consistent scrutinee.

### Observed

```
<Failure> Property shared scrutinee is invalid by bounded model checking for k=0

Counterexample:
  Node main ()
    == Locals ==
    r  0

  Node nd (main[L9C13])
    == Outputs ==
    t  A(1)

  Node nd (main[L9C13])
    == Outputs ==
    t  A(1)

  Node nd (main[L9C13])
    == Outputs ==
    t  A(1)
```

`r = 0` is a value the contract makes impossible. Note the **three** instances of `nd` reported at the single call site `main[L9C13]` — one per syntactic copy of the scrutinee: the arm-`A` tester, and the field projection substituted into each of the two arm bodies.

### Expected

`shared scrutinee` is valid: all arms of one `match` see one value.

### Controls

Same datatype and same imported node in each case:

| Scrutinee | Result |
|---|---|
| `match nd () with ...` (node call) | **invalid at k=0**, `r = 0` |
| `s = nd (); match s with ...` | valid (k=1) |
| `match f (x) with ...` where `f` is a `function` | valid (k=1) |

The function case is correct as-is: a function call is a single deterministic UF application and is safe to duplicate. The defect is specific to node calls.

### Cause

`LustreDesugarADTs.desugar_expr` computes `scrut'` once and then copies that AST into every arm:

https://github.com/kind2-mc/kind2/blob/develop/src/lustre/lustreDesugarADTs.ml#L613-L624

[...]

PR #1531, opened 2026-09-16
Match blocks

lorchrob:


src/lustre/lustreAstDependencies.ml
```diff
@@ -1618,26 +1665,54 @@ let mk_graph_eqn: node_summary
...
+    let scrut = LH.abstract_pre_subexpressions scrut in
+    let* scrut_gs = mk_graph_expr2 m scrut in
+    let scrut_g =
+      List.fold_left union_dependency_analysis_data
+        (resolve_inherited inherited (LH.vars_without_node_call_ids scrut))
+        scrut_gs
+    in
+    (* Several arms become a chain of testers gating every equation; a single
+       exhaustive arm collapses to its equations, which read only the binders *)
+    let gated_by_tester = List.length arms > 1 in
```
daniel-larraz, line 1693: This misses cycles through the scrutinee when a match block has a single **non-exhaustive** arm. The arm count isn't what matters here. `LustreDesugarMatchBlocks` still turns a single non-exhaustive arm into `WhenBlock (cond, items, [])` guarded by the tester (the `[(cond, items)]` case of `chain`), so the arm's equations depend on the scrutinee just as they do with several arms.

Reproducer (accepted, and `y >= 0` is reported valid):

```lustre
datatype T = A(a: int) | B(b: int);
node main(u: int) returns (y: int);
var s: T;
let
  s = if y > 0 then A(u) else B(u);
  frame (y) y = 0;
  let
    match s with | A(v): y = 1; end
  tel
  check y >= 0;
tel
```

Here `y` depends on `s` through the `A?` tester and `s` depends on `y`, so this is the cycle y -> s -> y. Adding a second arm `| B(w): y = 2;` gets the expected `Cyclic dependency detected in definition of identifiers: y -> s -> y`. The equivalent match expression (`y = match s with | A(v): 1 | _: 2 end`) is also rejected correctly.

Suggested fix: base `gated_by_tester` on whether the desugaring emits a tester, i.e. `List.length arms > 1 || not exhaustive`, using the same `LustreCheckMatchExpressions.is_exhaustive` check as the desugaring. If the type context isn't available here, a conservative syntactic version works: a single arm counts as ungated only when its pattern is a variable pattern.

lorchrob: Looking into this opened a bit of a can of worms. The difficulty is in trying to be
smart about it — avoiding a dependency between an arm's equation LHS and the
scrutinee unless one is strictly necessary. For now I've done the simpler thing and
always draw it; if that turns out to be annoying in practice we can pursue something
more precise later. 

While poking around I also hit a soundness issue. Desugaring a match block into when
blocks duplicates the scrutinee expression across the tester and the branches, and
each copy of a node call becomes a separate instance under a different clock, so a
stateful node advances at a different rate in each and the tester and the projection
read different points of its stream. For now I've conservatively disallowed node
calls and `any` in a match block scrutinee; functions and `choose` are fine. This is
not specific to match blocks — #1491 is the same thing for match expressions, where
the normalizer's node call cache happens to merge the copies for a concrete node but
not for an imported one. The general issue deserves more thought.

PR #1537, opened 2026-09-17
Address missing dependency analysis check

lorchrob:


src/lustre/lustreAstDependencies.ml
```diff
@@ -1615,29 +1630,71 @@ let mk_graph_eqn: node_summary
...
-  | IfBlock (_, _, nis1, nis2) :: items
-  | WhenBlock (_, _, nis1, nis2) :: items -> 
-    let* gs1 = mk_graph_node_items m nis1 in
-    let* gs2 = mk_graph_node_items m nis2 in
-    let* gs3 = mk_graph_node_items m items in
-    R.ok (union_dependency_analysis_data gs1 (union_dependency_analysis_data gs2 gs3))
-  | FrameBlock (_, _, nes, nis) :: items -> 
+    let* gs, lhss = mk_graph_node_items m items in
+    R.ok (union_dependency_analysis_data g gs, eqn_lhss eqn @ lhss)
+  | IfBlock (_, cond, nis1, nis2) :: items
```
daniel-larraz, line 1681: The PR adds the condition edges only to each node's own dependency graph, here in `mk_graph_node_items`. The node summary is built separately by `mk_node_summary`, which reads equations through `LH.extract_node_equation`, and that helper drops block conditions. So a node's summary never records that an output depends on the inputs its block condition reads, and a cycle that goes through a node call is still accepted.

Reproduction:
```lustre
node M(x:int) returns (y:int); let if x > 0 then y = 0; else y = 1; fi tel
node N() returns (z:int); let z = M(z); check z >= 0; tel
```
Kind 2 accepts this and reports the property as valid (k=1). The interpreter then fails with "Transition relation not satisfiable". Writing `M` as `y = if x > 0 then 0 else 1` is correctly rejected as a cyclic dependency.

Suggested fix: make `mk_node_summary` include the same condition edges, either through an equation extractor that keeps block conditions or by building the summary from what `mk_graph_node_items` produces.

lorchrob: Fixed — mk_node_summary now walks the node items itself instead of going through extract_node_equation, so a block's condition is recorded as a dependency of everything the block defines.

Two things that came out of it:

The same map also merged definitions by override rather than union, so a variable defined in both branches of a block kept only the last branch's dependencies. if x > 0 then y = w; else y = 1; fi recorded y as depending on nothing, and a cycle through w got through the same way. Now merged by union.

After merging main, match blocks had the same gap — the scrutinee is dropped from the summary just as conditions were — so the traversal threads that through too.

On the node's own graph I switched the condition over to the inherited graph that match blocks introduced, which does the same job as the lhs-collecting code I had and removes it.

